PR #22, opened 2012-10-04
Issue 1060236: Multi extends which set the same option, and += behaviour

benoitbryon:
This pull-request is related to https://bugs.launchpad.net/zc.buildout/+bug/1060236

This is an implementation of the proposal described as first example of the ticket above:
- when using multiple inheritance...
- bases are merged with each other
- distinct sections and options are preserved as is
- duplicate sections and options are combined

```
.. code-block:: cfg
   :filename: buildout.cfg

   [buildout]
   extends = base1.cfg base2.cfg
   parts += foo

.. code-block:: cfg
   :filename: base1.cfg

   [buildout]
   parts = base1

.. code-block:: cfg
   :filename: base2.cfg

   [buildout]
   parts = base2
```

We get `parts = base1 base2 foo` (in fact: "base1\nbase2\nfoo").

If the feature is accepted, I'm ok to add more documentation if you like.


mauritsvanrees:
This is probably fixed in release [3.1.1](https://pypi.org/project/zc.buildout/3.1.1/) from today.  Please reopen or recreate a PR if this is not the case, or open an issue.

PR #137, opened 2013-07-19
remove dead (?) code, typofix

jamur2:
When fixing some typos I found, I tried to add a test, but I couldn't figure out how to trigger the "Couldn't download distribution" exception, which requires ```Installer._fetch``` to return ```None```.  I can't see that happening, at least with a current ```pkg_resources.Distribution```.

If someone can give me a hint as to when this error would happen, I'd be happy to write a test.  If not, here's a pull request to remove dead code for your consideration.

### DIFF
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -497,10 +497,6 @@ class Installer:
             try:
                 dist = self._fetch(avail, tmp, self._download_cache)
 
-                if dist is None:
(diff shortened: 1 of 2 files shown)


reinout:
I've applied the one typo fix on master. The other change ("if ... None") should not be made. That isn't dead code, it really needs to happen (as far as I can see it). There *was* a similar typo in that piece of code, though, so I fixed that one, too.

Closing the pull request.

Issue #146, opened 2013-09-19
Bootstrap should not download ez_setup if site-packages or find-links has it


pombredanne:
especially if setuptools is already installed, for instance in a venv or if you want a specific version (see #149 )
This prohibits to bootstrap without a network connection, and this is regression from the previous behavior

PR #147, opened 2013-09-19
A saner way to bootstrap setuptools

pombredanne:
This is a fix for #146 and the regression introduced in https://github.com/buildout/buildout/commit/9bcc9042b5ff40b3a3fb24d1b4fce7d95549e51d 

With this patch bootstrap makes a best effort to use existing setuptools if installed or available in find links.

Either respects the --allow-site-packages options and use setuptools there if installed.
Or honor the --find-links if there is an ez_setup.py available there.
Or as a last resort fetch and use the remote ez_setup.py

### DIFF
--- a/bootstrap/bootstrap.py
+++ b/bootstrap/bootstrap.py
@@ -77,8 +77,6 @@ find_links = os.environ.get(
 
 ######################################################################
 # load/install setuptools
-
-
 def install_setuptools():
     '''Install setuptools either from an ez_setup.py found in find-links
     or from a remote URL'''


reinout:
There's code in the latest bootstrap that allows offline work. From the changelog:

The bootstrap script now accepts ``--to-dir``. Setuptools is installed
  there. If already available there, it is reused. This can be used to
  bootstrap buildout without internet access. Similarly, a local
  ``ez_setup.py`` is used when available instead of it being downloaded. You
  need setuptools 14.0 or higher for this functionality.

PR #162, opened 2014-02-07
Added --virtualenv option, to create a buildout that exists in a virtualenv.

toumorokoshi:
Hello,

It would be nice if bootstrap.py had a --virtualenv option to automatically generate virtualenvs, and link buildout to that environment.

Adding --virtualenv creates a virtualenv under ./venv/, then uses that python to run bootstrap.py

### DIFF
--- a/bootstrap/bootstrap.py
+++ b/bootstrap/bootstrap.py
@@ -91,12 +91,12 @@ except ImportError:
 
 def _setup_virtualenv():
     """ Setup and install virtualenv """
-    print "Downloading virtualenv..."
+    print("Downloading virtualenv...")
     import gzip, tarfile, io, tempfile, shutil
     tf = tarfile.open(fileobj=io.BytesIO(urlopen(VIRTUALENV_URL).read()))
     temp_folder = tempfile.mkdtemp()
     tf.extractall(temp_folder)
-    print "Calling virtualenv..."
+    print("Calling virtualenv...")
     subprocess.call([sys.executable, os.path.join(temp_folder, 'virtualenv-1.10', 'virtualenv.py'), 
                      '--no-setuptools', '--no-pip', '--no-site-packages', os.path.join(os.curdir, 'venv')])
     shutil.rmtree(temp_folder)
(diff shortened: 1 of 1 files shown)


bootstrap/bootstrap.py
```diff
@@ -18,10 +18,18 @@
 use the -c option to specify an alternate configuration file.
 """
 
+######################################################################
+# Download URLs/Constants
+VIRTUALENV_OPTION = "--virtualenv"
+
+VIRTUALENV_URL = "https://pypi.python.org/packages/source/v/virtualenv/virtualenv-1.10.tar.gz"
```
reinout: That's a hardcoded URL that we'd have to update constantly.

bootstrap/bootstrap.py
```diff
@@ -59,10 +67,16 @@
 parser.add_option("--allow-site-packages",
                   action="store_true", default=False,
                   help=("Let bootstrap.py use existing site packages"))
+parser.add_option(VIRTUALENV_OPTION, action="store_true", default=False,
+                  help=("Use virtualenv to sandbox buildout"))
 
 
 options, args = parser.parse_args()
 
+
```
reinout: Unnecessary whitelines

reinout:
My gut feeling is that this change makes bootstrap too big. bootstrap.py is already not the prettiest code and this pull request makes it even more complex.

On the other hand... I *can* imagine the use of this addition. Something needs to be done about the fixed virtualenv download URL, in any case.

jimfulton:
Could this be a Python 3 only feature that relies on built in virtualenv
support? At least initially?

Jim

reinout:
@jimfulton: sounds like a good idea. It would probably make the pull request quite a bit more light-weight.

reinout:
@toumorokoshi: what do you think about Jim's python 3 suggestion?

toumorokoshi:
@reinout from a personal standpoint, I think that it would be less confusing to just not offer the virtualenv option at all, rather than have an optional command that only works for Python 3 and doesn't work when someone tries to build a project with Python 2. As someone who works in both fairly regularly, I appreciate when behavior is identical between 2 and 3 versions. 

If no one is interested in this feature, please feel free to close it. I was under the impression this wasn't going to be considered after a couple months of inactivity, so I solved my problem a different way. However, thank you for looking at this!

Issue #171, opened 2014-03-06
Shouldn't the cache-directory be created if missing


lelit:
Currently the download-cache directory must be already present, if it is specified in the [buildout] section, otherwise buildout immediately exits with the following error:

```
bin/buildout 
While:
  Initializing.
Error: The specified download cache:
'/tmp/dc-test/download-cache'
Doesn't exist.
```

Shouldn't it create the directory instead? When it is set, buildout automatically creates the `dist` subdirectory under that if it is missing, but not the parent...

My current use-case is that I'm trying to make the installation of a product as simple as possible using buildout: given that the product setup may require the download of a third-party component (done by a custom recipe), I'd like to cache that to avoid repeated download (when the product is updated). I thought that buildout own download-cache setting was appropriate for this, but it suffer of the above problem.

Of course my custom recipe is able to create the directory by itself, but it seemed sensible to reuse the global setting...

I'm willing to code the fix with the needed tests, if you think this is reasonable.

WouterVH:
what is blocking this issue ?

lelit:
Good question indeed!

reinout:
The tilde-expansion is what I wanted.

Blocking? It needed someone trying to get through as many as possible of the 20+ pull requests. I think I've merged 8 or so in the last couple of weeks. 

Now we need to wait for a new release (#249). I cannot release buildout, so I'll drop an email into the mailinglist to ask for someone else to pick it up.

I'm closing this issue, but feel very, very free to add "yeah, release me" messages to #249 .

PR #172, opened 2014-03-09
Issue171

lelit:
The first commit is just a note in DEVELOPERS.txt, it always takes me a few iterations to understand why most test fails on my system...

The second one strictly fixes the issue #171 I opened.

The third one implements Jim's extra-bonus-award suggestion. Please check also the english error message, it seems ok to me, but I'm not a native speaker.

I didn't add a test against the `~/subdir` form, because it didn't seem appropriate creating/removing things in the user directory while testing... but I'm open to suggestions.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,14 @@ Unreleased
 - Add ``BUILDOUT_HOME`` as an alternate way to control how the user default
   configuration is found.
 
+- Treat ``download-cache`` and ``eggs-directory`` in a special manner:
+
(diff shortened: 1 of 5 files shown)


reinout:
Ok, this one looks nice and has been left alone for far too long.

I've pulled your changes into a separate branch so that I could commit a couple of small changes/fixes/improvements: #251

reinout:
Well, it is merged now! Please look at #251 to see what I improved. And to see if I did something horribly wrong :-)

PR #181, opened 2014-04-15
Added "extension_order" option

jajadinimueter:
Added "extension_order" option to the "buidout" section.
This allows you to specify the order in which the declared
extensions are loaded.

### DIFF
--- a/src/zc/buildout/buildout.py
+++ b/src/zc/buildout/buildout.py
@@ -966,6 +966,7 @@ class Buildout(DictMixin):
 
     def _load_extensions(self):
         __doing__ = 'Loading extensions.'
+        extension_order = self['buildout'].get('extension_order', '').split()
         specs = self['buildout'].get('extensions', '').split()
         for superceded_extension in ['buildout-versions',
                                      'buildout.dumppickedversions']:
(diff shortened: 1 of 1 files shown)


reinout:
Why do you need a separate extension order option? Cannot you get buildout to respect the order given in the actual `extensions=` option? Seems like that would be cleaner.

It might be that I mis-understand, though. In any case, we need documentation or tests.

Or is there some input or help that you need?

mauritsvanrees:
There was no action on this PR the last ten years, so I close it.
Feel free to reopen.

If you want to pick up this PR again:

* Please rebase your work on latest master and force push.  There have been lots of changes.
* Please explain the purpose of this change, what use case or problem this solves, write some documentation.
* I guess the main point is that the order in which `pkg_resources.iter_entry_points('zc.buildout.extension')` lists the entry points may be undefined.  Is that it?
* But then can't we use the order of `self['buildout'].get('extensions', '')`, instead of introducing a new `extension_order` option?
* If no new option is needed, then I don't think new tests are needed, as long as the current ones pass of course.  There are some tests with extensions in `buildout.txt`, and also `test_increment.py`.
* When checking `if ep.dist.key == item`, these days you should probably call `packaging.utils.canonicalize_name` on both sides of the equation.  We already pull `packaging` in as dependency.

PR #188, opened 2014-06-26
fix _distutils_script, check fail if the file start by #! but not is a python script

muchasuerte:
I found an error when I install a python egg that include a shell script.  
The fix allow the copy of non python script otherwise two brackets are missing after the *not* at line 1104

### DIFF
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -1101,9 +1101,12 @@ def _distutils_script(path, dest, script_content, initialization, rsetup):
         dest += '-script.py'
 
     lines = script_content.splitlines(True)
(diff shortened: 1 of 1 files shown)


reinout:
One year later, sorry...

a) this will need a test.

b) distutils scripts are quite old/rare. I only added support for distutils scripts because pyflakes needed it. Apart from pyflakes, buildout functioned quite well without distutils script support for quite some time. So... shell scripts as distutils scripts sounds like a corner case of a corner case to me. How common are they? Do we really need to support this?

My gut feeling is that it is barely used. Do you have a real usecase for it?

I'm closing the pull request for now, feel free to re-open it if you have more info (and at least one example) and a quick test.

PR #191, opened 2014-09-17
Adds autocompletion to interactive python shell

paschembri:
Adds autocompletion for linux, win32 and Mac OS X platforms.


mgedmin:
(This was replaced by #233.)

PR #195, opened 2014-10-30
Correctly calculate relative script path when path == relative_paths.

jackzelenkin:


reinout:
@youngking correctly saw that #278 fixes the same issue in mostly the same way. I'll close this one and merge #278 once someone has reviewed the code,

reinout:
@jackzelenkin: I just released 2.4.6 with the #277 fix in there.

PR #204, opened 2014-11-06
Add a version number to bootstrap.py

### LINKED ISSUE #30: add some versioning info to bootstrap.py
It's a pain that you have no idea whether you have an old/new version of it.

mgedmin:
Fixes #30

### DIFF
--- a/bootstrap/bootstrap.py
+++ b/bootstrap/bootstrap.py
@@ -25,7 +25,9 @@ import tempfile
 
 from optparse import OptionParser
 
-tmpeggs = tempfile.mkdtemp()
+__version__ = '2.2.6'  # of bootstrap.py itself
+
+tmpeggs = tempfile.mkdtemp(prefix='bootstrap-')
 
 usage = '''\
 [DESIRED PYTHON FOR BUILDOUT] bootstrap.py [options]
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,14 @@
 Change History
 **************
 
+2.2.6 (unreleased)
+==================
+
+- Added a version number to the bootstrap.py script.
+  There's no way to see it without looking at the script sources because
+  bootstrap.py already uses --version for a different purpose.
+  Closes #30.
+
 2.2.5 (2014-11-04)
 ==================
 


reinout:
So, the buildout version isn't the handiest version for the bootstrap? Separate versions might also a bit weird.

What about a textual comment at the top? "This bootstrap file shipped with buildout 2.2.3"?

On the other hand... what are we trying to accomplish? The bootstrap is generally something you'd like to have at the latest version. If a colleague has a problem with one of his buildouts, I often suggest to "grab the latest bootstrap" when they have such a problem.

Brainstorming... Perhaps let bootstrap add a `__bootstrapped_with__ = '2.2.3'` to the bin/buildout it generates? Buildout could read that and print a suggestion at the top if it knows it to be an old version.

Throwing the idea out there :-)

gotcha:
I support the idea of having a version number in bootstrap. It would help when supporting someone to be able to check version. Sthing like `bootstrap --version`. Further, I do not think it needs to be synced with `zc.buildout` version.

idgserpro:
What about ```__version__``` just having a date timestamp? Just my 0.02c to this discussion.

reinout:
I'm working on a couple of pull requests, one of them dealing with bootstrapping, so I'll probably be forgiven some bikeshedding/brainstorming.

- Yes, we need *some* marker/version inside the bootstrap.py

- Whatever the marker, it needs to be advertised somewhere. Preferably in buildout's own changelog. "We shipped with a new bootstrap".

- A version is dangerous. If there's no bootstrap change, you'll end up with a 2.2.3 bootstrap being used with a 2.5.1 buildout, which looks weird. Same is using a 1.2 bootstrap (if we'd use our own version numbers) with a 2.4 buildout.

- Timestamp seems like a nice compromise. Best approach is to print it when running as the `--version` is being used to pick a specific buildout version. Actually, I'd better rename that into --buildout-version. There can't be too many people using that option, right?

bootstrap/bootstrap.py
```diff
@@ -25,7 +25,9 @@
 
 from optparse import OptionParser
 
-tmpeggs = tempfile.mkdtemp()
+__version__ = '2.2.6'  # of bootstrap.py itself
+
+tmpeggs = tempfile.mkdtemp(prefix='bootstrap-')
```
reinout: This is an unrelated change, but quite a nice one.

I've included it in #248, where it sits next to other bootstrap improvements.

reinout:
I've done a more comprehensive change in https://github.com/buildout/buildout/commit/6d7b39c0fe94dfc0481b0b6ebc0d9b3915af803d (included in #248).

- Version is a datestamp. See the changelog in the commit I referenced why that's the best idea. The reason there is similar to my earlier comment.

- bootstrap.py now has `--version` that actually prints the version. I moved the older `--version` to `--bootstrap-version`.

Does that look OK?

reinout:
I'm closing the pull request to reduce confusion due to many pull requests being open. I'm assuming that's fine (assuming #248 gets accepted).

PR #213, opened 2014-11-24
Our VersionConflict handling could result in an UnpackError when rendering

### LINKED ISSUE #212: Buildout's VersionConflict doesn't always match setuptools' usage
We wrap `pkg_resources.VersionConflict` in our own `zc.buildout.easy_install.VersionConflict` to provide a nice error message.

Problem: our VersionConflict depends on implementation details of setuptools/pkg_resources that don't always apply. The version as we have it now:

    class VersionConflict(zc.buildout.UserError):

        def __init__(self, err, ws):
            ws = list(ws)
            ws.sort()
            self.err, self.ws = err, ws

        def __str__(self):
            existing_dist, req = self.err.args
            result = ["There is a version conflict.",
                      "We already have: %s" % existing_dist,
                      ]
            for dist in self.ws:
                if req in dist.requires():
                    result.append("but %s requires %r." % (dist, str(req)))
            return '\n'.join(result)

The problem is the first line in `__str__()`: it unpacks `self.err.args`. It assumes two arguments. There is however one spot in pkg_resources were pkg_resources only returns a single string instead of two arguments. The string is already a well-formatted friendly error message. This results in an unpack error:

      File "bin/buildout", line 12, in <module>
        sys.exit(zc.buildout.buildout.main())
      File "/home/vagrant/.buildout/eggs/zc.buildout-2.2.1-py2.7.egg/zc/buildout/buildout.py", line 1950, in main
        _doing()
      File "/home/vagrant/.buildout/eggs/zc.buildout-2.2.1-py2.7.egg/zc/buildout/buildout.py", line 1687, in _doing
        message = str(v)
      File "/home/vagrant/.buildout/eggs/zc.buildout-2.2.1-py2.7.egg/zc/buildout/easy_install.py", line 1304, in __str__
        existing_dist, req = self.err.args
    ValueError: need more than 1 value to unpack

I'm preparing a fix for this now.

reinout:
See #212 

Fixes #212 

Includes test (that indeed failed initially). Includes fix (of course) and a changelog entry (hurray!).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,14 @@
 Change History
 **************
 
+2.2.6 (unreleased)
+==================
+
+- Fix for #212. For certain kinds of conflict errors you'd get an UnpackError
+  when rendering the error message. Instead of a nicely formatted version
+  conflict message.
+  [reinout]
+
 2.2.5 (2014-11-04)
 ==================
 
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -1346,13 +1346,16 @@ class VersionConflict(zc.buildout.UserError):
         self.err, self.ws = err, ws
 
     def __str__(self):
-        existing_dist, req = self.err.args
-        result = ["There is a version conflict.",
-                  "We already have: %s" % existing_dist,
-                  ]
-        for dist in self.ws:
-            if req in dist.requires():
-                result.append("but %s requires %r." % (dist, str(req)))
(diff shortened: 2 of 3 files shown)


reinout:
2x LGTM and I've updated it to the latest master. Shall I merge it myself? I need similar functionality for the fix for #246, which is kinda really blocking.

tseaver:
I'd say go ahead and merge, given that my objection was to you matching the ridiculously-ugly existing style of the file.

Issue #217, opened 2014-12-14
bootstrap.py without --allow-site-packages fails with ImportError on Windows


fpatz:
On Windows, bootstrap.py does not work without --allow-site-packages. Bootstrapping fails with an ImportError, not finding "email" from the standard library (see traceback below). The reason is the code for allow_site_packages=False: it removes everything from sys.path that contains anything from site.getsitepackages(). But on Windows, for whatever reason, site.getsitepackages() contains the base installation path C:\Python27 (this is sys.prefix, as added by site.py for the Windows case). Thus the code path for allow_site_packages=False removes *everything* from sys.path.

I think this was introduced with buildout/buildout@682dd5a690ca9c39ba7c98d3aca35028ac04cd41; the failure is also visible on winbot.zope.org.

> c:\Python27\python bootstrap.py
Downloading https://pypi.python.org/packages/source/s/setuptools/setuptools-8.0.1.zip 
[...]
Traceback (most recent call last):
  File "bootstrap.py", line 92, in <module>
    ez['use_setuptools'](**setup_args)
  File "<string>", line 140, in use_setuptools
  File "<string>", line 128, in _do_download
  File "build\bdist.win32\egg\setuptools\__init__.py", line 5, in <module>
  File "c:\Python27\lib\distutils\core.py", line 19, in <module>
    from distutils.dist import Distribution
  File "c:\Python27\lib\distutils\dist.py", line 10, in <module>
    from email import message_from_file
ImportError: No module named email

mgedmin:
@jimfulton you asked for a nag ;)

This issue is causing ZODB tests to [fail on winbot](http://winbot.zope.org/builders/ZODB_dev%20py_270_win32/builds/1624/steps/bootstrap/logs/stdio) and is the only remaining unresolved issue on the daily zope-dev test failure report.

The fix in #218 seems pretty straightforward.

lelit:
Maybe a second nag is needed... FWIW, I confirm that #218 fixed the problem for me too.

mgedmin:
#218 was merged, now we wait for somebody who can to update the copy at https://bootstrap.pypa.io/

mgedmin:
FWIW the new bootstrap [fixed](http://winbot.zope.org/builders/ZODB_dev%20py_270_win32/builds/1716) ZODB tests on winbot.

agroszer:
winbot fixed: yeah!

tseaver:
We need a release *here* which fixes the issue.

[...]

PR #218, opened 2014-12-14
Potential fix for https://github.com/buildout/buildout/issues/217

fpatz:
On Windows, `sys.prefix` is a site-package directory, i.e. for a
standard installation `C:\Python27`. The code path for `not
allow_site_packages` removed *everything* from `sys.path` that contains
*any* of the the site-packages directories as reported by
 `site.getsitepackages`, which on Windows removes all pathes below
 `sys.prefix` from `sys.path`.


tseaver:
LGTM

tseaver:
@fpatz thanks for the diagnosis / fix.

Issue #86, opened 2013-02-24
Update boostrap.py on: http://downloads.buildout.org/2/bootstrap.py


rvanlaar:
Bootstrap.py on github is newer then the one on: http://downloads.buildout.org/2/bootstrap.py

Please upload a new one to buildout.org.

idgserpro:
Since we have bootstrap.py as "stable", wouldn't it just be easier point this url to https://raw.githubusercontent.com/buildout/buildout/master/bootstrap/bootstrap.py?

dstufft:
Just for completeness sake, Jim had asked us to pull the bootstrap.py file from the ``bootstrap-release`` branch. We can of course change that if it's no longer desireable to pull from that branch.

tseaver:
Seems like the fix for this issue is to make a new release (which will update `bootstrap/bootstrap.py` on the 'bootstrap-release' branch, presumably).

idgserpro:
@dstufft  The bootstrap.py from ```bootstrap-release``` is indeed, right now, equal to the one from
http://downloads.buildout.org/2/bootstrap.py (that redirects to https://bootstrap.pypa.io/bootstrap-buildout.py). I don't mind the bootstrap.py being from ```bootstrap-release``` branch. I just wanted to know where exactly it came from to avoid confusion. Thanks for the clarification. If this is true, this issue can be closed, am I correct? Or @rvanlaar is talking about master? Maybe he had the same doubts as me.

Do you think it's worth to explicitly state that the bootstrap.py from pypa comes from this branch, in zc.buildout readme or even bootstrap.py comments itself?

When the branch is updated, will you update the bootstrap.py in http://downloads.buildout.org/2/bootstrap.py? Is this process automated, to avoid issues like this one being opened to each bootstrap.py release? It *can* be automated?

dstufft:
The server backing bootstrap.pypa.io pulls the branches every 15 minutes and if I recall the CDN caches it for an hour. So it may take up to an hour and 15 minutes or so from when it gets merged till it updates.

idgserpro:
This is great to hear @dstufft. Do you think this can be written down in the readme as proposed in [my last comment](https://github.com/buildout/buildout/issues/86#issuecomment-114943615), along with the ```bootstrap-release``` info at the bootstrap.py section? What @tseaver @reinout @jimfulton think about it?

reinout:
A new bootstrap has been released.

[...]

PR #222, opened 2014-12-17
Run easy_install with specified setuptools_version

lrowe:
If setuptools is installed globally by easy_install then the easy_install.pth gets added to sys.path before PYTHONPATH and the wrong setuptools version is used.

### DIFF
--- a/bootstrap/bootstrap.py
+++ b/bootstrap/bootstrap.py
@@ -110,7 +110,12 @@ for path in sys.path:
 
 ws = pkg_resources.working_set
 
+setuptools_path = ws.find(
+    pkg_resources.Requirement.parse('setuptools')).location
+
+# Fix sys.path here as easy_install.pth added before PYTHONPATH
 cmd = [sys.executable, '-c',
+       'import sys; sys.path[0:0] = [%r]; ' % setuptools_path +
        'from setuptools.command.easy_install import main; main()',
        '-mZqNxd', tmpeggs]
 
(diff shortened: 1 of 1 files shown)


reinout:
Virtually the same change as in #232 for the actual buildout code. So this one is best merged, too.

Once a new release is out, the new bootstrap needs to be placed in some of the regular places. Who normally does this?

tseaver:
@reinout I think @jimfulton is the one who can manage download.buildout.org

vincentfretin:
@tseaver @reinout The bootstrap.py is at https://bootstrap.pypa.io/bootstrap-buildout.py
and it's updated every 15 minutes from master if I recall what @jimfulton said.

reinout:
The last modification date, if you look at the dir listing at https://bootstrap.pypa.io/ is 14-Dec-2014. So it doesn't look too automatic to me :-)

There is also a `bootstrap.py` at http://downloads.buildout.org/2/bootstrap.py, which is also mentioned in `easy_install.py` in an error message. The version there is identical to the abovementioned `bootstrap-buildout.py from pypa.io

Would it be a good idea to add information about updating the bootstrap to `DEVELOPERS.txt`?

reinout:
#86 has more info on how the boostrap.py actually ends up in the various places. I'm writing it down in `DEVELOPERS.txt` now.

reinout:
This one needs a release (#249), too, because of the inherent bootstrap.py release.

PR #223, opened 2014-12-18
Use local ez_setup.py if one exists; add --setuptools-download-base option

lrowe:
For a completely offline install would need to make download_setuptools recognize filesystem paths for download_base.

### DIFF
--- a/bootstrap/bootstrap.py
+++ b/bootstrap/bootstrap.py
@@ -61,7 +61,9 @@ parser.add_option("--allow-site-packages",
                   help=("Let bootstrap.py use existing site packages"))
 parser.add_option("--setuptools-version",
                   help="use a specific setuptools version")
-
+parser.add_option("--setuptools-to-dir",
+                  help=("allow for re-use of existing directory of "
+                        "setuptools versions"))
 
 options, args = parser.parse_args()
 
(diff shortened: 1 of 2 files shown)


lrowe:
The setuptools people ended up implementing a slightly different ``--to-dir`` option. I've updated this to reflect that and fixed the tests to use the bootstrap.py script which is different to buildout bootstrap.

WouterVH:
what's currently blocking this issue?

aclark4life:
Maybe nothing… maybe @jimfulton or @hannosch could review?

WouterVH:
When you don't have access to the public pypi, but only the a local pypi,
you still have to hack this script, to add a download-base, which is usually same as the find-links.

When we want to use a new setuptools-version, (which is very very often these days),
we needed to hacks the ez_setup.py + bootstrap.py to customize the hardcoded urls.

with this patch, we still only need to hack bootstrap.py to specify a download_base.

lrowe:
This PR needs updating to work with the slightly different approach they ended up implementing in ez_setup before it can be merged.

lrowe:
Looks like I just forgot I updated this to reflect the setuptools changes. I think its good to merge.

reinout:
I'll merge it, it looks good.

@lrowe, do you have some info for a quick changelog entry ("setuptools > x.y needed")?

lrowe:
@reinout it requires an ez_setup.py from 14.0 though using installing earlier setuptools versions should work. The changelog from setuptools is:

https://pythonhosted.org/setuptools/history.html#id14
* Bootstrap script now accepts --to-dir to customize save directory or allow for re-use of existing repository of setuptools versions. See Pull Request #112 for background.

Background in setuptools PR: https://bitbucket.org/pypa/setuptools/pull-request/112/allow-ez_setuppy-to-work-offline-by/activity

PR #224, opened 2014-12-23
allow user customize index-url

luthur:
For Chinese users, we may not always connected to the official pypi mirror because of the GFW. However, we can use douban pypi mirror.

I hope I can customize index-url through bootstrap.py
`python bootstrap.py -i http://pypi.douban.com/simple/ -e ez_setup.py`


reinout:
I guess this pull request can be closed as #223 effectively solves the same issue, though in a slightly different way. See the documentation change to `bootstrap.txt` in https://github.com/buildout/buildout/pull/223/files

This way you can add a ez_setup.py and a setuptools locally and get bootstrap to use them.

Would that be enough to solve your issue?

If it isn't enough, you now have an example in that #223 pull request to add a test. (and look in `DEVELOPERS.txt` on how to run the tests, two quick makefile instructions is enough to get them running).

Issue #176, opened 2014-03-17
extends with increments can result in missing values


djay:
The following test fails

```
def increment_buildout_with_multiple_extended_without_base_equals():
    r"""

    >>> write('buildout.cfg', '''
    ... [buildout]
    ... extends = base1.cfg base2.cfg
    ... parts += foo
    ... [foo]
    ... recipe = zc.buildout:debug
    ... [base1]
    ... recipe = zc.buildout:debug
    ... [base2]
    ... recipe = zc.buildout:debug
    ... ''')
    >>> write('base1.cfg', '''
    ... [buildout]
    ... extends = base3.cfg
    ... parts += base1
    ... ''')
    >>> write('base2.cfg', '''
    ... [buildout]
    ... extends = base3.cfg
    ... parts += base2
    ... ''')
    >>> write('base3.cfg', '''
    ... [buildout]
    ... ''')

    >>> print_(system(buildout), end='')
    Installing base2.
      recipe='zc.buildout:debug'
    Installing foo.
      recipe='zc.buildout:debug'
    Installing base1.
      recipe='zc.buildout:debug'
    """
```

Currently fails as follows

```
Expected:
    Installing base2.
      recipe='zc.buildout:debug'
    Installing foo.
      recipe='zc.buildout:debug'
    Installing base1.
      recipe='zc.buildout:debug'
Got:
    Installing base2.
      recipe='zc.buildout:debug'
    Installing foo.
      recipe='zc.buildout:debug'
```

If you remove [buildout] from base3.cfg or remove the extends=base3.cfg the test will pass.

Might be related to #1060236

distributist:
I've created a pull request that (I hope) fixes this issue: https://github.com/buildout/buildout/pull/226

PR #226, opened 2015-01-07
Fix for issue 176: extends with increments can result in missing values

distributist:
This is a suggested fix for issue 176. I've added a test case based on the one presented in that issue, but have expanded it to test various combinations of operations. 

The basic approach is to parse all the extended files before applying them, so that nothing gets lost along the way. Additionally, all annotations are kept so that even if a = overrides previous settings, the operation is captured so the problem can more easily be found.

I believe the fix behaves consistently, so that =, -= and += behave as expected within the order of extends operations. As the test case shows, you can do some stupid things, but it does them consistently. ;-)

### DIFF
--- a/src/zc/buildout/buildout.py
+++ b/src/zc/buildout/buildout.py
@@ -1576,18 +1576,30 @@ def _open(base, filename, seen, dl_options, override, downloaded):
     if root_config_file and 'buildout' in result:
         dl_options = _update_section(dl_options, result['buildout'])
 
+    eresults = []
     if extends:
         extends = extends.split()
-        eresult = _open(base, extends.pop(0), seen, dl_options, override,
-                        downloaded)
         for fname in extends:
-            _update(eresult, _open(base, fname, seen, dl_options, override,
-                    downloaded))
-        result = _update(eresult, result)
-
+            eresults.extend(
+                _open(
+                    base, 
+                    fname, 
+                    seen, 
+                    dl_options, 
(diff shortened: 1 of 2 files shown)


djay:
how does this relate to the existing fix in https://github.com/buildout/buildout/pull/177 ?

distributist:
Umm, I hadn't noticed that one. *blush*

I spent forever trying to figure out why a complex buildout was broken, finally found issue #176 which showed me why, and immediately set to work developing a patch I could deploy to move a project forward. 

Looking at #177, the approach is different. I have changed update_section minimally, and instead use the recursive call of _open to return a flattened set of each config file, which then all get applied in order at the very end. 

The test case tests several operations (+=, -=, =) at various levels, though I didn't test for the two items mentioned in #177: omitting the space before the += or having = and += in the same file. I'd be happy to do those if anyone thinks this fix has merit.

reinout:
I don't use the minus stuff myself (I do use plus). So I'm not familiar with the problems. A long while ago I added the plus-minus-issues label to related issues. 

Would accepting this PR help fix the others? Or are there incompatible wishes in the various PRs? As said, I haven't got the full picture about what people are doing with these options.

mauritsvanrees:
Does your PR #644 replace this one?

distributist:
> Does your PR #644 replace this one?

Yes. This was my original patch  from 9 years ago. ;-) I'll close it.

Issue #227, opened 2015-01-07
dev.py fails with recent setuptools.


mgedmin:
Someone complained on IRC that buildout's tests are failing.  You can see the same failure on Travis: https://travis-ci.org/buildout/buildout/jobs/46070031#L97

The commit that triggered that build appears to be unrelated to the failure.

Last successful build on Travis was 22 [days ago](https://travis-ci.org/buildout/buildout/jobs/44200651#L97) and used setuptools 8.0.4.  The failing build uses setuptools 11.3.

I'm pretty sure the following setuptools change broke buildout's dev.py: [Moved pkg_resources into its own package](https://bitbucket.org/pypa/setuptools/commits/02658cf09ced08b8ff4a81e24c614eee3d219c70).

Bitbucket makes it hard to see which releases contain that commit.  GitHub makes it [easy](https://github.com/jaraco/setuptools/commit/79a0c55d8c09d3fdd68157faa8658ecd85aa5d45): every release since setuptools 8.3.

jimfulton:
Sigh. Thanks.

PR #228, opened 2015-01-07
Don't assume pkg_resources is a module.

mgedmin:
In setuptools < 8.3 pkg_resources.py was a module, so

    import pkg_resources
    print(pkg_resources.__file__)

would print something like

    /home/mg/src/buildout/eggs/setuptools-8.0.4-py2.7.egg/pkg_resources.py

Starting from setuptools 8.3 pkg_resources became a package, so the
above code prints

    /home/mg/src/buildout/eggs/setuptools-11.3.1-py2.7.egg/pkg_resources/__init__.py

which means a single os.path.dirname() is no longer sufficient to find
the location of the .egg file/directory.

I'm pretty sure 'setuptools' itself was always a package, so use two
os.path.dirname()s on setuptools.__file__ to find the .egg location more
reliably with both old and new setuptools versions.

Fixes https://github.com/buildout/buildout/issues/227

### DIFF
--- a/dev.py
+++ b/dev.py
@@ -50,14 +50,15 @@ except ImportError:
 exec(urlopen('https://bootstrap.pypa.io/ez_setup.py').read(), ez)
 ez['use_setuptools'](to_dir='eggs', download_delay=0)
 
-import pkg_resources
+import pkg_resources, setuptools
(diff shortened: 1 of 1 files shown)


tseaver:
LGTM

PR #229, opened 2015-01-07
Avoid prefixing buildout dir to 'special' rpath elements.

### LINKED ISSUE #225: zc.recipe.egg unneeddedly expands $ORIGIN-variables in rpath
I'm using the rpath option in zc.recipe.egg build to compile software.
In order to make the buildout-directory relocatable (eg. deploy it in any directory in a production-machine), I'm using the special $ORIGIN-variable.

However, zc.recipe.egg thinks $ORIGIN is a relative path, and prepends the buildout-directory in front of it, while in fact it should be kept as it.

```
    [python-oracle-cxoracle]
    recipe = zc.recipe.egg:custom
    egg = cx_Oracle
    rpath =
    #   ${buildout:oracle-home-directory}  
        $ORIGIN/
        $ORIGIN/../../parts/python-oracle
```

Introspecting the compiled library-file:
```
> cd /apps/buildouts/test-base
>  readelf -d develop-eggs/cx_Oracle-5.1.3-py2.7-linux-x86_64.egg/cx_Oracle.so

    Dynamic section at offset 0x24030 contains 24 entries:
    Tag        Type                         Name/Value
   0x000000000000000f (RPATH)     Library rpath: [$ORIGIN/../lib:$ORIGIN/../../lib:$ORIGIN/../../../lib:$ORIGIN/../../../../lib:/apps/buildouts/test-base/$ORIGIN/:/apps/buildouts/test-base/$ORIGIN/../../parts/python-oracle]
 ```
where a correct RPATh would be

```
   0x000000000000000f (RPATH)     Library rpath: [$ORIGIN/../lib:$ORIGIN/:$ORIGIN/../../parts/python-oracle]
```

tseaver:
Fixes #225.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,11 @@
 Change History
 **************
 
+- Fixed: In ``zc.recipe.egg#custom`` recipe's ``rpath`` support, don't
+  assume path elements are buildout-relative if they start with one of the
+  "special" tokens (e.g., ``$ORIGIN``).  See:
+  https://github.com/buildout/buildout/issues/225.
+
 - Fixed: Buildout merged single-version requirements with
   version-range requirements in a way that caused it to think there
   wasn't a single-version requirement.  IOW, buildout throught that
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,9 +22,10 @@ Unreleased
   https://github.com/buildout/buildout/pull/222 .
   [lrowe]
 
-- Fixed: In ``zc.recipe.egg#custom`` recipe's ``rpath`` support, don't
-  assume path elements are buildout-relative if they start with one of the
-  "special" tokens (e.g., ``$ORIGIN``).  See:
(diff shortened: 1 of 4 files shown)


reinout:
I've re-merged with master and the tests pass (at least locally, travis isn't ready yet).

There's no test for this pull request. On first sight, it looks like a test for the `build_ext()` function (which doesn't actually do any compiling, it just returns a dict) wouldn't be too hard. On the other hand, that function isn't tested right now, so I wouldn't mind much.

Something that I think needs fixing: there's an entry in zc.buildout's changelog, but it is a change in **zc.recipe.egg** instead, right? So that one needs the changelog entry (and a new release).

reinout:
@tseaver: agree that the changelog entry should move to zc.recipe.egg?

reinout:
I've copied the changelog part, keeping it also in zc.buildout's. And I'm noting the (necessary) zc.recipe.egg release there. My changes have only been bookkeeping stuff, so I'm going to go ahead and merge it.

PR #231, opened 2015-01-07
Use 'rcN' for release candidate versions in tests

mgedmin:
Recent setuptools versions changed the canonical spelling of release
candidate versions from 'cN' to 'rcN'.  Using the older spelling causes
a UserWarning about the version requiring normalization, and test
failures where the expected output contains a different (normalized)
version number.

For the record, the UserWarning looks like this:

/home/mg/src/buildout/eggs/setuptools-11.3.1-py2.7.egg/setuptools/dist.py:283:
UserWarning: The version specified requires normalization, consider
using '1.2rc1' instead of '1.2c1'.

and the failures can be seen at
https://travis-ci.org/buildout/buildout/jobs/46215472

### DIFF
--- a/src/zc/buildout/downloadcache.txt
+++ b/src/zc/buildout/downloadcache.txt
@@ -37,10 +37,10 @@ download:
     <a href="demo-0.1-py2.4.egg">demo-0.1-py2.4.egg</a><br>
     <a href="demo-0.2-py2.4.egg">demo-0.2-py2.4.egg</a><br>
     <a href="demo-0.3-py2.4.egg">demo-0.3-py2.4.egg</a><br>
-    <a href="demo-0.4c1-py2.4.egg">demo-0.4c1-py2.4.egg</a><br>
+    <a href="demo-0.4rc1-py2.4.egg">demo-0.4rc1-py2.4.egg</a><br>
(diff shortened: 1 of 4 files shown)


mgedmin:
BTW I have issues running buildout's tests locally -- ccache's gcc wrapper is in my $PATH and becomes Very Unhappy when buildout changes HOME to a nonexistent directory -- so I'm [relying on Travis](https://travis-ci.org/mgedmin/buildout/builds/46219421) to see if my fix fixes *all* the real errors.

mgedmin:
Ok, this fixes the tests on Python 2.7, but 2.6 gets additional failures, due to a new DeprecationWarning:
```
/sample-buildout/eggs/setuptools-11.3.1-py2.6.egg/setuptools/command/egg_info.py:171: DeprecationWarning: Parameters to load are deprecated.  Call .resolve and .require separately.
```
See https://travis-ci.org/mgedmin/buildout/jobs/46220052 for the full log.

reinout:
LGTM. I tried it locally with a fresh 2.6 build and everything runs OK. I assume it was a travis glitch or something fixed in the latest setuptools.

I'll merge it because otherwise the tests fail with a new setuptools.

PR #232, opened 2015-01-08
Run easy_install with specified setuptools_version

lrowe:
If setuptools is installed globally by easy_install then the easy_install.pth gets added to sys.path before PYTHONPATH and the wrong setuptools version is used.

Any chance of a 2.2.6 with this fix included?

### DIFF
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -318,7 +318,9 @@ class Installer:
         try:
             path = setuptools_loc
 
-            args = [sys.executable, '-c', _easy_install_cmd, '-mZUNxd', tmp]
+            args = [sys.executable, '-c',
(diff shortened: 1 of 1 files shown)


tseaver:
LGTM

lrowe:
If we do have a 2.2.6 including this then the equivalent fix for bootstrap.py in #223 should probably be included too.

reinout:
Agree with the LGTM, merging it. I'll add a changelog entry on master.

reinout:
Changelog added in b84e20d6e0148eae3219cb206b1ce1fcbc5d5ac6, @lrowe, is it OK like that?

reinout:
@lrowe, you probably meant that #222 should be included? Instead of the #223 you mention? (Just making sure)

lrowe:
@reinout - yup, I mistyped. #222 is the equivalent fix for bootstrap.py

PR #233, opened 2015-01-08
Support of PYTHONSTARTUP environment variable

paschembri:
When creating the python interpreter, it will load PYTHONSTARTUP if available

### DIFF
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -1324,9 +1324,12 @@ if len(sys.argv) > 1:
 if _interactive:
     del _interactive
 
-    filename = os.environ.get('PYTHONSTARTUP')
-    if filename and os.path.isfile(filename):
-        exec(open(filename).read())
+    _py_startup = os.environ.get('PYTHONSTARTUP')
+    if _py_startup and os.path.isfile(_py_startup):
+        with open(_py_startup, 'U') as _f:
+            exec(compile(_f.read(), _py_startup, "exec"))
+
+    del _py_startup
 
     __import__("code").interact(banner="", local=globals())
 '''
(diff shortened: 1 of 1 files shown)


mgedmin:
💗

mgedmin:
I have some suggestions, but mobile GitHub doesn't let me comment the code.  Later

src/zc/buildout/easy_install.py
```diff
@@ -1322,6 +1323,11 @@ def _pyscript(path, dest, rsetup, initialization=''):
 
 if _interactive:
     del _interactive
+
+    filename = os.environ.get('PYTHONSTARTUP')
+    if filename and os.path.isfile(filename):
+        exec(open(filename).read())
+
```
mgedmin: I think you should use the `with` statement here, and probably open the file in 'U' mode like the branch above.  Also, an explicit compile() step, again like in the above code (lines 1321-1322), so syntax errors get the right filename.

I also think the `filename` global should be renamed to start with an underscore, to blend with the code around it.  These names are global and leak into the interactive session, unless you take care to `del` them, which I think you should also do.

paschembri:
The code has been updated with your remarks.

mgedmin:
Looks good to me!

src/zc/buildout/easy_install.py
```diff
@@ -1322,6 +1323,14 @@ def _pyscript(path, dest, rsetup, initialization=''):
 
 if _interactive:
     del _interactive
+
+    _py_startup = os.environ.get('PYTHONSTARTUP')
+    if _py_startup and os.path.isfile(_py_startup):
+        with open(_py_startup, 'U') as _f:
+            exec(compile(_f.read(), _py_startup, "exec"))
```
tseaver: Needs `del _f` outside the `with` block.

paschembri:
I created another branch with squashed commits. You can close this pull request.

PR #235, opened 2015-01-09
Support of PYTHONSTARTUP environment variable

paschembri:
When creating the python interpreter, it will load PYTHONSTARTUP if available

### DIFF
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -1329,7 +1329,7 @@ if _interactive:
         with open(_py_startup, 'U') as _f:
             exec(compile(_f.read(), _py_startup, "exec"))
 
-    del _f
+        del _f
     del _py_startup
 
     __import__("code").interact(banner="", local=globals())
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -1292,6 +1292,7 @@ py_script_template = script_header + '''\
 
 %(relative_paths_setup)s
 import sys
+import os
 
 sys.path[0:0] = [
   %(path)s
@@ -1322,6 +1323,15 @@ if len(sys.argv) > 1:
 
 if _interactive:
     del _interactive
+
+    _py_startup = os.environ.get('PYTHONSTARTUP')
+    if _py_startup and os.path.isfile(_py_startup):
+        with open(_py_startup, 'U') as _f:
+            exec(compile(_f.read(), _py_startup, "exec"))
+
+    del _f
+    del _py_startup
+
     __import__("code").interact(banner="", local=globals())
 '''
 


paschembri:
Squashed commits from #233

tseaver:
Can you add a test and an entry in `CHANGES.rst`?

reinout:
For adding a test, look in the `easy_install.txt` next to the `easy_install.py`. Somewhere around line 500 you'll see a `>>> cat(bin, 'demo')` that prints out a generated file. Perhaps duplicate those lines and save the existing os.environ and modify it, run buildout again, check the output and restore the old os.environ.

greatdesignisnotenough:
Thanks for the info. Actually, after testing the feature and using it, I am not sure it is a good idea to load PYTHONSTARTUP within the python interpreter generated.
It is quite implicit and has no warning. So if a user has some paths in its PYTHONSTARTUP file it may go against the idea of a buildout for isolated environment.

I think adding a parameter to explicitely allow the use of PYTHONSTARTUP may be a better idea. What do you think ?

paschembri:
Thanks for the info. Actually, after testing the feature and using it, I am not sure it is a good idea to load PYTHONSTARTUP within the python interpreter generated.
It is quite implicit and has no warning. So if a user has some paths in its PYTHONSTARTUP file it may go against the idea of a buildout for isolated environment.

I think adding a parameter to explicitely allow the use of PYTHONSTARTUP may be a better idea. What do you think ?

gotcha:
@paschembri I agree with your last comment and proposal.

reinout:
(Note: I haven't used PYTHONSTARTUP myself, so I'm mostly brainstorming here).

There might still be a mismatch between PYTHONSTARTUP (something individual users might want, and then they want it for all their buildouts) and the buildout config (which is used by multiple users, and not every one of them uses PYTHONSTARTUP in a buildout-compatible way).

So switching on the option in a buildout enables it for all the users. You rather want to enable it per user and not per buildout.

Would it help to document it as something you'd generally want to enable (or disable) in `~/.buildout/default.cfg`? Instead of in a specific project's buildout? In that case, a `use_pythonstartup=true` option in the `[buildout]` part might be a good idea.

paschembri:
OK, let's do that.

paschembri:
After digging into the source files, adding an option to generate the python interpreter script requires the
modification of the 'scripts' function signature (https://github.com/buildout/buildout/blob/master/src/zc/buildout/easy_install.py#L960) to pass buildout options.

The script function signature would be changed from

    def scripts(reqs, working_set, executable, dest=None,
            scripts=None,
            extra_paths=(),
            arguments='',
            interpreter=None,
            initialization='',
            relative_paths=False,
            )
to

    def scripts(reqs, working_set, executable, dest=None,
            scripts=None,
            extra_paths=(),
            arguments='',
            interpreter=None,
            initialization='',
            relative_paths=False,
            use_pythonstartup=False,
            )

Other internal functions may be changed but with minor modifications.

Finally, using the 'initialization' variable is too limited to do something clean (no indentation possible !).

mauritsvanrees:
I close this because there was no activity since long.  Feel free to reopen if you want to pick this up again.

Issue #230, opened 2015-01-07
Tests take forever to run


tseaver:
[14 - 23 minutes for each environment on Travis](https://travis-ci.org/buildout/buildout/builds/44200649)

mgedmin:
Ideas for speeding things up:

- add `sudo: false` to .travis.yml (see http://docs.travis-ci.com/user/workers/container-based-infrastructure/)
- cache CPython builds (see http://docs.travis-ci.com/user/caching/#Arbitrary-directories)

reinout:
I'm trying out the travis caching in a branch now. Once I got it working I'll submit a pull request. That should shave 2-3 minutes off the travis build time. Running the actual tests still takes the most time, though.

reinout:
Yeah! My improvements work. An old run (https://travis-ci.org/buildout/buildout/builds/44028728 for instance) might take up to 1 hour 20 minutes in total (for all five builds) to run.

Now (https://travis-ci.org/buildout/buildout/builds/69056029) it is at half an hour. Almost three times quicker.

If we want it even quicker, we need to look at the tests themselves. More offline work. Less hitting pypi (if that happens at all). Etc.

OK to close this issue for now?

PR #247, opened 2015-06-11
Travis speed improvement through caching

reinout:
Solves #230, a little bit at least.

I adjusted the makefile: the python releases are downloaded and compiled in a directory separate from the compiled results. The compiled results are cached in travis. Keeping the cache small is essential.

The travis config now uses containers ("sudo=false"), which allows us to cache the compiled python packages. This shaves 2-3 minutes off the build.

The time is down from 8-12 minutes to 4-10 minutes. So... most of the time is still in the actual tests.

Technical note: travis doesn't (yet) distinguish caches per environment variable, like we use. So every run, when completed, tries to save its results to the cache. So one of the 5 pythons will get cached as they run at the same time... So someone (probably me) needs to re-start individual builds for individual python versions to get the cache filled for every python version. That's the best I can do now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,8 @@ cache:
   - python3.2
   - python3.3
   - python3.4
+# Note on these cache directories: "make build" downloads and installs a fresh
+# python in those directories.
 
 notifications:
   email:
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,8 @@ notifications:
     - buildout-development@googlegroups.com
 
 install:
+    - ls -al python*
+    - ls -al python*/bin/
     - deactivate
     - make build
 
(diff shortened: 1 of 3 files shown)


reinout:
Ok to merge this? The only effect is on the travis test setup.

reinout:
Note: there's one drawback to this change, namely that I had to change the locations in the makefile in order to benefit from the travis-ci.org caching. The change means the makefile has to compile python again when you first run it.

reinout:
I mailed travis-ci as their caching doesn't take into account the environment variables, like I mentioned in this pull request's introduction. I got an answer back and they're aware of it: https://github.com/travis-ci/travis-ci/issues/3023

reinout:
No comments, so I assume the fix is OK. Only drawback is a one-time local re-compile due to the necessary change in directory names.

Issue #198, opened 2014-10-31
up/downgrade fails with extension (mr.developer)


agroszer:
Having the following buildout.cfg:

```
[buildout]
extensions = mr.developer
sources-dir = dep
auto-checkout = *
parts = scripts
develop = .
versions = versions

[versions]
zc.buildout = 2.2.0

#  jedi and epc are eggs needed for jedi.  Folded into the scripts
#  part in order to reduce the number of parts that take time.
[scripts]
recipe = zc.recipe.egg
eggs = jedi
       epc
       RBTools

interpreter = python

```

results in:
```
bin/buildout
While:
  Installing.
  Loading extensions.
Error: There is a version conflict.
We already have: zc.buildout 2.2.3
```

The current zc.buildout version is 2.2.3, that gets downloaded by bootstrap.py
The problem can be fixed by either ```python bootstrap.py --version=2.2.0``` or nailing zc.buildout = 2.2.3

reinout:
See pull request #248. Buildout already has code to remove develop eggs that it knows about. The problem is the ones it doesn't know about. So in #248, I'm simply removing the develop-eggs directory in buildout's own `bootstrap()` phase.

reinout:
#248 is part of the new 2.4.0 buildout release, so I think this issue can be closed now. Feel free to re-open when there's still an error.

Note that you'd better also grab the latest bootstrap.py, as that also has a fix dealing with global paths.

PR #248, opened 2015-06-15
Bootstrap fixes, including develop-eggs removal

reinout:
Fixed bootstrap: a wrong combined try/except could lead to unintended breakage. In the end I removed the early import of setuptools altogether as it doesn't work well anymore with the newest ez_setup.py files.

The bootstrap phase of buildout itself now removes the develop-eggs directory first. I've grown tired of telling collegues to "remove the develop-eggs/ directory" when they have buildout problems. Some old bootstrap scripts could leave an incorrect `setuptools.egg-link` in there. Same with the `osc.recipe.sysegg` (there's a better version in `syseggrecipe` now that does the right thing). Anyway, removing the directory when bootstrapping is a good safety valve.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,10 @@ Unreleased
   higher for this functionality.
   [lrowe]
 
+- The bootstrap script now uses ``--buildout-version`` instead of
+  ``--version`` to pick a specific buildout version.
+  [reinout]
+
 - Updated buildout's `travis-ci <https://travis-ci.org/buildout/buildout>`_
   configuration so that tests run much quicker so that buildout is easier and
   quicker to develop.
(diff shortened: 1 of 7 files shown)


reinout:
The change to remove the develop-eggs/ directory upon buildout bootstrap is something that might need an OK by @jimfulton . Mentioning it here as he advocated a more subtle approach at the end of the long #198 thread. I can't see a reliable way to do it more subtle, as the problem is exactly the develop eggs that buildout itself doesn't know about.

reinout:
Looking at the mailinglist messages, merging and releasing is OK.

reinout:
I just got to stress-test this change on an unsuspecting colleague. Using the 2.4.0 buildout completely solved his problem, despite incorrect items in the `develop-eggs/` directory. I'm happy :-)

Issue #148, opened 2013-10-05
UnicodeDecodeError when running buildout


hvelarde:
if you have a package that contains a directory with a name encoded, you will see the following error when running buildout:

```python
(python-2.7)# hvelarde@nanovac (jenkins-csslint) ~/plone/recipe.codeanalysis 
# bin/buildout -D
Develop: '/home/hvelarde/plone/recipe.codeanalysis/.'
While:
  Installing.
Traceback (most recent call last):
  File "/home/hvelarde/.buildout/eggs/zc.buildout-2.2.1-py2.7.egg/zc/buildout/buildout.py", line 1942, in main
    getattr(buildout, command)(args)
  File "/home/hvelarde/.buildout/eggs/zc.buildout-2.2.1-py2.7.egg/zc/buildout/buildout.py", line 521, in install
    self._compute_part_signatures(install_parts)
  File "/home/hvelarde/.buildout/eggs/zc.buildout-2.2.1-py2.7.egg/zc/buildout/buildout.py", line 766, in _compute_part_signatures
    sig = _dists_sig(pkg_resources.working_set.resolve([req]))
  File "/home/hvelarde/.buildout/eggs/zc.buildout-2.2.1-py2.7.egg/zc/buildout/buildout.py", line 1633, in _dists_sig
    result.append(dist.project_name + '-' + _dir_hash(location))
  File "/home/hvelarde/.buildout/eggs/zc.buildout-2.2.1-py2.7.egg/zc/buildout/buildout.py", line 1601, in _dir_hash
    hash.update(' '.join(dirnames).encode())
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc4 in position 21: ordinal not in range(128)

Starting pdb:
> /home/hvelarde/.buildout/eggs/zc.buildout-2.2.1-py2.7.egg/zc/buildout/buildout.py(1601)_dir_hash()
-> hash.update(' '.join(dirnames).encode())
(Pdb) l
1596 	        dirnames[:] = sorted(n for n in dirnames if n not in ignore_directories)
1597 	        filenames[:] = sorted(f for f in filenames
1598 	                              if (not (f.endswith('pyc') or f.endswith('pyo'))
1599 	                                  and os.path.exists(os.path.join(dirpath, f)))
1600 	                              )
1601 ->	        hash.update(' '.join(dirnames).encode())
1602 	        hash.update(' '.join(filenames).encode())
1603 	        for name in filenames:
[...]

Issue #98, opened 2013-03-09
bin/buildout (2.0.1) fails with Unicode named directories in dist-packages


jandd:
I use zc.buildout  for a Django project on Debian (Wheezy). I just ran into trouble on a machine that has the system package python-pyramid installed. The package contains a file named `/usr/lib/python2.7/dist-packages/pyramid/tests/fixtures/static/héhé` (the accented `é`s are stored UTF-8 encoded on the file system). This leads to the following problem with running bin/buildout:

```
While:
  Installing.

An internal error occured due to a bug in either zc.buildout or in a
recipe being used:
Traceback (most recent call last):
  File "/home/jan/.buildout/eggs/zc.buildout-2.0.1-py2.7.egg/zc/buildout/buildout.py", line 1808, in main
    getattr(buildout, command)(args)
  File "/home/jan/.buildout/eggs/zc.buildout-2.0.1-py2.7.egg/zc/buildout/buildout.py", line 505, in install
    self._compute_part_signatures(install_parts)
  File "/home/jan/.buildout/eggs/zc.buildout-2.0.1-py2.7.egg/zc/buildout/buildout.py", line 749, in _compute_part_signatures
    sig = _dists_sig(pkg_resources.working_set.resolve([req]))
  File "/home/jan/.buildout/eggs/zc.buildout-2.0.1-py2.7.egg/zc/buildout/buildout.py", line 1515, in _dists_sig
    result.append(dist.project_name + '-' + _dir_hash(location))
  File "/home/jan/.buildout/eggs/zc.buildout-2.0.1-py2.7.egg/zc/buildout/buildout.py", line 1487, in _dir_hash
    hash.update(' '.join(dirnames).encode())
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1: ordinal not in range(128)
```

Uninstalling python-pyramid helps (running python -S bootstrap.py / python -S bin/buildout does not). Uninstalling python-pyramid is not really an option because there are packaged system wide available applications that require it.

idgserpro:
@reinout Shouldn't we try to close at least two issues relating unicode problems? We have this one, https://github.com/buildout/buildout/issues/148 and https://github.com/buildout/buildout/issues/152.

Fixing it would be really nice in a next release (https://github.com/buildout/buildout/issues/249). This is really, really confusing for beginners. Even an unicode named directory in ```~/.local``` breaks zc.buildout.

reinout:
[...]

PR #250, opened 2015-06-29
Buildout no longer breaks on packages with non-ascii filenames in them.

reinout:
Fixes #89 and #148.

The error only occurred on python 2, python 3 is fine. The main cause:

`os.walk()` would return filenames as non-ascii bytestrings. A `' '.join()` would fail with an error. (Or the subsequent `.encode()` call, I don't remember which of the two was an error. We now convert all filenames to unicode. The same approach is used by zest.releaser and it works fine there.

Additionally, the md5 `hash.update()` would fail with non-ascii characters. There's now an explicit call to `.encode('utf-8')`, just like with the file contents, which we already assumed could be non-ascii.

The tests run on both 2 and 3. The original test case failed on 2.x and ran just fine on 3.x the first time it ran on travis.ci, nicely demonstrating that it worked fine on python 3.

Note that there was initially an error in my implementation which was caught just fine by the test case. So the test coverage in this area is pretty good :-)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,11 @@ Unreleased
   higher for this functionality.
   [lrowe]
 
+- Buildout no longer breaks on packages that contain a file with a non-ascii
+  filename. Fixes #89 and #148.
+  [reinout]
+
+
 2.3.1 (2014-12-16)
 ==================
 
--- a/src/zc/buildout/buildout.py
+++ b/src/zc/buildout/buildout.py
@@ -1602,7 +1602,7 @@ def _open(base, filename, seen, dl_options, override, downloaded):
     return result
 
 
-ignore_directories = u'.svn', u'CVS', u'__pycache__'
+ignore_directories = '.svn', 'CVS', '__pycache__'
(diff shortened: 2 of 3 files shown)


reinout:
I could really prefer if someone with more python3 porting experience would look at it. It are just a few lines of changed code, so it shouldn't take too long.

@mgedmin, you've done a lot with python3. Could you take a quick peek?

reinout:
https://github.com/buildout/buildout/issues/98#issuecomment-117317197 validates that the fix works.

reinout:
The fix is known to work, so I'm merging it. We need to get a new release out ASAP and merging it will help getting it out.

If there's anything that might need to change regarding this fix, I suspect it is that the fix does too much. So it won't be doing anything wrong, but one or two lines might be unneeded. So if someone spots them: please point them out. For now, I'm gonna hit the "merge" button.

idgserpro:
Shouldn't we close all unicode related issues?

reinout:
@idgserpro you're right. I've closed the two bugs and left one "we need to do more unicode work" issue open.

PR #252, opened 2015-07-01
Better, simpler fix than #250

reinout:
See discussion on the distutils-dev mailinglist. @wichert said unicode shouldn't be necessary on python 2. So I dived in a bit more and discovered the fix from #250 can be made much simpler.


idgserpro:
It works. [This is the test made](https://github.com/buildout/buildout/issues/98#issuecomment-117674532).

wichert:
Looks good to me.

Issue #32, opened 2012-12-04
zc.buildout fails behind a cntlm proxy


ale-rt:
When behind a cntlm proxy [1] used to authenticate to forefront [2], zc.buildout fails downloading files as reported in https://bugs.launchpad.net/zc.buildout/+bug/484735.
The patch in the bug report fixes the problem [3].

[1] http://cntlm.sourceforge.net
[2] http://technet.microsoft.com/en-us/forefront/ee807302
[3] the patch.
```diff
diff --git a/src/zc/buildout/download.py b/src/zc/buildout/download.py
index ee6865a..c0d5c3a 100644
--- a/src/zc/buildout/download.py
+++ b/src/zc/buildout/download.py
@@ -25,6 +25,7 @@ import re
 import shutil
 import tempfile
 import urllib
+import urllib2
 import urlparse
 import zc.buildout
 
@@ -177,7 +178,10 @@ class Download(object):
 
         try:
             try:
-                tmp_path, headers = urllib.urlretrieve(url, tmp_path)
+                tmp_sock = urllib2.urlopen(url)
+                tmp_file = open(tmp_path, 'w')
+                tmp_file.write(tmp_sock.read())
+                tmp_file.close()
                 if not check_md5sum(tmp_path, md5sum):
                     raise ChecksumError(
                         'MD5 checksum mismatch downloading %r' % url)
```

stefano-m:
I have been bitten by the same issue on buildout 2.4.0 when trying to download and HTTPS url through an unauthenticated proxy.

This is likely related to the use of urllib's `urlretrieve` and `URLopener` on Python 2 that are affected by [Python issue 24599](https://bugs.python.org/issue24599), urllib2 on Python 2 does not suffer from the bug (fixed in [Python issue 1424152](https://bugs.python.org/issue1424152)).

I noticed that pull request #33 has been closed by @jimfulton and not merged. I think that a strategy to solve this problem that does not require to wait for the Python bug to be fixed and a new Python 2 release should be considered.

I know that his issue is quite old now, but I'd like to hear from the buildout developers before opening a new one.

stefano-m:
This issue should probably be closed since PR #253 that fixes it has been merged.

PR #253, opened 2015-07-10
Enable downloading behind a proxy by using urllib2

stefano-m:
Behind a proxy, `urllib` does not work with Python2 due to [Python issue
24599](https://bugs.python.org/issue24599).

Instead of using `urllib`, let's implement `urlretrieve` with `urllib2`, which
instead works and leave the Python3 implementation untouched (since that
one works too).

This fixes buildout issue #32.

There's no need to change the `try: except:IOError` block where `urlretrieve` is
used (line 195 in download.py) since the exceptions raised by `urllib2`
are subclasses of `IOError` and will then be caught.

### DIFF
--- a/src/zc/buildout/download.py
+++ b/src/zc/buildout/download.py
@@ -22,17 +22,26 @@ try:
     # Python 3
     from urllib.request import FancyURLopener, URLopener, urlretrieve
     from urllib.parse import urlparse
-    from urllib import request as urllib # for monkey patch below :(
+    from urllib import request
+
+    class URLOpener(FancyURLopener):
+        http_error_default = URLopener.http_error_default
+
+    request._urlopener = URLOpener()  # Ook! Monkey patch!
+
 except ImportError:
     # Python 2
-    from urllib import FancyURLopener, URLopener, urlretrieve
     from urlparse import urlparse
-    import urllib
 
-class URLOpener(FancyURLopener):
-    http_error_default = URLopener.http_error_default
+    import urllib2
 
-urllib._urlopener = URLOpener() # Ook! Monkey patch!
+    def urlretrieve(url, path):
+        """Work around Python issue 24599
+        """
+        url_obj = urllib2.urlopen(url)
+        with open(path, 'wb') as fp:
+            fp.write(url_obj.read())
+        return path, url_obj.info()
 
 
 from zc.buildout.easy_install import realpath
@@ -44,6 +53,7 @@ import shutil
 import sys
 import tempfile
 import zc.buildout
+
 
 class ChecksumError(zc.buildout.UserError):
     pass
--- a/src/zc/buildout/download.py
+++ b/src/zc/buildout/download.py
@@ -24,10 +24,10 @@ try:
     from urllib.parse import urlparse
     from urllib import request
 
-    class URLOpener(FancyURLopener):
+    class PatchedURLopener(FancyURLopener):
         http_error_default = URLopener.http_error_default
 
-    request._urlopener = URLOpener()  # Ook! Monkey patch!
+    request._urlopener = PatchedURLopener()  # Ook! Monkey patch!
 
 except ImportError:
     # Python 2


stefano-m:
Please note that I have successfully run `tox` on this change, albeit only on Python 2.7

stefano-m:
Is there any chance that this pull request is going to be considered?

I appreciate any comments and am available to make the required changes if needed.

leorochael:
Looks fine to me.

@jimfulton?

reinout:
Looks mostly OK to me.

There is one thing that worries me a bit: the try/except results in different imported modules being available. `URLOpener` is being imported on python3, but not anymore on python2, for instance. As long as the tests run, it is probably OK, but...

It looks like either something is being imported (in python3 for instance) that isn't necessary. Or something is missing in python2. (Or I'm overlooking something).

And the case difference between `URLOpener` and `URLopener` is very very small and easy to overlook. Perhaps it cannot be avoided, but as one of them is apparently only used in a monkey patch, perhaps a less-similar-looking name (`PatchedURLOpener` or so) is better?

So: the patch does not look immediately obviously problem-free to me. It *does* look like it solves the issue. Could you look at the issues I've raised?

- Exact same imports in the python3 and python2 part. Perhaps move an import inside one of the monkey-patch thingies?

- Name/case confusion. Perhaps just a comment, perhaps a name change.

If you fix something and I don't respond within two days, feel free to remind me to take a look :-)

stefano-m:
@reinout thanks for your comment.

* Regarding your first question:

Since  the Python issue 24599 only affects Python 2, I have tried to work around that only on the Python 2 side by redefining the `urlretrieve` function and moving the existing monkey-patching into the Python 3 code block (I also import `urllib.request` using its original name since there's no need to monkey-patch in Python 2).
The Python 2 block does not need to be monkey-patched any more because it uses a different mechanism, namely `urllib2` to emulate the broken `urlretrieve`. Hence the difference in imports.

I tried to be as conservative as possible in my changes, that's why I have essentially not modified the Python 3 code.

* Regarding your second question:

I agree that the class naming is a bit unclear, especially the `URLOpener` vs the patched `URLopener`, but I did not want to change anything more that what would've fixed the issue.

It's definitely a good thing to remove the source of confusion, but I feel that it does not belong to this PR, probably it's best done in a separate commit. That said, I'll be happy to change the names in this PR if you think it's appropriate.

Let me know what you think.

reinout:
- I'd adjust the name(s) right here in the pull request. I don't have any problems with a few extra cleanup-type lines in a pull request :-)

- Regarding different imports for python 2 and python 3: I don't know if you use pyflakes, but it warns for unused and/or missing imports. I love it, as it helps me prevent errors. The current code looks like it could cause problems as it either has **too many** or **too few imports**, depending on whether it is run on python 2 or 3. I mean: I should be able to remove the try/except code (just keeping either the python2 or python3 variant) and get a clean result without warnings out of pyflakes. As long as that's the case, I'm fine with it.

pbauer:
See also the rejected pr https://github.com/buildout/buildout/pull/33

src/zc/buildout/download.py
```diff
@@ -22,17 +22,26 @@
     # Python 3
     from urllib.request import FancyURLopener, URLopener, urlretrieve
     from urllib.parse import urlparse
-    from urllib import request as urllib # for monkey patch below :(
+    from urllib import request
+
+    class PatchedURLopener(FancyURLopener):
```
stefano-m: Renamed for clarity as requested by @reinout

stefano-m:
@reinout, I have just pushed a commit to disambiguate the monkey-patched URL opener.

On the other hand...

> Regarding different imports for python 2 and python 3: I don't know if you use pyflakes, but it warns for unused and/or missing imports. I love it, as it helps me prevent errors. The current code looks like it could cause problems as it either has too many or too few imports, depending on whether it is run on python 2 or 3. I mean: I should be able to remove the try/except code (just keeping either the python2 or python3 variant) and get a clean result without warnings out of pyflakes. As long as that's the case, I'm fine with it.

I use flake8 that in turn uses (also) pyflakes. I am not getting any warning about unused imports on both Python 2 and Python 3. To double check I have also run pyflakes directly on the `download.py` module but I didn't get any warning about imports. Am I missing something?

As far as I can see, the imports seem to be in the right blocks.

* `urllib` is only used by Python 3
* `urllib2` is only used by Python 2
* `urlparse` is used by both Python 2 and Python 3, although it must be imported from different modules

These are the warnings that I get from flake8 on `download.py`:

    Line  Col  Level   ID        Message
       61   1 warning  E302   expected 2 blank lines, found 1
      204  31 warning  E128   continuation line under-indented for visual indent

Would you mind posting the errors you get from pyflakes? That would be really helpful.

reinout:
I didn't get any pyflakes errors as I only looked at your diff visually :-) I was only worrying about the different conditional imports being used later on by others that don't realize they're in a try/except. But on second thought we don't need to worry about that. The test runner will catch that. And there's no need to complicate the code further.

Looks good, I'll merge it.

reinout:
I've added a changelog entry (directly in master). See https://github.com/buildout/buildout/commit/21344c76afdd1a5767d5af484210cddd77ada2d3

Can you check if that's the right text?

reinout:
I've added a test fix on master (https://github.com/buildout/buildout/commit/29a13cbbe24da159d226f77932ffab07c47da472, completely unrelated to your pull request) that hopefully gets the winbot tests running again. Those are scheduled in 1.5 hours.

I'm waiting for the results from that and I think I'll just make a release. Nothing mayor has changed, but waiting for a bigger change isn't needed, right?

stefano-m:
@reinout, thanks for the merge. This will help me pushing the use of buildout where I work!

I've added a comment on 21344c7 with a possible better wording.

reinout:
Hm, the windows issue is almost fixed. I'll open a separate issue for that.

I've changed the changelog a bit. Mentioning the python 2 issue is a good idea.

I've released 2.4.1. Can you check to make sure it works for you?

stefano-m:
@reinout, thanks for that. I've just realized that we should've also said that this closes issue #32 (and that issue should be closed accordingly). Of course, it's not worth re-relasing just because of that. But it would be nice to have `CHANGES.rst` mention the issue number since it's already being done.

I will update those files in my next PR. If not, feel free to remind me before merging!

Also, I will confirm the fix in the released version ASAP.

reinout:
I've closed #32 :-) I tend to forget those things sometimes.

stefano-m:
@reinout, I have downloaded the latest release and I can confirm that this PR fixes #32 

Thanks for merging it!

Issue #259, opened 2015-09-01
os.makedirs instead of os.mkdir?


guyzmo:
Hi,

I'm using the following scheme for all my buildouts:

```ini
develop = .
develop-dir        = ${buildout:directory}/var/clone/
eggs-directory     = ${buildout:directory}/var/eggs
develop-eggs-directory = ${buildout:directory}/var/develop-eggs
parts-directory    = ${buildout:directory}/var/parts
```

the issue is that when I run `buildout`, it fails loudly (as expected) because `var` does not exists. So I have to add in all my repositories a `var/.placeholder` so my `buildout` always works fine. If I'm using such a scheme, it's because I find it tidier to have only two directories being changed by buildout instead of having `clone, eggs, develop-eggs, parts…` in the source's root. It's been a few years that I've been doing that, but I wondered why I couldn't ask for getting buildout to get along more nicely with my preference.

And actually python makes it quite easy to create a full path directory by switching to `os.makedirs`. So I had a look to the sources, was about to make a PR just for that when I saw that note in the sources, at commit 735aee2d:

```python
                    # Note: os.mkdir only creates the dir if the parent
                    # exists. This is the way we want it.
```

* So what is the rationale behind that? 
* Do _we_ really want it? I actually do not! ;-)
* Why would this be any better than `os.makedirs`, and what would it be protecting against?
* would you happily accept a PR switching `os.mkdir` to `os.makedirs`, and if so, would I have to protect the users' against any edge case I wouldn't think of?

reinout:
I remember adding the "this is the way we want it" comment when I refactored that piece of code a bit. I don't remember why anymore, probably because there was a test for it somewhere. So... I replaced the `os.mkdir` with `os.makedirs` and ran all the tests.

All the tests run just fine. So I've got no hint why I added the comment in the first place.

The only reason I can think of is that buildout wants to prevent people from accidentally creating deeply nested directories in the wrong places. But... that's not really a problem, I think. Buildout prints out the directory name if it creates it.

Perhaps someone else knows the reason? Otherwise I'd say, make a pull request.

mgedmin:
+1 for os.makedirs

jimfulton:
I'm willing to defer, but I don't generally like the idea of creating multiple levels of directories.
It means that typos go unchecked.  I suppose that's less of an issue if the directories are rooted
in some known location, like the buildout directory, or are relative paths.

guyzmo:
[...]

Issue #206, opened 2014-11-06
Travis CI does not test pull requests


mgedmin:
If you submit a pull request for buildout, you'll notice that GitHub doesn't show a Travis build status for it.

You'll also notice that https://travis-ci.org/buildout/buildout/pull_requests is empty.

This reminds me of https://github.com/zopefoundation/zope.githubsupport/issues/1.  Assuming the cause of the problem is the same, the easiest way to fix this would be to go into the settings and disable then re-enable Travis CI integration.

I cannot do this because I don't have administrative access to the buildout repository.

mgedmin:
> \<J1m> I'm looking at the settings for buildout in travis and it says build pull requests is on.
> \<mgedmin> can you turn it off and then back on?
> \<J1m> sure
> \<J1m> done

Let's hope that fixes it.

mgedmin:
> Let's hope that fixes it.

Nope :(.  I just created https://github.com/buildout/buildout/pull/231 and https://travis-ci.org/buildout/buildout/pull_requests remains empty.

reinout:
Switching "enable PR builds" off and back on again didn't solve it. If I read @mgedmin's linked issues correctly, the whole travis integration would have to be switched off and back on again on the *github* side of things. That might reset the wrong setting, but I'm afraid it also resets the cache and other things.

Anyway, I've created a travis issue: https://github.com/travis-ci/travis-ci/issues/4739. I hope they either fix the button and/or enable PR builds by hand for our repo.

Doing clicky-clicky in the github settings is something we can try later on.

reinout:
Hurray, fixed by travis in https://github.com/travis-ci/travis-ci/issues/4739#issuecomment-137046189

<img width="867" alt="screen shot 2015-09-02 at 14 32 26" src="https://cloud.githubusercontent.com/assets/121433/9631211/a7b031dc-517f-11e5-9e0c-a47e156a95b4.png">

mgedmin:
\o/

PR #260, opened 2015-09-01
Changed os.mkdir to os.makedirs to make it support nested directory creation

guyzmo:
Patch to sort out issue GH-259.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,8 @@ Change History
 2.4.3 (unreleased)
 ==================
 
-- Nothing changed yet.
+- Added nested directory creation support 
+  [guyzmo]
 
 
 2.4.2 (2015-08-26)
--- a/src/zc/buildout/buildout.py
+++ b/src/zc/buildout/buildout.py
@@ -392,10 +392,8 @@ class Buildout(DictMixin):
             if cache:
                 cache = os.path.join(options['directory'], cache)
                 if not os.path.exists(cache):
-                    # Note: os.mkdir only creates the dir if the parent
-                    # exists. This is the way we want it.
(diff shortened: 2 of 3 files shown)


reinout:
Can you make it complete by adding a quick `CHANGES.rst` entry?

Are you able to get the tests running locally? Two commands ought to set you up:

    make PYTHON_VER=2.7 build
    make PYTHON_VER=2.7 test

I'd look at the 'relative path' section at the end of `downloadcache.txt` as a template for adding a sub-sub-directory. Check the presence with a "ls" as in that example.

(Otherwise just say so and I'll try and add a test).

mgedmin:
Do we not have Travis for this?

Oh.  #206.

reinout:
@mgedmin : #206 is irritating, yes.

The pull request itself runs without test errors, though. I quickly made the same change locally and ran the tests to see if I could find something.

guyzmo:
ok, so I've updated the documentation and wrote a test in `downloadcache.txt` as suggested by @reinout and added an entry in `CHANGES.rst`. 

```
  Ran 346 tests with 0 failures, 0 errors and 0 skipped in 9 minutes 56.038 seconds.
```

guyzmo:
oh! travis is back ;-) 🍻

reinout:
Yes, travis is  back :-) See #206; I fired off an issue to travis based on @mgedmin's detailed information and they fixed our setup within a few hours.

Thanks for the test and the changelog, I'll merge it and do a release.

reinout:
Ouch, I closed it instead of merging it :-)

Anyway, can you check 2.4.3 to see if it works for you as intended?

guyzmo:
it works lovely! thank you 🍺!

Issue #257, opened 2015-08-24
Download config is broken with https://user:password@host:port/myconfig.cfg


projekt01:
The latest changes in version 2.4.1 will break downloading a config file from a private pypi server.
```
bin/buildout -c https://user:password@pypi.myhost.com/myconfig.cfg
```
will end in:
```
  An internal error occurred due to a bug in either zc.buildout or in a recipe being used:
  Traceback (most recent call last):
    File "/xeebo/python27/lib/python2.7/site-packages/zc.buildout-2.4.1-py2.7.egg/zc/buildout/buildout.py", line 1986, in main
      user_defaults, command, args)
    File "/xeebo/python27/lib/python2.7/site-packages/zc.buildout-2.4.1-py2.7.egg/zc/buildout/buildout.py", line 231, in __init__
      data['buildout'].copy(), override, set()))
    File "/xeebo/python27/lib/python2.7/site-packages/zc.buildout-2.4.1-py2.7.egg/zc/buildout/buildout.py", line 1574, in _open
      path, is_temp = download(filename)
    File "/xeebo/python27/lib/python2.7/site-packages/zc.buildout-2.4.1-py2.7.egg/zc/buildout/download.py", line 117, in __call__
      local_path, is_temp = self.download(url, md5sum, path)
    File "/xeebo/python27/lib/python2.7/site-packages/zc.buildout-2.4.1-py2.7.egg/zc/buildout/download.py", line 196, in download
      tmp_path, headers = urlretrieve(url, tmp_path)
    File "/xeebo/python27/lib/python2.7/site-packages/zc.buildout-2.4.1-py2.7.egg/zc/buildout/download.py", line 41, in urlretrieve
      url_obj = urllib2.urlopen(url)
    File "/xeebo/python27/lib/python2.7/urllib2.py", line 127, in urlopen
      return _opener.open(url, data, timeout)
    File "/xeebo/python27/lib/python2.7/urllib2.py", line 404, in open
      response = self._open(req, data)
    File "/xeebo/python27/lib/python2.7/urllib2.py", line 422, in _open
      '_open', req)
    File "/xeebo/python27/lib/python2.7/urllib2.py", line 382, in _call_chain
      result = func(*args)
    File "/xeebo/python27/lib/python2.7/urllib2.py", line 1222, in https_open
      return self.do_open(httplib.HTTPSConnection, req)
    File "/xeebo/python27/lib/python2.7/urllib2.py", line 1153, in do_open
[...]

PR #263, opened 2015-09-07
Disabled patching of FancyURLopener in the download.py if used Python>=3.3.

Cykooz:
Starting with version Python 3.3 the module **urllib.request** no longer contains the global variable **_urlopener**.

### DIFF
--- a/src/zc/buildout/download.py
+++ b/src/zc/buildout/download.py
@@ -25,12 +25,9 @@ try:
     from urllib.parse import urlparse
     from urllib import request
 
-    class PatchedURLopener(FancyURLopener):
-        http_error_default = URLopener.http_error_default
-
-    if sys.version_info[:2] >= (3, 3):
-        request.install_opener(PatchedURLopener())
-    else:
+    if sys.version_info[:2] < (3, 3):
+        class PatchedURLopener(FancyURLopener):
+            http_error_default = URLopener.http_error_default
         request._urlopener = PatchedURLopener()  # Ook! Monkey patch!
 
 except ImportError:
--- a/src/zc/buildout/download.py
+++ b/src/zc/buildout/download.py
@@ -13,6 +13,7 @@
 ##############################################################################
 """Buildout download infrastructure"""
 
+import sys
 try:
     from hashlib import md5
 except ImportError:
@@ -27,7 +28,10 @@ try:
     class PatchedURLopener(FancyURLopener):
         http_error_default = URLopener.http_error_default
 
-    request._urlopener = PatchedURLopener()  # Ook! Monkey patch!
+    if sys.version_info[:2] >= (3, 3):
+        request.install_opener(PatchedURLopener())
+    else:
+        request._urlopener = PatchedURLopener()  # Ook! Monkey patch!
 
 except ImportError:
     # Python 2
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,13 +4,13 @@ Change History
 2.4.4 (unreleased)
 ==================
 
-- Added support for Python>=3.3 into the download.py.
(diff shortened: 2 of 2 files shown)


reinout:
Weird that the test cases didn't find this? 

Hm. I see that the travis checks for this PR failed due to the ez_setup/bootstrap issue of earlier today. Could you add a quick `CHANGES.rst` entry? That ought to run the tests again :-)

reinout:
@stefano-m, you wrote https://github.com/buildout/buildout/pull/253 and this PR apparently fixes something for python 3.3. Does this look good to you?

Cykooz:
Hmm, something went wrong :)

reinout:
@Cykooz: it are the python 3.3+ builders that are failing, so they're using your if/else.

Perhaps the signature of the new method you're calling has changed? It returns something that needs to get `.read()` called on it, for instance?

Did you run the test locally? (See `DEVELOPERS.txt` in the buildout root for the two "make" commands that you need to run in case you didn't spot that yet).

Cykooz:
Well, I think I rushed to conclusions. In the case of Python> = 3.3 just do not need to patch anything.
I corrected my code.
```bash
$ make PYTHON_VER=3.4 build
$ make PYTHON_VER=3.4 test
...
Ran 346 tests with 0 failures, 0 errors and 0 skipped in 8 minutes 12.942 seconds.
```
PS:
In fact, I was trying to figure out how to fix *lovely.buildouthttp* for use it with *zc.buildout>=2.4.1*. During the study, I noticed that patch used in the *download.py*   does not change anything if used with *Python>=3.3*.

src/zc/buildout/download.py
```diff
@@ -24,10 +25,10 @@
...
     from urllib import request
 
-    class PatchedURLopener(FancyURLopener):
-        http_error_default = URLopener.http_error_default
-
-    request._urlopener = PatchedURLopener()  # Ook! Monkey patch!
+    if sys.version_info[:2] < (3, 3):
+        class PatchedURLopener(FancyURLopener):
+            http_error_default = URLopener.http_error_default
+        request._urlopener = PatchedURLopener()  # Ook! Monkey patch!
```
stefano-m: I am not sure that this change is needed at all. If `_urlopener` is not used any more in Python's 3.3 `request`, adding it should be innocuous.

Cykooz:
@stefano-m, Ok, I agree with you.

Cykooz:
What would you say if I integrates the functionality of *lovaly.buildouthttp*  into the *zc.buildout*?

reinout:
lovely.buildouthttp is for other authentication schemes, it seems? On a first look, it seems like something that could be better done as an extension, like it is now. That's just a first look, it might be something we would like to add, so don't mind me too much :-)

A second note: https://github.com/buildout/buildout/issues/257, that's about an error with the python 3 url parsing regarding username/password. It just might be that you've got some code lying aroudn that could fix it.

Regarding lovely.buildouthttp integration: if you're really looking into that, it is probably best to open a separate issue for it.

Cykooz:
I use *lovely.buildouthttp*  because they let me do not store in the buildout.cfg the credentials for private PyPi-server. Without this functionality, I have to save the username and password into the repository in this form:
```
find-links = http://login:password@mypypi.server.com/simple/private.package
```

reinout:
I've got the same sort of line in some of my own buildouts, I recognize them.

stefano-m:
@Cykooz, integrating `lovely.buildouthttp` is not for me to decide as I am just an occasional contributor.
I think that @jimfulton or @reinout would be able to advise you.

In my opinion, I don't see any rush in integrating the functionality since it's provided by the extension.

PR #265, opened 2015-10-01
Support universal wheels

graingert:


mgedmin:
:+1: LGTM

reinout:
It *is* a bit weird to release buildout as a wheel if it itself doesn't support wheels. But this way buildout is at least dipping a single toe into the wheel waters :-)

I think there's no drawback to doing this. Right?

mgedmin:
setuptools is packaged as a universal wheel, even though easy_install doesn't support installing wheels.

I see no drawbacks.

reinout:
I'll add a changelog entry and make a new release.

reinout:
@graingert, can you test if 2.4.4 works for you as a wheel?

Actually, we should probably add wheel support to zc.recipe.egg, too, right?

reinout:
Ok, I've gone ahead and made a pull request: #266 

Can you take a look there just to make sure?

PR #270, opened 2015-10-11
fix for win32 '#!"python"' vs '#!python' shebang

agroszer:
fixes the 2 failures at

http://winbot.zope.org/builders/zc_buildout_dev%20py_270_win32%20master/builds/1144/steps/test/logs/stdio

### DIFF
--- a/src/zc/buildout/tests.py
+++ b/src/zc/buildout/tests.py
@@ -3539,6 +3539,10 @@ def test_suite():
                  "Unused options for buildout: 'scripts' 'eggs'."),
                 # Python 3.4 changed the wording of NameErrors
                 (re.compile('NameError: global name'), 'NameError: name'),
+                # fix for test_distutils_scripts_using_import_are_properly_parsed
+                # and test_distutils_scripts_using_from_are_properly_parsed
+                # win32 apparently adds a " around sys.executable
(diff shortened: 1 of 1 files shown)


reinout:
Looks fine. Can you report back whether it worked? Or: when do the winbot tests run? I thought every night or so?

agroszer:
yep nightly

agroszer:
OMG, it's green:

http://winbot.zope.org/builders/zc_buildout_dev%20py_270_win32%20master/builds/1145

agroszer:
I might be tempted to send the failure emails of that builder to a list (distutils?)

reinout:
Well... I've got an alternative to spamming distutils.

It would be valuable if the winbot would test pull requests. Is that handy to arrange?

Note: I don't mind having to log in and press a button or something like that if it is handy, especially for PRs that I expect to be a bit windows-unfriendly.

In exchange for giving me some button I can press (instead of waiting for a nightly build), I don't mind reinout@vanrees.org being put in the "it breaks" email list :-)

mgedmin:
If we used Jenkins, we could have plugins for setting GitHub commit status and also plugins for testing GitHub pull requests.

Unfortunately we use buildbot.

agroszer:
@reinout I added your e-mail addr

Issue #24, opened 2012-10-15
Buildout does not install setup.scripts, even with zc.recipe.egg:scripts


specialunderwear:
http://pypi.python.org/pypi/zc.recipe.egg/1.3.2#script-generation does not work.

You can verify this with the following test package: http://pypi.python.org/pypi/tompoes/0.0.2.

It seems zc.buildout.easy_install isn't doing it's work properly.

There is one more issue, which might or might not be related. In any case, the example package has only 1 module; tompoes.py. This module is NOT installed. That means the endpoint `tom` isn't working!

Since I've got no idea if this bug is caused by the failure to install scripts, I added it to this ticket instead of creating a separate issue.

specialunderwear:
I found another bug, while it now seems to work properly with eggs installed from pypi, it does not work when installed from a source::
```
[sources]
tompoes = git file:///Users/henk/Sites/tompoes

[tompoes]
recipe = zc.recipe.egg:scripts
eggs =
    tompoes
dependent-scripts = true

**Comment by tseaver** [comment]
@specialunderwear, the `[sources]` section is only meaningful to the [`mr.developer` extension](https://pypi.python.org/pypi/mr.developer), and is used to tell the extension where to find *remote* repositories.  You don't need it for "local" ones -- I would spell your config above as:
```ini
[buildout]
develop =
    /Users/henk/Sites/tompoes
parts =
    tompoes

[tompoes]
recipe = zc.recipe.egg:scripts
eggs =
    tompoes
dependent-scripts = true
```

specialunderwear:
Good point, it could be a bug in mr.developer. Needless to say the same problem occurs with real remote sources. I will test if the script is properly installed as a local source, the way you demonstrate.

tseaver:
Can you post a gist of the *full* buildout.cfg that reproduces your problem?  Ideally, it would point to a git checkout from a public repository.

specialunderwear:
Also does not work with

```
[buildout]
develop =
    /Users/henk/Sites/tompoes
parts =
    tompoes

[tompoes]
recipe = zc.recipe.egg:scripts
eggs =
    tompoes
dependent-scripts = true
```

Will push tompoes to github and create a minimal buildout.cfg

specialunderwear:
Look here: https://github.com/specialunderwear/tompoes
I created several buildout configurations and a test script that tests them all.

Only the egg variant install the script.

tseaver:
[...]

PR #273, opened 2015-10-14
distutils scripts for develoment eggs are also installed

reinout:
See the last half of #24. Eggs installed as develop eggs or via mr.developer dont' get their distutils scripts generated. Now they do.

Main problem: `setup.py develop` doesn't record metadata on those scripts, so we have to keep track of them ourselves. Sadly, I seem to have to do it in a module-level variable. The "tompoes" testcase-example from the #24 works with this change. And there's a test case.


src/zc/buildout/easy_install.py
```diff
@@ -882,6 +883,45 @@ def _copyeggs(src, dest, suffix, undo):
...
+    dir_contents = os.listdir(directory)
+    egginfo_filenames = [filename for filename in dir_contents
+                         if filename.endswith('.egg-link')]
+    if not egginfo_filenames:
+        return
+    egg_name = egginfo_filenames[0].replace('.egg-link', '')
+    marker = 'EASY-INSTALL-DEV-SCRIPT'
+    scripts_found = []
+    for filename in dir_contents:
+        dev_script_content = open(os.path.join(directory, filename)).read()
```
mgedmin: I suggest
```python
with open(...) as fp:
    dev_script_content = fp.read()
```
to avoid ResourceWarnings on Python 3.x.

Also, is it possible for this directory to contain subdirectories or other kinds of files that cannot be read (e.g. broken symlinks)?  If so, maybe this needs a `try:`/`except:`.

src/zc/buildout/easy_install.py
```diff
@@ -882,6 +883,45 @@ def _copyeggs(src, dest, suffix, undo):
...
+        if marker in dev_script_content:
+            # The distutils bin script points at the actual file we need.
+            for line in dev_script_content.splitlines():
+                match = DUNDER_FILE_PATTERN.search(line)
+                if match:
+                    actual_script_filename = match.group('filename')
+                    actual_script_content = open(
+                        actual_script_filename).read()
```
mgedmin: with open()...

mgedmin:
LGTM otherwise.

I'm a bit curious about the dance with filename/__file__ assignment parsing/actual_script_filename, and I wish there was a comment explaining that, but I guess it doesn't matter -- I'm sure there's a good reason to do it that way.

src/zc/buildout/easy_install.py
jamadden: `with` is also important for PyPy, even for reads. PyPy can run out of file descriptors before doing a garbage collection (depending on the specifics, of course)

reinout: Ok, I'll add a "with" here. @mgedmin: I'll also add an `os.path.isfile()` check. It shouldn't be necessary, but I *did* have to add such a check elsewhere already because of some python 3 `__pycache__` or so directory.

reinout:
@mgedmin: the actual_script_filename is because there is an intermediary file. In the "tompoes" example repo that I've tested with, you ought to get a "tom_main" console script and a "tom" distutils script.

At some point, buildout generates a develop egg and tells setuptools to put the output in some temp dir. I looked there and you've got a .egg-link, a "tom" and a "tom_main" file. Normally, buildout only copies the .egg-link file to the develop-eggs/ directory.

The "tom" file is a four-line thing with `__package__ = 'package-version'` and `__file__ = /path/to/the/actual/script' and lastly some weird `exec(...__package__, __file__ ...)` line. So I extract the name of the actual script from the `__file__` line.

That's the only place I could find the name of the distutils script! It isn't inserted in the metadata, sadly.

PR #274, opened 2015-10-15
Fixed winbot test. Didn't think about -script and .exe

reinout:
See http://winbot.zope.org/builders/zc_buildout_dev%20py_270_win64%20master/builds/1149/steps/test/logs/stdio

@agroszer: this is the way to fix it, right? Or is there a handier way?

Note: the emails I'm getting help.


mgedmin:
I've seen platform differences papered over with regexps in the doctest RENormalizer before.  I think I prefer your approach.

reinout:
Well, it is my copied-over-from-somewhere-else-in-the-tests approach :-)

agroszer:
LGTM

PR #275, opened 2015-10-17
Editing while reading

jean:


src/zc/buildout/buildout.txt
```diff
@@ -1088,10 +1088,10 @@ Extending sections (macros)
 ---------------------------
 
 A section (other than the buildout section) can extend one or more
-other sections using the ``<`` option.  Options from the referenced
```
freddrake: Describing this as the ``<`` option is correct; the ``=`` comes from the ``option = value`` structure.

src/zc/buildout/buildout.txt
```diff
@@ -1264,10 +1264,10 @@ Expressions must not contain either the ``#`` or the ``;`` character.
 Adding and removing options
 ---------------------------
 
-We can append and remove values to an option by using the + and -
+We can append and remove values to an option by using the ``+`` and ``-``
 operators.
 
-This is illustrated below; first we define a base configuration.
+This is illustrated below; first we define a base configuration::
```
freddrake: Changing this to ``::`` isn't strictly necessary since the example shown is a doctest and is detected properly.

src/zc/buildout/buildout.txt
jean: Ah OK, I'll rethink this and clarify or remove the change.

src/zc/buildout/buildout.txt
jean: I have to admit that the `::` is for Vim's restructured-text syntax highlighting to work.

jean: Well, Vim's highlighting reveals the issue, but the fact is that:
> A text block that is indented relative to the preceding text, without preceding markup indicating it to be a literal block or other content, is a block quote. [Block Quotes](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#block-quotes)

So I do think marking code blocks as correct RST (literal blocks or using [directives](http://sphinx-doc.org/markup/index.html#sphinxmarkup)) has value.

freddrake: I agree.  While I generally prefer using ``::`` myself, I wanted to point out that the change wasn't strictly necessary since it's already recognized as a Doctest block:

http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#doctest-blocks

(Thinking about ReST while editing in MD makes me sad.  Silly GitHub.)

jean: > http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#doctest-blocks

Oh! I wasn't aware that this is part of ReST. I managed to miss that
part of the docs. So Vim's syntax file should be fixed.

> (Thinking about ReST while editing in MD makes me sad. Silly GitHub.)

Agreed completely ...

-- 
jean                                              . .. .... //\\\oo///\\

reinout:
Thanks for the tidying. I apparently missed this PR, sorry for not merging earlier.

PR #277, opened 2015-10-24
Make relative_paths option work when path is equal to the relative path.

youngking:
For example when you use the develop option like:

  [buildout]
  develop= .

### DIFF
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -1152,6 +1152,8 @@ def _relative_path(common, path):
 def _relativitize(path, script, relative_paths):
     if path == script:
         raise AssertionError("path == script")
+    if path == relative_paths:
+        return "base"
     common = os.path.dirname(os.path.commonprefix([path, script]))
     if (common == relative_paths or
         common.startswith(os.path.join(relative_paths, ''))


src/zc/buildout/easy_install.py
```diff
@@ -1152,6 +1152,8 @@ def _relative_path(common, path):
 def _relativitize(path, script, relative_paths):
     if path == script:
         raise AssertionError("path == script")
+    if path == relative_paths:
+        return "base"
```
reinout: I'd expect `base` instead of the string `"base"`.

reinout:
Your comment is not completely clear to me. Can you explain what goes wrong exactly?

I think your fix might be handy, though it looks like there's an error in your fix.

- Could you add a changelog entry? That should immediately make it clear which error is being fixed :-)

- Could you add a small test that proves your fix works? It doesn't have to be a very elaborate test. In any case, a test would probably demonstrate the small error of always returning `"base"` instead of `base`.

youngking:
zc.recipe.egg has an option `relative-paths` which allows a buildout to be moved without breaking egg paths.  When a `buildout.cfg` file live with  a project,  it will always use a `develop` option like this:

```
[buildout]
develop = .
```

But then the generated script will include the  realpath of the current project,for example:

```
...
sys.path[0:0] = [
  '/Users/xxx/projects/someproject',
  join(base, '.buildout/eggs/gunicorn-19.3.0-py2.7.egg'),
  join(base, '.buildout/eggs/tox-2.1.1-py2.7.egg'),
  join(base, '.buildout/eggs/pluggy-0.3.0-py2.7.egg'),
  ...
  ]
...
```
The first path make the project non-portable, so I would like it to be:

```
sys.path[0:0] = [
  base,
  join(base, '.buildout/eggs/gunicorn-19.3.0-py2.7.egg'),
  join(base, '.buildout/eggs/tox-2.1.1-py2.7.egg'),
  join(base, '.buildout/eggs/pluggy-0.3.0-py2.7.egg'),
  ...
  ]
...
```  

This fix is ok,  the `_relativitize ` function should return a path as a string. I had tested it on my local machine. It seems this `buildout` project only has doctest, I am not very accustomed to it and can't  run it locally, I'll try it again.

src/zc/buildout/easy_install.py
reinout: Ah, you're right. Below there's `"join(base, %r)"`, so `"base"` is indeed the correct syntax.

reinout:
I'll try to make a doctest for it tomorrow.

In case you want to try it yourself: this should let you run the tests:

    $ make PYTHON_VER=2.7 build
    $ make PYTHON_VER=2.7 test

(The first step is only needed once, btw).

gotcha:
I think I never mentioned it anywhere, but there is another fix needed around relative_paths: setuptools and zc.buildout egg paths are not relativized.

youngking:
@gotcha It seems that if you use the systemwide zc.buildout, then the `site-packages` will be included in the generated script's sys.path. But if you use the `bootstrap.py`  to generate a buildout script under the bin directory, it will be clean.

In our company, we had customed a buildout script like [this](https://github.com/youngking/buildout-package/blob/master/bin/buildout), and make a deb package  based on it to distribute to  every production machine. Use this customed buildout command, the finally generated script will not include the site-packages.And the `setuptools` and `zc.buildout` will all be in the relative path.

reinout:
I've merged your PR into #278, where I added a small test, so I'm closing this one. @youngking, can you take a look?

Note: the next time you make a PR, please make a branch first. That's clearer.

youngking:
Ok, thank you. There's another problem, the tests doesn't work under the mac os x, it will try installed a python version without compiled openssl, although openssl-1.0.2d_1 is installed on my mac.But it works on debian if I had installed the libssl-dev first, so this is not a big problem.

reinout:
I've got it running on OSX. If you use the makefile instructions I copied above (they're also in `DEVELOPERS.txt`), you get a freshly compiled empty python. Should work on OSX, too.

PR #278, opened 2015-10-26
Same as #277, but with test and changelog entry

reinout:
This one supersedes #277. The fix is the same, but I added a test and changelog entry.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,9 @@ Change History
 2.4.6 (unreleased)
 ==================
 
-- Nothing changed yet.
+- Relative paths are now also correctly generated for the current directory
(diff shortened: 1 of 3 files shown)


youngking:
I should say #195 is the same too, I didn't notice it earlier.

reinout:
@youngking: you're right, that's a similar fix you fond in #195!

reinout:
Anyone: does it look good enough to merge?

mgedmin:
LGTM.

reinout:
@youngking: I just releaser 2.4.6, can you check if that one works for you?

youngking:
It works for me, thanks.

PR #284, opened 2015-12-16
Trying to fix #283

### LINKED ISSUE #283: TypeError: Option values must be strings with setuptools 19.1
Since todays `19.1` release of `setuptools` buildout fails like this:

```
$ bin/buildout
...
While:
  Installing.

recipe being used:
  File "/var/lib/jenkins/zope/eggs/zc.buildout-2.5.0-py2.7.egg/zc/buildout/buildout.py", line 1992, in main
  File "/var/lib/jenkins/zope/eggs/zc.buildout-2.5.0-py2.7.egg/zc/buildout/buildout.py", line 565, in install
    self._compute_part_signatures(install_parts)
  File "/var/lib/jenkins/zope/eggs/zc.buildout-2.5.0-py2.7.egg/zc/buildout/buildout.py", line 811, in _compute_part_signatures
    options['__buildout_signature__'] = ' '.join(sig)
  File "/var/lib/jenkins/zope/eggs/zc.buildout-2.5.0-py2.7.egg/zc/buildout/buildout.py", line 1370, in __setitem__
    raise TypeError('Option values must be strings', value)
TypeError: ('Option values must be strings', u'zc.recipe.egg-1.3.2-py2.7.egg setuptools-19.1-py2.7.egg zc.buildout-2.5.0-py2.7.egg')
```

Not sure whether this is an issue with `zc.buildout`, `setuptools`, or something else.

reinout:
The problem is also present on travis. So if the pull request works, that is enough of a test for me. Because if it doesn't fix the issue, we cannot run the tests :-)

**Rapid** code review requested as it is an issue that breaks many peoples' automatic tests and many peoples' buildouts because bootstrap doesn't work.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,16 @@ Change History
 2.5.1 (unreleased)
 ==================
 
-- Nothing changed yet.
+- Fix for #283: setuptools 19.1's unicode package names don't break buildout anymore.
+
+  The problem is that setuptools 19.1 apparently returns unicode package names
+  on python 2 instead of strings as previously. Buildout expects strings on
+  python 2. So I copied an `encode()` if from somewhere else in buildout that
+  fixes it.
+
+  Also see
+  https://bitbucket.org/pypa/setuptools/issues/476/setuptools-191-breaks-zcbuildout
+  [reinout]
 
 
 2.5.0 (2015-11-16)
--- a/src/zc/buildout/buildout.py
+++ b/src/zc/buildout/buildout.py
@@ -808,6 +808,8 @@ class Buildout(DictMixin):
             recipe, entry = _recipe(options)
             req = pkg_resources.Requirement.parse(recipe)
             sig = _dists_sig(pkg_resources.working_set.resolve([req]))
+            if isinstance(sig, text_type):
+                sig = sig.encode()
             options['__buildout_signature__'] = ' '.join(sig)
 
     def _read_installed_part_options(self):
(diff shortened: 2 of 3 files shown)


lukasgraf:
Looks good to me - your latest change fixes the issue for me when testing locally.

Not sure if I'm a fan of encoding the unicode string with the default system encoding though - we know (or assume) it should only ever contain ASCII, why not make this assertion explicit? I don't have any strong feelings about this though.

reinout:
I would have tried to think a bit more about default encodings, normally. But I didn't in this case as I could simply copy some code from `buildout.py` that fixes the same kind of issue. One dealing with directory hashes. If that one doesn't break, this one probably is also OK.

Especially as the problem now is a `u"19.1"` :-)

lukasgraf:
Hmm. From the test failures it looks like setting option values really is just the first place where issues with this surfaced. I'm considering whether this shouldn't be fixed in setuptools instead.

lukasgraf:
@reinout FYI, looks like @jaraco just committed a fix for this in setuptools: https://bitbucket.org/pypa/setuptools/commits/6df29b1f97d0

reinout:
Yes, I saw the fix in setuptools. A fix there sure is better. It is hard to verify if the fix really fixes the issue as I weirdly don't see the issue locally (only in the buildout tests and it is close to impossible to mock/hack the setuptools egg there).

lukasgraf:
I could somewhat verify his fix by downloading the updated `ez_setup.py` and have [`bootstrap.py` pick it up from there](https://github.com/buildout/buildout/blob/master/bootstrap/bootstrap.py#L89) when bootstrapping again. That seemed to work for me (previously failing buildout ran through).

reinout:
See https://travis-ci.org/buildout/buildout/jobs/91337683, it has been fixed by setuptools by adjusting the `ez_setup.py` script.

I'm retracting the pull request as it isn't needed anymore, hurray!

PR #291, opened 2016-04-06
Turn proposed patch into a pull request.

gyst:
Refs #257. Provide a pull request for patch provided in
https://github.com/buildout/buildout/issues/257#issuecomment-159423299

### DIFF
--- a/src/zc/buildout/download.py
+++ b/src/zc/buildout/download.py
@@ -31,17 +31,28 @@ try:
 
 except ImportError:
     # Python 2
+    import base64
     from urlparse import urlparse
-
+    from urlparse import urlunparse
     import urllib2
 
-    def urlretrieve(url, path):
-        """Work around Python issue 24599
+    def urlretrieve(url, tmp):
+        """Work around Python issue 24599 includig basic auth support
         """
-        url_obj = urllib2.urlopen(url)
-        with open(path, 'wb') as fp:
+        scheme, netloc, path, params, query, frag = urlparse(url)
+        auth, host = urllib2.splituser(netloc)
+        if auth:
+            url = urlunparse((scheme, host, path, params, query, frag))
+            req = urllib2.Request(url)
+            base64string = base64.encodestring(auth)[:-1]
+            basic = "Basic " + base64string
+            req.add_header("Authorization", basic)
+        else:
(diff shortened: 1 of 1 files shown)


src/zc/buildout/download.py
```diff
@@ -31,17 +31,28 @@ class PatchedURLopener(FancyURLopener):
...
-        """Work around Python issue 24599
+    def urlretrieve(url, tmp):
+        """Work around Python issue 24599 includig basic auth support
         """
-        url_obj = urllib2.urlopen(url)
-        with open(path, 'wb') as fp:
+        scheme, netloc, path, params, query, frag = urlparse(url)
+        auth, host = urllib2.splituser(netloc)
+        if auth:
+            url = urlunparse((scheme, host, path, params, query, frag))
```
reinout: The `url = urlunparse(...)` and `req = urllib2.Request(url)` can be moved outside the "if" so that they're only needed once, right?

The "if auth" would then only need to compose and add the header. Little bit shorter that way.

src/zc/buildout/download.py
```diff
@@ -31,17 +31,28 @@ class PatchedURLopener(FancyURLopener):
...
     # Python 2
+    import base64
     from urlparse import urlparse
-
+    from urlparse import urlunparse
     import urllib2
 
-    def urlretrieve(url, path):
-        """Work around Python issue 24599
+    def urlretrieve(url, tmp):
```
reinout: Why have you renamed "path" to "tmp"? It might be the right decision, but I'd like to know the reason :-)

gyst:
Sorry @reinout this is literally just a copy-paste from the patch @projekt01 contributed in
https://github.com/buildout/buildout/issues/257#issuecomment-159423299

I got hit by the same bug and had to downpin zc.buildout. I have no decisions and reasons other than wanting to get this fixed :-)

reinout:
I looked at the piece of code where it is used and it is called `tmp_path` there :-) I'll merge and release.

reinout:
Note that I don't think a specific test is needed for this. The download function is tested well enough. I don't think a specific test for logging in is needed. We'd probably need to adjust the test web server for that anyway...

PR #296, opened 2016-05-25
Fix for #283: unicode package names don't break buildout anymore

ridha:


tseaver:
According to @reinout, the PyPA folks fixed #283  with a [new setuptools release](https://bitbucket.org/pypa/setuptools/issues/476/setuptools-191-breaks-zcbuildout).

ridha:
Looks like #283 either reappeared or wasn't ever fixed.

jimfulton:
A test that demonstrates that the fix is needed (and correct) would be help.

mauritsvanrees:
I close this because there was no activity since long.  Feel free to reopen if you want to pick this up again and if a fix is still needed.

PR #304, opened 2016-07-12
No python 3.3 and newer versions of the other pythons

do3cc:


reinout:
Hm. I might have to look at the travis caching I set up a while ago. The makefile places the compiled python into a "2.7" dir, so it doesn't differentiate between 2.7.3 and 2.7.11...

reinout:
I've zapped the cache and I've restarted the test runs. This will take a while :-)

reinout:
So, the cache is filled again and everything's OK.

PR #308, opened 2016-08-18
When removing an egg directory gives an OSError, retry once.

### LINKED ISSUE #307: Parallel buildout runs can break a new egg
When you run two buildouts next to each other and they both install the same distribution that you did not have before, then this can lead to a broken egg.
This assumes eggs as directories and not as zipped files, but that is the default.
What happens, is this:

- Both buildouts create the egg in their own temporary directory.
- The first buildout starts moving the files of the egg to the `eggs/package.egg` directory.
- The second buildout is slightly later, sees that `eggs/package.egg` already exists, and starts to remove it.
- The first buildout was not quite finished moving all the files, so it keeps doing that.
- The second buildout has removed all the files that it saw and tries to remove the directory, but this fails with an `OSError` because the directory is no longer empty. It does not do anything fancy here, it simply fails while calling `shutil.rmtree`.
- The first buildout may or may not continue, but it will leave behind an egg directory with only a part of the files.
- Subsequent buildout runs see that the egg is there, so they happily continue. But anything that tries to use the egg, will run into problems like `ImportErrors` or other missing files.

I see this happen too often on the Continuous Integration server of Plone: http://jenkins.plone.org. We have four nodes (machines). Each node has three jobs or executors. Each job runs a buildout. All jobs on a node share the same egg cache to prevent using three times as much disk space.
When I make a change in Plone to use a new package, multiple buildouts are started and they may end up on the same node. Most of the time this goes fine. But too often we run into the situation I describe above.

A sample traceback you get is this, which is what happens in the second buildout:

```
While:
  Installing robot.
  Getting distribution for 'Products.CMFEditions==2.2.21'.

recipe being used:
  File "/home/jenkins/.buildout/eggs/zc.buildout-2.5.2-py2.7.egg/zc/buildout/buildout.py", line 1995, in main
  File "/home/jenkins/.buildout/eggs/zc.buildout-2.5.2-py2.7.egg/zc/buildout/buildout.py", line 666, in install
  File "/home/jenkins/.buildout/eggs/zc.buildout-2.5.2-py2.7.egg/zc/buildout/buildout.py", line 1410, in _call
    return f()
  File "/home/jenkins/.buildout/eggs/zc.recipe.egg-2.0.3-py2.7.egg/zc/recipe/egg/egg.py", line 126, in install
    reqs, ws = self.working_set()
  File "/home/jenkins/.buildout/eggs/zc.recipe.egg-2.0.3-py2.7.egg/zc/recipe/egg/egg.py", line 84, in working_set
    allow_hosts=self.allow_hosts)
  File "/home/jenkins/.buildout/eggs/zc.buildout-2.5.2-py2.7.egg/zc/buildout/easy_install.py", line 883, in install
  File "/home/jenkins/.buildout/eggs/zc.buildout-2.5.2-py2.7.egg/zc/buildout/easy_install.py", line 664, in install
    for_buildout_run=for_buildout_run):
  File "/home/jenkins/.buildout/eggs/plone.versioncheck-1.3-py2.7.egg/plone/versioncheck/tracking.py", line 22, in get_dist
    dists = old_get_dist(self, requirement, *ags, **kw)
  File "/home/jenkins/.buildout/eggs/zc.buildout-2.5.2-py2.7.egg/zc/buildout/easy_install.py", line 552, in _get_dist
    dist.location, ws, self._dest, dist)
  File "/home/jenkins/.buildout/eggs/zc.buildout-2.5.2-py2.7.egg/zc/buildout/easy_install.py", line 403, in _call_easy_install
    shutil.rmtree(newloc)
  File "/usr/lib/python2.7/shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "/usr/lib/python2.7/shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "/usr/lib/python2.7/shutil.py", line 256, in rmtree
    onerror(os.rmdir, path, sys.exc_info())
  File "/usr/lib/python2.7/shutil.py", line 254, in rmtree
    os.rmdir(path)
OSError: [Errno 39] Directory not empty:
'/home/jenkins/.buildout/eggs/Products.CMFEditions-2.2.21-py2.7.egg/Products/CMFEditions'
```

See for example http://jenkins.plone.org/job/plone-5.1-python-2.7-robot/707/console

In the first buildout it may lead to this traceback:

```
  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/usr/lib/python2.7/py_compile.py", line 181, in <module>
    sys.exit(main())
  File "/usr/lib/python2.7/py_compile.py", line 173, in main
    compile(filename, doraise=True)
  File "/usr/lib/python2.7/py_compile.py", line 106, in compile
    with open(file, 'U') as f:
IOError: [Errno 2] No such file or directory:
'/home/jenkins/.buildout/eggs/Products.CMFEditions-2.2.21-py2.7.egg/
Products/CMFEditions/UniqueIdHandlerTool.py'
While:
  Installing robot.
  Getting distribution for 'Products.CMFEditions==2.2.21'.
```

See for example http://jenkins.plone.org/job/plone-5.1-python-2.7/693/console

Result: someone has to login to the node with ssh, [remove the broken egg](https://github.com/plone/jenkins.plone.org/issues/188), and restart a job. And it leads to me proposing [crazy workarounds](https://github.com/plone/jenkins.plone.org/pull/166) for our Jenkins server to avoid running multiple buildouts at the same time. I would rather fix it in zc.buildout.

The first traceback points to a place where we may be able to fix this. The `shutil.rmtree` call fails. I propose that we catch this error, wait five seconds, then try it again. If it still fails, we crash.

I will make a pull request, but wanted to write down the problem first.

mauritsvanrees:
This avoids problems when running two buildouts at the same time, when they try to install the same new egg.
Fixes #307.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,10 @@ Change History
 2.5.3 (unreleased)
 ==================
 
+- When removing an egg directory gives an OSError, retry once.  This
+  avoids problems when running two buildouts at the same time, when
+  they try to install the same new egg.  Fixes #307.  [maurits]
+
 - Removed Python 2.6 and 3.2 support.
   [do3cc]
 
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -34,6 +34,7 @@ import shutil
 import subprocess
 import sys
 import tempfile
+import time
 import zc.buildout
 import warnings
 
@@ -400,7 +401,22 @@ class Installer:
                 newloc = os.path.join(dest, os.path.basename(d.location))
                 if os.path.exists(newloc):
                     if os.path.isdir(newloc):
-                        shutil.rmtree(newloc)
+                        try:
+                            shutil.rmtree(newloc)
+                        except OSError:
+                            # This can happen when running two buildouts in
+                            # parallel.  Or it is simply a permission error.
+                            # So we retry once.  See
+                            # https://github.com/buildout/buildout/issues/307
+                            seconds = 5
+                            logger.warn(
+                                "OSError while removing %s.\nThis can happen "
+                                "when two buildouts run at the same time. "
+                                "Or it is a permission error. "
+                                "Retrying once in %d seconds.",
+                                newloc, seconds)
+                            time.sleep(seconds)
+                            shutil.rmtree(newloc)
                     else:
                         os.remove(newloc)
                 os.rename(d.location, newloc)


mauritsvanrees:
I have rebased this on pull request #309 which fixes an unrelated test failure.

tseaver:
@mauritsvanrees I just merged #309.

mauritsvanrees:
Thanks. I have rebased again for clarity. Previous Travis check passed.

leorochael:
Sorry guys, but I think this is the wrong strategy. I was writing a long explanation about it when this was merged, so let me just state quickly the problem I see:

`shutil.rmtree()` is a non atomic operation. Retrying it once will just risk causing another race condition later.

Instead we should question whether `_call_easy_install()` should be calling `rmtree()` in the first place. If the package it is trying to install wasn't there before, but suddenly is there now, it should just refrain from installing, instead of erasing.

mauritsvanrees:
@leorochael This pull request is not merged yet. Tres was talking about another pull request I created today, which I temporarily included in this one to fix the tests.

mauritsvanrees:
True, if the directory already exists, we could probably accept it and continue. A warning may be nice. But that is probably a good alternative. I may try that out.

mauritsvanrees:
Let me close this one in favour of #310.

PR #310, opened 2016-08-18
Do not remove an existing egg.

### LINKED ISSUE #307: Parallel buildout runs can break a new egg

mauritsvanrees:
When installing an egg to a location that already exists, keep the current location (directory or file).

This can only happen when the location at first did not exist and this changed during the buildout run.  We used to remove the previous location, but this could cause problems when running two buildouts at the same time, when they try to install the same new egg.

Fixes #307.
This is an alternative for pull request #308 , addressing the concern that @leorochael raised there.

### DIFF
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -397,7 +397,7 @@ class Installer:
 
             result = []
             for d in dists:
-                newloc = _move_to_eggs_dir(d, dest)
+                newloc = _move_to_eggs_dir_and_compile(d, dest)
                 [d] = pkg_resources.Environment([newloc])[d.project_name]
                 result.append(d)
 
@@ -518,8 +518,7 @@ class Installer:
 
                 if dist.precedence == pkg_resources.EGG_DIST:
                     # It's already an egg, just fetch it into the dest
-                    newloc = _move_to_eggs_dir(dist, self._dest)
-                    redo_pyc(newloc)
+                    newloc = _move_to_eggs_dir_and_compile(dist, self._dest)
 
                     # Getting the dist from the environment causes the
                     # distribution meta data to be read.  Cloning isn't
@@ -532,7 +531,6 @@ class Installer:
                     dists = self._call_easy_install(
                         dist.location, ws, self._dest, dist)
                     for dist in dists:
-                        redo_pyc(dist.location)
                         if for_buildout_run:
                             # ws is the global working set and we're
                             # installing buildout, setuptools, extensions or
@@ -772,9 +770,6 @@ class Installer:
                 dists = self._call_easy_install(
                     base, pkg_resources.WorkingSet(),
                     self._dest, dist)
-
-                for dist in dists:
-                    redo_pyc(dist.location)
 
                 return [dist.location for dist in dists]
             finally:
@@ -1547,8 +1542,10 @@ class IncompatibleConstraintError(zc.buildout.UserError):
 IncompatibleVersionError = IncompatibleConstraintError # Backward compatibility
 
 
-def _move_to_eggs_dir(dist, dest):
+def _move_to_eggs_dir_and_compile(dist, dest):
     """Move distribution to the eggs destination directory.
+
+    And compile the py files, if we have actually moved the dist.
 
     Its new location is expected not to exist there yet, otherwise we
     would not be calling this function: the egg is already there.  But
@@ -1604,6 +1601,10 @@ def _move_to_eggs_dir(dist, dest):
                 "We will accept it.\n"
                 "If this contains a wrong package, please remove it yourself.",
                 newloc)
+        else:
+            # There were no problems during the rename.
+            # Do the compile step.
+            redo_pyc(newloc)
     finally:
         # Remember that temporary directories must be removed
         shutil.rmtree(tmp_dest)
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -397,16 +397,8 @@ class Installer:
 
             result = []
             for d in dists:
-                newloc = os.path.join(dest, os.path.basename(d.location))
-                if os.path.exists(newloc):
-                    if os.path.isdir(newloc):
-                        shutil.rmtree(newloc)
-                    else:
-                        os.remove(newloc)
-                os.rename(d.location, newloc)
-
+                newloc = _move_to_eggs_dir(d, dest)
                 [d] = pkg_resources.Environment([newloc])[d.project_name]
-
                 result.append(d)
 
             return result
@@ -526,18 +518,7 @@ class Installer:
 
                 if dist.precedence == pkg_resources.EGG_DIST:
                     # It's already an egg, just fetch it into the dest
-
-                    newloc = os.path.join(
-                        self._dest, os.path.basename(dist.location))
-
-                    if os.path.isdir(dist.location):
-                        # we got a directory. It must have been
-                        # obtained locally.  Just copy it.
-                        shutil.copytree(dist.location, newloc)
-                    else:
-                        setuptools.archive_util.unpack_archive(
-                            dist.location, newloc)
-
+                    newloc = _move_to_eggs_dir(dist, self._dest)
                     redo_pyc(newloc)
 
                     # Getting the dist from the environment causes the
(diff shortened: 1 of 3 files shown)


src/zc/buildout/easy_install.py
```diff
@@ -399,11 +400,11 @@ def _call_easy_install(self, spec, ws, dest, dist):
...
-                    if os.path.isdir(newloc):
-                        shutil.rmtree(newloc)
-                    else:
-                        os.remove(newloc)
-                os.rename(d.location, newloc)
+                    logger.warn(
+                        "Path %s already exists.\nIf this contains a wrong "
+                        "package, please remove it yourself.", newloc)
+                else:
+                    os.rename(d.location, newloc)
```
leorochael: While the change of another buildout doing its own the `os.rename()` between this buildout calling `exists()` and then `rename()` is slim, it could still happen.

I suggest instead just doing the `os.rename()` first, then if it fails, check that the destination exists (and perhaps try creating a `pkg_resources.Environment([newloc])` on it and checking it), and then give up with a warning. Else raise an error.

leorochael: Speaking of non-atomic ops, the function `_get_dist()` below calls `shutil.copytree()` directly into the destination when the source of a dist is an egg directory.

This is a rare situation (when was the last time you saw an unpacked **egg** outside of the buildout be used as the `find-links` of a package in the buildout?), but if it happens, one buildout could see that an egg exists and attempt to use it before another buildout is finished copying it.

I suggest the `copytree()` be done in a `tempdir()` inside the egg directory and then `os.rename()`d like `_call_easy_install()` does here.

freddrake: On Thu, Aug 18, 2016 at 2:19 PM, Leonardo Rochael Almeida
<notifications@github.com> wrote:
> This is a rare situation (when was the last time you saw an unpacked egg outside of the
> buildout be used as the find-links of a package in the buildout?), but if it happens, one
> buildout could see that an egg exists and attempt to use it before another buildout is
> finished copying it.

Rare, perhaps, but useful.  At Zope (back in the day), we used a
directory of built eggs as a cache when building packages for use by
OS packaging tools.  Reading something while someone else was writing
it wasn't a situation we ran into, but certainly could have happened
had we applied a little more parallelism (whether automated or human).


  -Fred

-- 
Fred L. Drake, Jr.    <fred at fdrake.net>
"A storm broke loose in my mind."  --Albert Einstein

mauritsvanrees: See also discussion at #307 . I am updating and amending the pull request to reflect this discussion.

src/zc/buildout/easy_install.py
```diff
@@ -1564,3 +1540,72 @@ class IncompatibleConstraintError(zc.buildout.UserError):
...
+    See discussion at https://github.com/buildout/buildout/issues/307
+
+    We return the new location.
+    """
+    # First make sure the destination directory exists.  This could suffer from
+    # the same kind of race condition as the rest: if we check that it does not
+    # exist, and we then create it, it will fail when a second buildout is
+    # doing the same thing.
+    try:
+        os.makedirs(dest)
```
mauritsvanrees: I added this `makedirs` line to avoid a failure in a test that calls easy_install with a destination directory that does not exist: https://github.com/buildout/buildout/blob/2.5.2/src/zc/buildout/easy_install.txt#L209-L223
Seems a harmless side effect of this pull request.

leorochael: Won't this fail if the directory does already exist?

```
>>> import os
>>> os.makedirs('/tmp')
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 17] File exists: '/tmp'
```

leorochael: Nevermind, the code works nonetheless

leorochael:
:+1: LGTM

src/zc/buildout/easy_install.py
mauritsvanrees: Yes, it is surrounded by a try/except OSError.

mauritsvanrees:
I have rebased this due to a conflict in `CHANGES.rst`. It's been ready for a while. :-)

reinout:
I've released 2.5.3, tnx for the fix.

PR #311, opened 2016-08-19
In the tests, filter out an invisible control character.

mauritsvanrees:
I am getting a few test failure on my laptop.  Mac OSX 10.11.6.

The first one is this, made clearer by adding `doctest: +REPORT_UDIFF`:

```
File ".../src/zc/buildout/debugging.txt", line 70, in debugging.txt
Failed example:
    print_(system(buildout+" -D", """\
    up
    p sorted(self.options.keys())
    q
    """), end='')  # doctest: +REPORT_UDIFF
Differences (unified diff with -expected +actual):
    @@ -1,5 +1,5 @@
     Develop: '/sample-buildout/recipes'
     Installing data-dir.
    -> buildout.py(NNN)__getitem__()
    +> buildout.py(NNN)__getitem__()
     -> raise MissingOption("Missing option: %s:%s" % (self.name, key))
     (Pdb) > /sample-buildout/recipes/mkdir.py(NNN)install()
```

It looks like no real difference. When inspecting this, it turns out that I get a few special characters at the front of the line:

```
Installing data-dir.\n\x1b[?1034h> /.../src/zc/buildout/buildout.py(1369)__getitem__()\
```

This could be fixed this by adding ellipsis in the doctest in front of that line.

The second test that goes wrong is this one:

```
File "/.../zc/recipe/egg/README.rst", line 215, in README.rst
Failed example:
    print_(system(join(sample_buildout, 'bin', 'py-demo'),
    """import os, sys
    for p in sys.path:
        if 'demo' in p:
            _ = sys.stdout.write(os.path.basename(p)+'\\n')

    """).replace('>>> ', '').replace('... ', ''), end='')
    # doctest: +ELLIPSIS +NORMALIZE_WHITESPACE
Expected:
    demo-0.2-pyN.N.egg
    demoneeded-1.1-pyN.N.egg
Got:
    demo-0.2-pyN.N.egg
    demoneeded-1.1-pyN.N.egg
    <BLANKLINE>
    <BLANKLINE>
```

That one also looks strange, as the extra blank lines should be handled just fine by the normalize whitespace.  This turns out to be the same problem as the first one.  The string is:

```
\x1b[?1034hdemo-0.2-py2.7.egg\ndemoneeded-1.1-py2.7.egg\n\n\n
```

Adding ellipsis here is tricky because it is needed on the first line, which would confuse the doctest into thinking it is part of the command that should be executed.

According to this page, this is a control character that sets the 8-bit input mode:
http://stackoverflow.com/questions/21135991/

It may be better to strip this character sequence in the `print_` function:

```
def print_(*args, **kw):
    sep, end, file = _print_options(**kw)
    if file is None:
        file = sys.stdout
    result = sep.join(map(str, args))+end
    # This might contain an invisible control character '\x1b[?1034h'
    # which is very annoying when comparing output in doctests:
    # a difference is reported but you don't see it.
    # So we remove it.
    result = result.replace('\x1b[?1034h', '')
    file.write(result)
```

But this may have unwanted effects in real buildout runs, instead of only helping in the tests.  So in the end I decided to fix it in the regular expressions in the tests.
Note that one of the changes is in zc.recipe.egg, so this would need another release eventually.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@ Change History
 2.5.3 (unreleased)
 ==================
 
+- In ``zc.buildout.testing.system``, set ``TERM=dumb`` in the environment.
+  This avoids invisible control characters popping up in some terminals,
+  like ``xterm``.  Note that this may affect tests by buildout recipes.
+  [maurits]
+
 - Removed Python 2.6 and 3.2 support.
   [do3cc]
 
--- a/src/zc/buildout/testing.py
+++ b/src/zc/buildout/testing.py
@@ -110,12 +110,18 @@ def clean_up_pyc(*path):
 MUST_CLOSE_FDS = not sys.platform.startswith('win')
 
 def system(command, input='', with_exit_code=False):
+    # Some TERMinals, expecially xterm and its variants, add invisible control
+    # characters, which we do not want as they mess up doctests.  See:
+    # https://github.com/buildout/buildout/pull/311
+    # http://bugs.python.org/issue19884
+    env = dict(os.environ, TERM='dumb')
     p = subprocess.Popen(command,
                          shell=True,
                          stdin=subprocess.PIPE,
                          stdout=subprocess.PIPE,
                          stderr=subprocess.PIPE,
-                         close_fds=MUST_CLOSE_FDS)
+                         close_fds=MUST_CLOSE_FDS,
+                         env=env)
     i, o, e = (p.stdin, p.stdout, p.stderr)
     if input:
         i.write(input.encode())


leorochael:
@mauritsvanrees have you figured out what is actually producing this escape sequence? From the code above It seems `print_` is receiving it from `args`, which in this case come from `system()`.

Maybe the right fix here is either filtering it on `system()` itself, or figuring out if there is something we could change in the environment of the subprocesses spawned by `system()` to stop them from needing to print that escape character in the first place.

It could be something in the shell that `system()` spawns to execute the command.

I agree filtering it on `print_` might not be the best idea, unless we add another option like `+FILTER_ESCAPE_SEQUENCES`.

Alternatively, we could consider it part of `+NORMALIZE_WHITESPACE`.

But I'd prefer we find a way to stomp it at the source.

leorochael:
Anyway, I think your idea of filtering it in the normaliser is not a bad workaround.

mauritsvanrees:
I wonder if it is simply an artifact of my Mac terminal.

@reinout Do you see the same test failures on your Mac?

leorochael:
@mauritsvanrees it could well be an artifact of a specific terminal, perhaps fiddling with the `$TERM` environment variable might cause some difference in output.

If that's the case, it is at the very least an interaction between the terminal and either the shell that `system()` uses to spawn the final process or the final process spawned by `system()` itself (i.e. it's stdin handling library).

In any case I think it's useful to try to isolate where this is coming from.

reinout:
@mauritsvanrees: I don't have the error. (iterm2, bash). 

I googled a little bit and, to my surprise, I apparently got bitten by this myself in 2009: http://reinout.vanrees.org/weblog/2009/08/14/readline-invisible-character-hack.html

Can you check if $TERM is set to `xterm` in your terminal? 

See also http://bugs.python.org/issue19884, they're looking into it.

mauritsvanrees:
Aha. I have `TERM=xterm-256color`. When I set it to `iterm2` like you have, the tests on master pass.  On the branch they pass too, so my patch may be fine to include.

leorochael:
Given the new info, I would prefer a patch that set the TERM environment variable of processes spawned by buildout tests, so I'm kind of -0 on the current PR.

zc.recipe.egg_/src/zc/recipe/egg/tests.py
```diff
@@ -54,6 +54,8 @@ def test_suite():
                (re.compile('[d-]  setuptools-[^-]+-'), 'setuptools-X-'),
                (re.compile(r'eggs\\\\demo'), 'eggs/demo'),
                (re.compile(r'[a-zA-Z]:\\\\foo\\\\bar'), '/foo/bar'),
+                # control character '\x1b[?1034h' is added on some computers
+                (re.compile(r'\x1b\[\?1034h'), ''),
```
leorochael: Consider adding the link to the Python issue about readline writing these characters to the comment.

jimfulton:
+1 on setting TERM

mauritsvanrees:
I have modified this pull request to set `TERM` to `dumb` when it currently is `xterm` (or a variant with `xterm` in the name).
I am not sure about the exact `TERM` to take, but this one seems fine.  Both `dumb` and `linux` tp the trick for me.

mauritsvanrees:
I have rebased and squashed for clarity.

The original commits are in https://github.com/mauritsvanrees/buildout/tree/remove-control-character-from-print-orig if anyone needs them.

src/zc/buildout/testing.py
```diff
@@ -110,12 +110,20 @@ def clean_up_pyc(*path):
 MUST_CLOSE_FDS = not sys.platform.startswith('win')
 
 def system(command, input='', with_exit_code=False):
+    env = dict(os.environ)
+    if 'xterm' in env.get('TERM', ''):
+        # xterm and its variants add invisible control characters,
+        # which we do not want as they mess up doctests.  See:
+        # https://github.com/buildout/buildout/pull/311
+        # http://bugs.python.org/issue19884
+        env['TERM'] = 'dumb'
```
leorochael: What do you think about always setting it to `dumb` regardless of the original terminal?

To me it would be simpler and less surprising. I keep imagining someone else a few months or years from now pulling their hair out trying to figure out why subprocesses of buildout tests only get `dumb` terminals under some circumstances but not others...

Besides, a line like this feels oddly satisfying:

    env = dict(os.environ, TERM='dumb')

reinout: Is setting `TERM=dumb` OK on other OSs? Especially windows?

mauritsvanrees: Seems like it would be fine on Windows. At least cygwin used to set it to dumb: http://kizzx2.com/blog/index.php/2009/01/07/cygwin-warning-terminal-is-not-fully-functional/
That then showed a warning when you do `git diff`, saying the terminal is not fully functional, but I have the same when I try that on OSX.
Other links possibly of interest:
http://stackoverflow.com/questions/28415681/how-to-set-term-environment-variable-in-windows-so-that-it-works-with-bash-via-s
http://stackoverflow.com/questions/7949956/why-does-git-diff-on-windows-warn-that-the-terminal-is-not-fully-functional
Note that on OSX when I say `export TERM=asldkjfasdlfaskdfjha` all still seems fine. :-) So unknown TERM settings don't usually completely break anything.

But in the end it is probably easier to use the same TERM setting for everyone.
Also: this is only in `testing.py` so it is not used when running buildout normally.
I will update the pull request.

CHANGES.rst
```diff
@@ -4,6 +4,10 @@ Change History
 2.5.3 (unreleased)
 ==================
 
+- In the tests, set ``TERM=dumb`` in the environment.  This avoids
+  invisible control characters popping up in some terminals, like
+  ``xterm``.  [maurits]
+
```
leorochael: I suggest clarifying that this happens under the ' system()'  function used in tests.

mauritsvanrees: Right, `system` is used in the tests of recipes, so others might be affected by this, so they need to know.
I will update the text.
I have run the `collective.recipe.backup` tests with this branch and they pass.

mauritsvanrees:
Does anyone want to have a go at merging this? :-)

PR #315, opened 2016-11-04
Test pypy and pypy3

gforcada:


reinout:
The two new python versions fail on the Makefile:
  
    Makefile:19: *** Please specify desired PYTHON_MINOR for Python pypy3.  Stop.

I don't know how pypy works with major/minor versions?

jimfulton:
It would be nice to get rid of the Makefile.

I wonder if we can drop dev.py's refusal to work with a local setuptools, now that setuptools is stable/boring.

icemac:
This could be the starting point to resolve #342.

PR #323, opened 2017-01-28
Update to work with setuptools 34

jimfulton:


jamadden:
I've read this, and I don't see anything that obviously jumps out at me as bad/wrong, but I'm not really competent enough in this area to give an actual review.

jimfulton:
Thanks.

agroszer:
winbot seems to like it

PR #327, opened 2017-01-31
Fixes #326

### LINKED ISSUE #326: Buildout 3.6.0 always offers add itself and setuptools into the versions section.
Buildout 3.6.0 always offers add itself and setuptools into the versions section if the show-picked-versions is true.
For example:
```
# buildout.cfg
[buildout]
show-picked-versions = true
versions = versions

[versions]
setuptools = 33.1.1
zc.buildout = 2.6.0
```
Console:
```
$ ./bin/buildout
...
Versions had to be automatically picked.
The following part definition lists the versions picked:
[versions]
setuptools = 33.1.1
zc.buildout = 2.6.0
```

jimfulton:
Fixed a bug introduced in 2.6.0:
zc.buildout and its dependeoncies were reported as picked even when
their versions were fixed in a ``versions`` section.  Worse, when the
``update-versions-file`` option was used, the ``versions`` section was
updated needlessly on every run.

### DIFF
--- a/src/zc/buildout/tests.py
+++ b/src/zc/buildout/tests.py
@@ -2408,12 +2408,12 @@ def wont_downgrade_due_to_prefer_final():
     ... parts =
     ... ''')
 
-    >>> [v] = [l.split('= >=', 1)[1].strip()
+    >>> [v] = [str(l.split('= >=', 1)[1].strip())
     ...        for l in system(buildout+' -vv').split('\n')
     ...        if l.startswith('zc.buildout = >=')]
-    >>> v == pkg_resources.working_set.find(
+    >>> v == str(pkg_resources.working_set.find(
     ...         pkg_resources.Requirement.parse('zc.buildout')
-    ...         ).version
+    ...         ).version)
     True
 
     >>> write('buildout.cfg',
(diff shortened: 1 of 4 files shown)


src/zc/buildout/tests.py
```diff
@@ -2408,12 +2408,12 @@ def wont_downgrade_due_to_prefer_final():
     ... parts =
     ... ''')
 
-    >>> [v] = [l.split('= >=', 1)[1].strip()
+    >>> [v] = [str(l.split('= >=', 1)[1].strip())
```
jamadden: I'm curious why the `str` was needed in these two places?

src/zc/buildout/tests.py
```diff
@@ -3062,6 +3062,39 @@ def test_abi_tag_eggs():
...
+    >>> _ = system(join('bin', 'buildout'))
+    >>> with open('versions.cfg') as f:
+    ...     versions = f.read()
+    >>> _ = system(join('bin', 'buildout'))
+    >>> _ = system(join('bin', 'buildout'))
+    >>> _ = system(join('bin', 'buildout'))
+    >>> with open('versions.cfg') as f:
+    ...     versions == f.read()
+    True
+    >>> cat('versions.cfg') # doctest: +ELLIPSIS
```
jamadden: I assume there's no risk of ELLIPSIS hiding extra lines getting appended to the file.

src/zc/buildout/tests.py
jimfulton: Because I got a spurious failure when one was unicode and the other a str in Python 2,

src/zc/buildout/tests.py
jamadden: Never mind, that doesn't even make sense given the above equality test.

jimfulton: The check above makes sure that versions.cfg didn't change after the first run.  The check on the 3090 sanity checks that something was added on the first run.  I'll rearrange this to make it clearer.

src/zc/buildout/tests.py
jamadden: Hmm. Well that seems odd. I don't blame you for fixing the test here though.

jimfulton: It's really odd, now that you mention it. :/

So odd I now can't reproduce it. WTF? reverted.

src/zc/buildout/tests.py
jamadden: Thanks, I do find that much clearer.

jimfulton:
OK, I made the logic for controlling "picking" more explicit.

Riding ``dest`` was mostly OK mut could have some edge-case failures in offline mode, where dest is also None.

Buildout logic has evolved over the years, often under duress, and entropy has set in.

Thanks for fighting entropy. :)

jamadden:
I figured this was a duress situation so I was trying not to be too picky :) 

Thanks for that change, that makes it a lot more clear to me.

PR #331, opened 2017-02-09
windows: scripts are implemented with 2 files, avoid burping on that

### LINKED ISSUE #330: on windows metadata_listdir('scripts') returns .exe's too
see:
https://github.com/buildout/buildout/blob/master/src/zc/buildout/easy_install.py#L1106

in turn ```dist.get_metadata('scripts/' + name)``` fails to read that binary .exe

see:
http://winbot.zope.org/builders/ZODB_dev%20py_340_win32/builds/201/steps/buildout/logs/stdio

agroszer:
would fix #330

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,10 @@ Change History
 2.7.2 (unreleased)
 ==================
 
-- Nothing changed yet.
+- Fixed a bug (windows, py 3.4)
+  When processing metadata on "old-style" distutils scripts, .exe stubs
+  appeared in ``metadata_listdir``, in turn reading those burped with
+  ``UnicodeDecodeError``. Skipping .exe stubs now.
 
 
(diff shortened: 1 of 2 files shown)


jimfulton:
What is the user visible impact of this?

Could you updated CHANGES.rst?

agroszer:
@jimfulton winbot burped on this:

http://winbot.zope.org/builders/ZODB_dev%20py_340_win32/builds/203/steps/buildout/logs/stdio

reinout:
Those two files: they have been generated this way for ages. Do you know what has changed to result in an error now?

agroszer:
@reinout Actually I haven't looked at winbot for a while, it was also down for a while, so no clue since when it was failing. Maybe it's setuptools as it gets upgraded to the latest.

PR #332, opened 2017-02-13
Added a hook to enable a soon-to-be-released buildout extension

jimfulton:
to provide wheel support.

### DIFF
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -83,6 +83,9 @@ setuptools_path = buildout_and_setuptools_path
 FILE_SCHEME = re.compile('file://', re.I).match
 DUNDER_FILE_PATTERN = re.compile(r"__file__ = '(?P<filename>.+)'$")
 
+def wheel_to_egg(dist, dest):
+    raise zc.buildout.UserError("Wheels are not supported")
+
 class _Monkey(object):
     def __init__(self, module, **kw):
         mdict = self._mdict = module.__dict__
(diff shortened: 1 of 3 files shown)


src/zc/buildout/easy_install.py
```diff
@@ -513,6 +516,9 @@ def _get_dist(self, requirement, ws, for_buildout_run=False):
                     raise zc.buildout.UserError(
                         "Couldn't download distribution %s." % avail)
 
+                if dist.location.endswith('.whl'):
+                    dist = wheel_to_egg(dist, tmp)
```
leorochael: In case the download cache is activated, `tmp` here **is** the download cache.

The result is that the download cache now has the egg AND the wheel at the same time. And the egg gets overwritten with the contents of the wheel every time it's written.

Humpty will gladly do it for you, so you'll never notice the constant overwriting, but it bit me trying to replace humpty with `wheel install` with an exception like:

    ValueError: Wheel file zest/releaser/tests/gitsvn.txt would overwrite /path/to/buildout/downloads/dist/zest.releaser-6.9-py2.7.egg/zest/releaser/tests/gitsvn.txt. Use force if this is intended

An actual, separate, temporary directory needs to be used in this case.

jimfulton: That's probably the best approach. An alternative would be to reduce the precedence of wheels to 2, in which case the egg would be used next time, an advantage being that it requires less of a code change. :)

Note that this is largely moot. Anyone who sets up an egg cache is likely to set up a shared egg directory, so they're not going to be constantly re-processing the wheel.

Issue #336, opened 2017-02-15
upgrade fixes


jimfulton:
We often check for new buildout and setuptools versions before deciding that we're not able to upgrade.  This is wasteful.

- We should decide sooner that we can't upgrade and not check for new versions if we can't.

- We should not be willing to upgrade any eggs if the existing versions are outside the eggs directory.  This should fix #312.

- We should consider buildout and all of its dependencies (buildout, setuptool and all of the new and future (wheel) setuptools dependencies.

leorochael:
What if, during a buildout run, we only check dependencies of buildout itself and bail if they don't match? 

We can tell the user to run `buidout bootstrap` to fix this situation and let bootstrap deal with making sure the right versions of buildout and dependencies are set.

This way we never upgrade nor restart buildout during its run. If the user wants to upgrade buildout or dependencies she can run `buildout bootstrap` again with unpinned versions.

This means bootstrap would have to read the buildout configuration, to download the correct versions of any pinned dependencies.

jimfulton:
@leorochael so if someone updates a buildout dependency and it's not pinned, buildout should fail? I don't like that idea. :)  It also doesn't really solve a problem afaict. After all, that upgrade could fail, just as current ones do sometimes and the user is just as stuck as they are now.

The new docs encourage people to not use a local buildout (and explain the tradeoff involved). What's emerging is an explicit way to control automatic upgrade:

- If you don't want auto-upgrade, use a non-local buildout.  This is what I do, mainly because I'm too lazy to bootstrap my projects.  (I'm veeeery lazy.) I only feel upgrading breakage when I read about it in the news. :)

- If you want the latest of everything, or want to pin everything, use a local buildout and let buildout make it so.

mdeguzis:
Would this fix situations where something requires setuptools in a high version? 

```
Error: There is a version conflict. 
We already have: setuptools 0.9.8 
but zc.buildout 2.9.2 requires 'setuptools>=8.0'
```

jimfulton:
@mdeguzis I don't understand what scenario you're describing.

mdeguzis:
If I have `newest = true` and do not pin setuptools, eggs / extensions like Mr. Developer fail the deployment since they request other versions it seems:

Log: https://gist.github.com/mdeguzis/2452391c0fbc6f4717268f41224f7f83
buildout.cfg: https://gist.github.com/mdeguzis/abf18dc9dd1c65d6d7ba411c85187cbb

[...]

PR #335, opened 2017-02-15
Start on new docs

jimfulton:
Making this a PR because:

- I want to make the request for review explicit.

- My itch to fix docs is so strong, I'm working pretty hard on this. I'm almost finished with a first draft on getting started.

- I see no harm on getting this on master, as the old doc content is preserved (such as it is) in a separate directory.

PR #337, opened 2017-02-15
draft getting started

jimfulton:
Don't merge this.

I'd like to get feedback on the text. Once the text is close, I'll add manuel tests for the examples.


doc/topics/index.rst
```diff
@@ -6,3 +6,8 @@ Buildout Topics
    :maxdepth: 2
 
 .. todo:
+
+   variables-extensing-and-substitutions
```
tseaver: Should that be "extending" or "extensions"?

doc/getting-started.rst
```diff
@@ -0,0 +1,565 @@
...
+  A directory that provides a default location for installed parts.
+
+Buildout configuration files use an `INI syntax
+<https://en.wikipedia.org/wiki/INI_file>`_ [#configparser]_.
+Configuration is arranged in sections, beginning with section names in square
+brackets. Section options are names, followed by equal signs, followed
+by values.  Values may be continued over multiple lines as long as the
+continuation lines start with whitespace.
+
+Buildout is all about building stuff and the stuff to be built are
```
tseaver: "stuff is" or "things are"?

doc/topics/index.rst
jimfulton: extending. thanks

doc/getting-started.rst
jimfulton: I changed this to things

doc/getting-started.rst
```diff
@@ -0,0 +1,565 @@
...
+
+   - A particular application of Buildout, which has a buildout
+     configuration.
+
+     We'll use lower case to refer to these.
+
+   - A ``buildout`` section in a Buildout configuration (in a
+     particular buildout).
+
+     We'll use a lowercase fix-width font for these.
```
leorochael: `fix-width` :arrow_right: `fixed-width` or `monospaced`

doc/getting-started.rst
```diff
@@ -0,0 +1,565 @@
...
+<https://en.wikipedia.org/wiki/INI_file>`_ [#configparser]_.
+Configuration is arranged in sections, beginning with section names in square
+brackets. Section options are names, followed by equal signs, followed
+by values.  Values may be continued over multiple lines as long as the
+continuation lines start with whitespace.
+
+Buildout is all about building things and the things to be built are
+specified using *parts*.  The parts to be built are listed in the
+``parts`` option.  For each part, there must be a section with the same
+name that names the software to build the part and optionally provides
```
leorochael: `name that names` sounds weird...
How about `...that specifies the software to build the part`.

Also, I think `optionally` is not strictly necessary here.

doc/getting-started.rst
```diff
@@ -0,0 +1,565 @@
...
+
+  buildout
+
+Then a number of things will happen:
+
+- ``zc.recipe.egg`` will be downloaded and installed in your ``eggs``
+  directory.
+
+- ``ZEO`` and its dependencies will be downloaded and installed. (ZEO
+  is a small Python database server.)
```
jimfulton: In this example, the directories should all be empty. Maybe I should say that, but I don't want to be caught in a lie. :) There may be eggs in the eggs directory because of an aborted upgrade, see https://github.com/buildout/buildout/issues/336.

Maybe I should show some partial listings later, when there should be things there. I'm not sure this would help. I'd like more opinions, but barring any I'll defer. :)

WRT windows, no-can-do.  If we decide ``bin`` listings are helpful I'd be happy to include a listing, but I'd need someone to contribute one.

jimfulton:
@tseaver thank you for the comments!

doc/getting-started.rst
```diff
@@ -0,0 +1,565 @@
...
+      -a 127.0.0.1:8200
+
+Here we've added a new ``server`` part that uses ``zc.zdaemonrecipe``.
+We used a ``program`` option to define what program should be run.
+There are a couple of interesting things to note about this option:
+
+- We used :doc:`variable substitutions
+  <topics/variables-extending-and-substitutions>`:
+
+  ``${buildout:directory}``
```
leorochael: Both of these variables are buildout internal variables, not visible in the `[buildout]` section above.

Wouldn't it be better to introduce variable substitute with concrete variables defined in some section, and then explain that both buildout and the part recipes can add implicit variables of their own?

doc/getting-started.rst
tseaver: > In this example, the directories should all be empty.

Won't there be eggs for `zc.recipe.egg`, `ZEO`, `ZODB`, etc.?

doc/getting-started.rst
```diff
@@ -0,0 +1,565 @@
...
+  the eggs directory.
+
+- A ``server`` script is added to the ``bin`` directory.  This script
+  is generated by the recipe.  It can be run like::
+
+    bin.server start
+
+  to start a server and::
+
+    bin.server stop
```
leorochael: Isn't it `bin/server stop` and `bin/server start` instead of `bin.server ...`?

doc/getting-started.rst
```diff
@@ -0,0 +1,565 @@
...
+      transcript /Users/jim/t/0214/parts/server/transcript.log
+    </runner>
+
+    <eventlog>
+      <logfile>
+        path /Users/jim/t/0214/parts/server/transcript.log
+      </logfile>
+    </eventlog>
+
+  The **details aren't important**, other than the fact that the script
```
leorochael: > other than the fact that the ~script~ _configuration file_ reflect the actual buildout location and options coming from the part.

doc/getting-started.rst
```diff
@@ -0,0 +1,565 @@
...
+application affect other applications, including buildouts. This can
+lead to unexpected errors.   This is why it's recommended to use a
+`virtual environment <https://virtualenv.pypa.io/en/stable/>`_ or a
+"clean python" built from source with no third-party packages
+installed.
+
+(It's a little hypocritical to recommend installing Buildout into an
+otherwise clean environment, which is why Buildout provides a
+:doc:`bootstrapping mechanism <topics/bootstrapping>` which allows
+setting up a buildout without having to sully a virtual environment or
```
leorochael: > without having to sully a virtual environment...

On first read, I parsed "without having to **supply** a virtual environment".

Don't know how to improve it, though.

doc/getting-started.rst
```diff
@@ -0,0 +1,565 @@
...
+
+To limit dependence on the operating system, people sometimes install
+libraries or even database servers as Buildout parts.
+
+Modern Linux container technology (e.g. `Docker
+<https://www.docker.com/>`_) makes it a lot easier to control the
+environment.  If you develop entirely with respect to a particular
+container image, you can have repeatability with respect to that
+image, which is usually good enough because the environment, defined
+by the image is itself repeatable and unshared with other
```
leorochael: missing "closing" comma after `by the image`

doc/getting-started.rst
jimfulton: Doh, I'm sorry. I thought this comment was earlier in the doc.

OK, so sure, there would be contents.

I'll add some (non-windows( listings and I'll incorporate a windows listing if someone gives me one.

doc/getting-started.rst
jimfulton: Thanks. Will do.

doc/getting-started.rst
jimfulton: Yup. Thanks

doc/getting-started.rst
jimfulton: dirty?

doc/getting-started.rst
jimfulton: fixed :)

doc/getting-started.rst
jimfulton: OK, added

doc/getting-started.rst
jimfulton: fixed

doc/getting-started.rst
jimfulton: fixed

doc/getting-started.rst
jimfulton: I changed to "contaminate"

doc/getting-started.rst
jimfulton: added

jimfulton:
Reminder: a semi-rendered version can be found here:

https://github.com/buildout/buildout/blob/getting-started/doc/getting-started.rst

jimfulton:
It's gotten so quiet here.

@tseaver  and @leorochael are you happy with the shape of this?

Should I go ahead and add manuel tests of (most of) the examples? (This won't affect the visible documentation, because manuel.)

doc/getting-started.rst
```diff
@@ -0,0 +1,593 @@
...
+If a Python installation is shared, packages installed by one
+application affect other applications, including buildouts. This can
+lead to unexpected errors.   This is why it's recommended to use a
+`virtual environment <https://virtualenv.pypa.io/en/stable/>`_ or a
+"clean python" built from source with no third-party packages
+installed.
+
+(It's a little hypocritical to recommend installing Buildout into an
+otherwise clean environment, which is why Buildout provides a
+:doc:`bootstrapping mechanism <topics/bootstrapping>` which allows
```
leorochael: I thought we were going to deprecate self-bootstraping...

doc/getting-started.rst
jimfulton: OK, changed

doc/getting-started.rst
jimfulton: No. :)  

I think in a document like this, I'd rather keep things simple and goal oriented.  We can't explain everything at once. I haven't explained what the significance of ``buildout::`` is so there's no reason for people to wonder why these aren't defined there. I provide a link to a topic where they can get the details.

These aren't internal variables. They're part of Buildout's API and will be documented in the reference.  The variables-extending-and-substitutions will mention that variables have defaults (among many other related things).

I added a sentence referring them to the reference to find out what settings are available.

doc/getting-started.rst
jimfulton: Even if we deprecate it, I think we should talk about it given history.

We're definitely not going to emphasize it like we used to.

My plan is to mention that, because it depends on setuptools self-bootstrapping, it might go away in the future.

Buildout is opinionated and self-bootstrapping is an ideal consistent with Buildout's philosophy, even if it becomes impractical at some point.

As a compromise, I moved the mention of self-bootstrapping to a footnote.

jimfulton:
@leorochael sorry I missed some comments. I think I've at least responded to all of them.

leorochael:
LGTM!

doc/getting-started.rst
```diff
@@ -0,0 +1,596 @@
...
+
+   In the Buildout documentation, we'll use the word
+   *buildout* to refer to:
+
+   - The Buildout software
+
+     We'll capitalize the word when we do this.
+
+   - A particular application of Buildout, which has a buildout
+     configuration.
```
mgedmin: I tend to read "application" as "application program", which makes this a bit confusing.  Maybe use more concrete terms, such as "a directory with a `buildout.cfg`"?

doc/getting-started.rst
```diff
@@ -0,0 +1,596 @@
...
+package installer, like ``pip``. Using Buildout recipes, we can
+install custom scripts and configuration files, and much more. For
+example, we could use `configure and make
+<https://pypi.python.org/pypi/zc.recipe.cmmi>`_ to install non-Python
+software from source, we could run JavaScript builders, or do anything
+else that can be automated with Python.
+
+Buildout is a simple automation framework.  There are hundreds of
+recipes to choose from and :doc:`writing new recipes is easy
+<topics/writing-recipes>`.
```
mgedmin: It would be good to let the user know how they can find those hundreds of recipes to choose from.
Do we have we list?  Maybe a PyPI search page?  Is there a Trove classifier for Buildout recipes?
(A short curated list of the most common/most useful/best recipes in a sibling document could be handy as well.)

doc/getting-started.rst
```diff
@@ -0,0 +1,596 @@
...
+   ...
+
+The ``develop`` option takes one more more paths to project `setup.py
+<https://docs.python.org/3.6/distutils/setupscript.html>`_ files or,
+more commonly, directories containing them. Buildout then creates
+"develop eggs" [#develop-eggs]_ for the corresponding projects.
+
+With develop eggs, you can modify the sources and the modified sources
+are reflected in future Python runs (or after `reloads
+<https://docs.python.org/3/library/imp.html?highlight=reload#imp.reload>`_).
```
mgedmin: The `imp` module is deprecated since Python 3.4 and tells you to use https://docs.python.org/3/library/importlib.html#importlib.reload instead.

doc/getting-started.rst
```diff
@@ -0,0 +1,596 @@
...
+
+Then a number of things will happen:
+
+- ``zc.recipe.egg`` will be downloaded and installed in your ``eggs``
+  directory.
+
+- ``ZEO`` and its dependencies will be downloaded and installed. (ZEO
+  is a small Python database server.)
+
+  After this, the eggs directory will look something like::
```
hvelarde: @jimfulton are you aware of the `.. code-block::` directive? it could make things more readable; here for instance you can use `.. code-block:: console`.

doc/getting-started.rst
```diff
@@ -0,0 +1,596 @@
...
+
+Generating configuration and custom scripts
+===========================================
+
+The ``runzeo`` program doesn't daemonize itself. Rather, it's meant to
+be used with a dedicated daemonizer like `zdaemon
+<https://pypi.python.org/pypi/zdaemon>`_ or `supervisord
+<http://supervisord.org/>`_.  We'll use a `recipe to set up zdaemon
+<https://pypi.python.org/pypi/zc.zdaemonrecipe>`_.  Our Buildout
+configuration becomes::
```
hvelarde: @jimfulton and here, `.. code-block:: ini`.

doc/getting-started.rst
```diff
@@ -0,0 +1,596 @@
...
+
+  to start a server and::
+
+    bin/server stop
+
+  to stop it.  The script references a zdaemon configuration file
+  generated by the recipe in ``parts/server/zdaemon.conf``.
+
+- A zdaemon configuration script is generated in
+  ``parts/server/zdaemon.conf`` that looks something like::
```
hvelarde: @jimfulton and here, probably something like `.. code-block:: xml`.

doc/getting-started.rst
jimfulton: Except I specifically say "application of".  I also say "which has a buildout configuration". Hm.

How about: 

A particular use of Buildout, a directory having a Buildout configuration file.

doc/getting-started.rst
jimfulton: I wanted to link to a search result, but...

I tried searching for "buildout recipe", but that yielded lots of false positives.  I then copy/pasted the results and grepped for recipe.  This yields 432 results. An admittedly quick visual scan didn't reveal  any obvious false positives.  I'm sure there are some.  This ad hoc technique probably resulted in false negatives as well.  Of course, it's very common for many recipes to be forks/variations on common ones.

WRT trove classifiers, there are:

  Framework :: Buildout
  Framework :: Buildout :: Extension
  Framework :: Buildout :: Recipe

Is there a way to search trove classifiers? I just looked and didn't see anything. I doubt that many recipes set them and I'm sure the classifiers above didn't exist when many recipes were written.  I tend to view trove setting classifiers as a mostly an empty bureaucratic exercise.  I mean, WTF you can't search them? (Hopefully, I just missed it.)

I would love to be able to link to a reasonable search result.  If someone finds one, I'll happily include it.

A curated list would be great too, but probably a lot of effort and a political minefield, but maybe this would be an interesting topic of a sprint at PyCon or EuroPython.  I think a curated list is out of scope for this PR.

doc/getting-started.rst
jimfulton: Sigh. Hadn't noticed this. ``reload`` -> ``imp.reload`` -> ``importlib.reload``.

Thanks. I'll adjust that link.

doc/getting-started.rst
jimfulton: I was vaguely aware, but I don't usually bother.

Thanks for the suggestion. I'll add these.

doc/getting-started.rst
icemac: Searching for trove classifiers is hidden behind the "browse packages" link:

* [Framework :: Buildout](https://pypi.python.org/pypi?:action=browse&show=all&c=512) 463 hits
* [Framework :: Buildout :: Recipe](https://pypi.python.org/pypi?:action=browse&show=all&c=512&c=545) 51 hits
* [Framework :: Buildout :: Extension](https://pypi.python.org/pypi?:action=browse&c=512&c=546) 8 hits

It seems that many recipes only use `Framework :: Buildout`.

doc/getting-started.rst
jimfulton: Done.

doc/getting-started.rst
jimfulton: Done

doc/getting-started.rst
jimfulton: Done

doc/getting-started.rst
jimfulton: Thanks @icemac.

I suggest adding a footnote with the first link and briefly mention that it has some false positives and negatives.

jimfulton:
I've added manuel tests for the examples.  This is close to being ready to merge, except that https://github.com/buildout/buildout/pull/335 needs to be reviewed/merged first.

I'll start working on a topic shortly.

jimfulton:
Wow, the python 2 failures were because the example buildouts were taking too long.

Can anyone suggest a reasonable but much smaller server example than ZEO?  I don't want to take the time to try to find something myself right now.  I knew this wasn't a great example for this reason, **but I didn't want the distraction** of finding a cheaper example myself **now**.

doc/getting-started.rst
jimfulton: done

PR #352, opened 2017-02-23
Installer: Encapsulate all uses of `._env` and `._path`

leorochael:
All write access to `easy_install.Installer._env` and all access to `easy_install.Installer._path` are now done through `easy_install.Installer` methods, so that buildout extensions can install subclasses that alter their behaviour.

### DIFF
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -379,38 +379,12 @@ class Installer:
 
         tmp = tempfile.mkdtemp(dir=dest)
         try:
-            path = setuptools_path
-
-            args = [sys.executable, '-c',
-                    ('import sys; sys.path[0:0] = %r; ' % path) +
-                    _easy_install_cmd, '-mZUNxd', tmp]
-            level = logger.getEffectiveLevel()
-            if level > 0:
-                args.append('-q')
-            elif level < 0:
-                args.append('-v')
-
-            args.append(spec)
-
-            if level <= logging.DEBUG:
-                logger.debug('Running easy_install:\n"%s"\npath=%s\n',
-                             '" "'.join(args), path)
-
-            sys.stdout.flush() # We want any pending output first
-
-            exit_code = subprocess.call(list(args))
+            paths = call_easy_install(spec, tmp)
 
             dists = []
-            env = pkg_resources.Environment([tmp])
+            env = pkg_resources.Environment(paths)
             for project in env:
                 dists.extend(env[project])
-
-            if exit_code:
-                logger.error(
-                    "An error occurred when trying to install %s. "
-                    "Look above this message for any errors that "
-                    "were output by easy_install.",
-                    dist)
 
             if not dists:
                 raise zc.buildout.UserError("Couldn't install: %s" % dist)
@@ -437,9 +411,7 @@ class Installer:
 
             result = []
             for d in dists:
-                newloc = _move_to_eggs_dir_and_compile(d, dest)
-                [d] = pkg_resources.Environment([newloc])[d.project_name]
-                result.append(d)
+                result.append(_move_to_eggs_dir_and_compile(d, dest))
 
             return result
 
@@ -558,23 +530,7 @@ class Installer:
                     raise zc.buildout.UserError(
                         "Couldn't download distribution %s." % avail)
 
-                if dist.location.endswith('.whl'):
-                    dist = wheel_to_egg(dist, tmp)
-
-                if dist.precedence == pkg_resources.EGG_DIST:
-                    # It's already an egg, just fetch it into the dest
-                    newloc = _move_to_eggs_dir_and_compile(dist, self._dest)
-
-                    # Getting the dist from the environment causes the
-                    # distribution meta data to be read.  Cloning isn't
-                    # good enough.
-                    dists = pkg_resources.Environment([newloc])[
-                        dist.project_name]
-                else:
-                    # It's some other kind of dist.  We'll let easy_install
-                    # deal with it:
-                    dists = self._call_easy_install(
-                        dist.location, self._dest, dist)
+                dists = [_move_to_eggs_dir_and_compile(dist, self._dest)]
 
             finally:
                 if tmp != self._download_cache:
@@ -588,39 +544,46 @@ class Installer:
             dists = [dist]
 
         if not self._install_from_cache and self._use_dependency_links:
-            for dist in dists:
-                if dist.has_metadata('dependency_links.txt'):
-                    for link in dist.get_metadata_lines('dependency_links.txt'):
-                        link = link.strip()
-                        if link not in self._links:
-                            logger.debug('Adding find link %r from %s',
-                                         link, dist)
-                            self._links.append(link)
-                            self._index = _get_index(self._index_url,
-                                                     self._links,
-                                                     self._allow_hosts)
+            self._add_dependency_links_from_dists(dists)
 
         if self._check_picked:
-            # Check whether we picked a version and, if we did, report it:
-            for dist in dists:
-                if not (
-                    dist.precedence == pkg_resources.DEVELOP_DIST
-                    or
-                    (len(requirement.specs) == 1
-                     and
-                     requirement.specs[0][0] == '==')
-                    ):
-                    logger.debug('Picked: %s = %s',
-                                 dist.project_name, dist.version)
-                    self._picked_versions[dist.project_name] = dist.version
-
-                    if not self._allow_picked_versions:
-                        raise zc.buildout.UserError(
-                            'Picked: %s = %s' % (dist.project_name,
-                                                 dist.version)
-                            )
+            self._check_picked_requirement_versions(requirement, dists)
 
         return dists
+
+    def _add_dependency_links_from_dists(self, dists):
+        reindex = False
+        links = self._links
+        for dist in dists:
+            if dist.has_metadata('dependency_links.txt'):
+                for link in dist.get_metadata_lines('dependency_links.txt'):
+                    link = link.strip()
+                    if link not in links:
+                        logger.debug('Adding find link %r from %s',
+                                     link, dist)
+                        links.append(link)
+                        reindex = True
+        if reindex:
+            self._index = _get_index(self._index_url, links, self._allow_hosts)
+
+    def _check_picked_requirement_versions(self, requirement, dists):
+        """ Check whether we picked a version and, if we did, report it """
+        for dist in dists:
+            if not (dist.precedence == pkg_resources.DEVELOP_DIST
+                or
+                (len(requirement.specs) == 1
+                 and
+                 requirement.specs[0][0] == '==')
+                ):
+                logger.debug('Picked: %s = %s',
+                             dist.project_name, dist.version)
+                self._picked_versions[dist.project_name] = dist.version
+
+                if not self._allow_picked_versions:
+                    raise zc.buildout.UserError(
+                        'Picked: %s = %s' % (dist.project_name,
+                                             dist.version)
+                        )
 
     def _maybe_add_setuptools(self, ws, dist):
         if dist_needs_pkg_resources(dist):
@@ -1583,6 +1546,84 @@ class IncompatibleConstraintError(zc.buildout.UserError):
 IncompatibleVersionError = IncompatibleConstraintError # Backward compatibility
 
 
+def call_easy_install(spec, dest):
+    """
+    Call `easy_install` from setuptools as a subprocess to install a
+    distribution specified by `spec` into `dest`.
+    Returns all the paths inside `dest` created by the above.
+    """
+    path = setuptools_path
+
+    args = [sys.executable, '-c',
+            ('import sys; sys.path[0:0] = %r; ' % path) +
+            _easy_install_cmd, '-mZUNxd', dest]
+    level = logger.getEffectiveLevel()
+    if level > 0:
+        args.append('-q')
+    elif level < 0:
+        args.append('-v')
+
+    args.append(spec)
+
+    if level <= logging.DEBUG:
+        logger.debug('Running easy_install:\n"%s"\npath=%s\n',
+                        '" "'.join(args), path)
+
+    sys.stdout.flush() # We want any pending output first
+
+    exit_code = subprocess.call(list(args))
+
+    if exit_code:
+        logger.error(
+            "An error occurred when trying to install %s. "
+            "Look above this message for any errors that "
+            "were output by easy_install.",
+            spec)
+    return glob.glob(os.path.join(dest, '*'))
+
+
+def unpack_egg(location, dest):
+    # Buildout 2 no longer installs zipped eggs,
+    # so we always want to unpack it.
+    dest = os.path.join(dest, os.path.basename(location))
+    setuptools.archive_util.unpack_archive(location, dest)
+
+
+WHEEL_TO_EGG_WARNING = """
+Using unpack_wheel() shim over deprecated wheel_to_egg().
+Please update your wheel extension implementation for one that installs a .whl
+handler in %s.UNPACKERS
+""".strip() % (__name__,)
+
+def unpack_wheel(location, dest):
+    logger.warning(WHEEL_TO_EGG_WARNING)
+    # Deprecated backward compatibility shim. Please do not use.
+    basename = os.path.basename(location)
+    dists = setuptools.package_index.distros_for_location(location, basename)
+    wheel_to_egg(list(dists)[0], dest)
+    
+
+UNPACKERS = {
+    # Buildout 2 no longer installs zipped eggs, so we always want to unpack it.
+    '.egg': unpack_egg,
+    '.whl': unpack_wheel,
+}
+
+
+def _get_matching_dist_in_location(dist, location):
+    """
+    Check if `locations` contain only the one intended dist.
+    Return the dist with metadata in the new location.
+    """
+    # Getting the dist from the environment causes the
+    # distribution meta data to be read.  Cloning isn't
+    # good enough.
+    env = pkg_resources.Environment([location])
+    dists = [ d for project_name in env for d in env[project_name] ]
+    dist_infos = [ (d.project_name, d.version) for d in dists ]
+    if dist_infos == [(dist.project_name, dist.version)]:
+        return dists.pop()
+
 def _move_to_eggs_dir_and_compile(dist, dest):
     """Move distribution to the eggs destination directory.
 
@@ -1594,7 +1635,7 @@ def _move_to_eggs_dir_and_compile(dist, dest):
     running in parallel.  So we copy to a temporary directory first.
     See discussion at https://github.com/buildout/buildout/issues/307
 
-    We return the new location.
+    We return the new distribution with properly loaded metadata.
     """
     # First make sure the destination directory exists.  This could suffer from
     # the same kind of race condition as the rest: if we check that it does not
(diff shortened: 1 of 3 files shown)


leorochael:
BTW, this PR should add no change in behaviour to buildout. It just moves things around.

jimfulton:
Thanks!

I'll look at this soon.

src/zc/buildout/easy_install.py
```diff
@@ -83,6 +83,26 @@ def realpath(path):
 FILE_SCHEME = re.compile('file://', re.I).match
 DUNDER_FILE_PATTERN = re.compile(r"__file__ = '(?P<filename>.+)'$")
 
+
+class reify(object):
```
jimfulton: I've written this damn thing too many times. :/

src/zc/buildout/easy_install.py
```diff
@@ -213,20 +233,34 @@ def __init__(self,
...
-        self._env = pkg_resources.Environment(path)
         self._index = _get_index(index, links, self._allow_hosts)
         self._requirements_and_constraints = []
         self._check_picked = check_picked
 
         if versions is not None:
             self._versions = normalize_versions(versions)
 
+    @reify
+    def _env(self):
```
jimfulton: Why does this have to be lazy?

Why not just cale this make_env and call it from ``__iniit__``?

src/zc/buildout/easy_install.py
```diff
@@ -213,20 +233,34 @@ def __init__(self,
...
+    def _env_rescan_dest(self):
+        self._env.scan([self._dest])
+
+    def _env_get_best_match(self, requirement, ws):
+        dist = self._env.best_match(requirement, ws)
+        return dist
+
+    def _env_get_dists_for_project(self, project_name):
+        return self._env[project_name]
+
```
jimfulton: I realize that these are hooks.  But It annoys me to add more levels of call, mainly because it makes the code harder to read.

The code on the buildout.wheel branch could just as easily live here.  At which many of these functions can go away, because they'll no-longer be hooks. There are even todos over there to move them here. Let's go ahead and do that.

Please move the implementations from the humpty-had-a-great-fall branch of buildout.wheel here and inline the small functions that are only called once.

leorochael: > I realize that these are hooks. But It annoys me to add more levels of call, mainly because it makes the code harder to read.

It's more than just hooks. By wrapping all access to `._env` I can make sure all it's inputs and outputs are interceptable in subclasses of Installer without creating a subclass of Environment.

For installing wheels, I can get away with intercepting either only the places that modify `._env` or the places that read from it, noot necessarily both. But if I leave some `._env` access unwrapped, there's a risk another direct `._env` access is introduced later that will let non-egg-precedence dists leak in where we're expecting egg-precedence dists.

Anyway, for now, I'll wrap all write accesses to `._env`, which should be enough for `buildout.wheel`.

> Please move the implementations from the humpty-had-a-great-fall branch of buildout.wheel here and inline the small functions that are only called once.

All of them? Including the ones that try to scan for non-egg dists inside `._dest`? That will make `buildout.wheel` tiny. I'll assume silence is consent ;-)

jimfulton:
On Sun, Feb 26, 2017 at 2:06 PM, Leonardo Rochael Almeida <
notifications@github.com> wrote:

> *@leorochael* commented on this pull request.
> ------------------------------
>
> In src/zc/buildout/easy_install.py
> <https://github.com/buildout/buildout/pull/352#discussion_r103105950>:
>
> > @@ -83,6 +83,26 @@ def realpath(path):
>  FILE_SCHEME = re.compile('file://', re.I).match
>  DUNDER_FILE_PATTERN = re.compile(r"__file__ = '(?P<filename>.+)'$")
>
> +
> +class reify(object):
>
> It's one of those decorators that, since it's so simple, either it's in
> the stdlib, or nobody is going to rely on a pip published version lest they
> fall victim to some "npm left pad" situation
> <https://www.theregister.co.uk/2016/03/23/npm_left_pad_chaos/>.
>
Years ago I discussed this with Guido and we agreed to add a ``default``
decorator like this, but then I never followed through. :(  It's hard to be
motivated to add things to the standard library that I won't benefit from
until years later.

...

> I realize that these are hooks. But It annoys me to add more levels of
> call, mainly because it makes the code harder to read.
>
> It's more than just hooks. By wrapping all access to ._env I can make
> sure all it's inputs and outputs are interceptable in subclasses of
> Installer without creating a subclass of Environment.
>

You just described hooks. :)


> For installing wheels, I can get away with intercepting either only the
> places that modify ._env or the places that read from it, noot
> necessarily both. But if I leave some ._env access unwrapped, there's a
> risk another direct ._env access is introduced later that will let
> non-egg-precedence dists leak in where we're expecting egg-precedence dists.
>
> Anyway, for now, I'll wrap all write accesses to ._env, which should be
> enough for buildout.wheel.
>
I don't think that's necessary.

> Please move the implementations from the humpty-had-a-great-fall branch of
> buildout.wheel here and inline the small functions that are only called
> once.
>
> All of them?
>

Yes.


> Including the ones that try to scan for non-egg dists inside ._dest?
>
Yes, why not?

This stuff is going to land eventually, why wait.  I want to avoid the
hooks, which complicate the code.

> That will make buildout.wheel tiny.
>
That's a good thing.

The only reason to keep this as an extension is to avoid introducing new
dependencies why the wheel installation machinery is in such flux.

I'm thinking that at PyCon sprints I'd work on moving bits of pip and maybe
wheel that we're using into packaging.

leorochael:
@jimfulton, I've moved here the methods from `buildout.wheel` and streamline the `_call_easy_install` interface somewhat (it can be simplified a bit more still without loss of functionality).

Should I go ahead and remove the `wheel_to_egg()` hook since it's no longer needed?

jimfulton:
On Mon, Feb 27, 2017 at 11:24 PM, Leonardo Rochael Almeida <
notifications@github.com> wrote:

> @jimfulton <https://github.com/jimfulton>, I've moved here the methods
> from buildout.wheel and streamline the _call_easy_install interface
> somewhat
>

Cool. I'll look at this soon.


> (it can be simplified a bit more still without loss of functionality).
>

Sound like a good idea. :)


> Should I go ahead and remove the wheel_to_egg() hook since it's no longer
> needed?
>
It's needed by the existing buildout.wheel extension.

Did you see my comment on the buildout.wheel PR about using the hook rather
than an Installer subclass?  Personally, I think it would be cleaner to
adjust the way the hook is used slightly and keep using it.

leorochael:
> > ([`_call_easy_install`] can be simplified a bit more still without loss of functionality).
> 
> Sound like a good idea. :)

I've simplified it as much as I could. It now only expects a spec and a destination directory and returns a list of paths to distributions installed inside it. It's caller, `_call_easy_install_for_dist()` does all the heavy lifting from there.

> > Should I go ahead and remove the wheel_to_egg() hook since it's no longer needed?
> 
> It's needed by the existing buildout.wheel extension.

I don't think the extension, as it is, is being extensively used. In any case, we could release a new `zc.buildout` and a new `buildout.wheel` extension at the same time. Or release a newer point release of the old `buildout.wheel` restricting the `zc.buildout` version with a `<=` comparison.

> Did you see my comment on the buildout.wheel PR about using the hook rather than an Installer subclass?  Personally, I think it would be cleaner to adjust the way the hook is used slightly and keep using it.

I looked at all comments in that PR, including the collapsed ones, but couldn't find one asking me to use the hook instead of a subclass. The closes I found is your comment that you "don't like the coupling to the pypa packages", which I interpreted to mean "no subclassing of classes in other projects".

In any case, as it is now, the hook is neither useful, as it operates on a level too high and demands things that are no longer necessary (a conversion to an `egg`, manual cleanup of intermediate artifacts due to #332), nor recommended (again due to #332).

If you'd like me to engineer it such that, to add wheel support, an extension shouldn't have to subclass the `Installer` (which is already possible now: an extension could just monkey-patch `Installer._call_easy_install`), then I can create another hook to replace `wheel_to_egg()`, say `unpack_wheel(location, dest)` that would be called by the current `_call_easy_install()`.

Another alternative is to move `Installer._call_easy_install()` to a top-level function (`call_easy_install()`), since it doesn't depend on `self` at all, and let extensions override that instead of an explicit wheel hook.

In any case, since I love removing code, now that I proved that the `wheel_to_egg()` hook is unnecessary, I'm itching to remove it, but you're the maintainer, so I'm happy to defer to your decision.

As for this PR, unless you'd like further changes as discussed above, I consider it in shape to be merged, subject to review and approval by you or others.

jimfulton:
On Wed, Mar 1, 2017 at 12:35 PM, Leonardo Rochael Almeida <
notifications@github.com> wrote:

> ([_call_easy_install] can be simplified a bit more still without loss of
> functionality).
>
> Sound like a good idea. :)
>
> I've simplified it as much as I could. It now only expects a spec and a
> destination directory and returns a list of paths to distributions
> installed inside it. It's caller, _call_easy_install_for_dist() does all
> the heavy lifting from there.
>
Cool. I'll look soon.



> Should I go ahead and remove the wheel_to_egg() hook since it's no longer
> needed?
>
> It's needed by the existing buildout.wheel extension.
>
> I don't think the extension, as it is, is being extensively used.
>
IDK, but I don't want to punish early adopters.

> In any case, we could release a new zc.buildout and a new buildout.wheel
> extension at the same time. Or release a newer point release of the old
> buildout.wheel restricting the zc.buildout version with a <= comparison.
>
> Did you see my comment on the buildout.wheel PR about using the hook
> rather than an Installer subclass? Personally, I think it would be cleaner
> to adjust the way the hook is used slightly and keep using it.
>
> I looked at all comments in that PR, including the collapsed ones, but
> couldn't find one asking me to use the hook instead of a subclass.
>
Dang, I don't see my comment either. :-[

OK, let me try again. Over on the wheel branch, you're unpacking the wheel
from ``call_easy_install``.  This is the sole reason for the Installer
subclass.  It's also a little odd to do it there, since we're not going to
invoke a setup script.

Why not do this from the hook instead?  We could then update buildout to
deal with the slightly different result.  BTW, it looks like
call_easy_install returns a sequence of dists, but you seem to be returning
a location.  Anyway, I'd modify buildout to check the returned location and
just use the returned dist if it's location is in the dest.

leorochael:
> Over on the wheel branch, you're unpacking the wheel from ``call_easy_install``.
> This is the sole reason for the Installer
subclass.
> It's also a little odd to do it there, since we're not going to invoke a setup script.

Indeed, the method becomes a misnomer if we're not actually invoking `easy_install`.

> Why not do this from the hook instead?

Because the hook expects a `dist` as result, and will further process this `dist`, which forces the hook implementation to handle tempdirs if it will be done safely (see #332), in duplication of what `_call_easy_install()` already does.

And yet, whatever we end up installing on `_dest` needs to be recognized later as a `dist` by buildout without the help of the `wheel_to_egg()` hook, like `_call_easy_install()` already does when returning the installed dists, but also like `_satisfied()` does when scanning `_dest`.

I hate code duplication, almost as much as I dislike special cases. And removing humpty forced me to insert all those code duplications inside the hook with the knowledge that the same things I did there would be repeated with whatever I returned from the hook later on.

It was too much for me ;-)

> We could then update buildout to deal with the slightly different result.

You now know how I feel about them special cases ;-)

> BTW, it looks like call_easy_install returns a sequence of dists, but you seem to be returning
a location.

So, before this PR, `_call_easy_install()` returned dists. In this PR, I renamed it to `_call_easy_install_for_dist()` that still returns dists.

I then extracted the core of the method, which only invoked `easy_install` and returned only locations, and called THAT `_call_easy_install()`, leaving behind all tempdir manipulation and dist interpretation in the outer, renamed method.

So, in the extension, all I have to do is unpack the wheel into a directory and return it in the inner monkey-patched `_call_easy_install()`, while the outer `_call_easy_install_for_dist()` method proceeds as usual.

> Anyway, I'd modify buildout to check the returned location and just use the returned dist if it's location is in the dest.

That's kind of what Buildout already does with the result of the inner `_call_easy_install()` I extracted above, so I'd prefer not to duplicate it.

If you don't want the overridden method not to be called  `_call_easy_install()`, we can extract the lines between [574](https://github.com/buildout/buildout/blob/2cf98bdbe3ad667b87aa6c09beaad8d8e5cf5881/src/zc/buildout/easy_install.py#L574) and the `finally:` line into the outer  `_call_easy_install_for_dist()` method and rename it to `_install_dist()`

Then inside it we can do a pattern match to dispatch to various unpacking/installing functions depending on file extension, (including, eventually, an `unpack_wheel()` hook) that falls back to the inner `_call_easy_install()`.

What do you think?

leorochael:
@jimfulton, this last commit removed all excuses ;-)

Not only it adds a much cleaner, special-case-free hook to wheel support, it does so while at the same time keeping compatibility with the old hook (with appropriate deprecation warnings), and fixing #332 to boot.

(it also allowed me to drop another handful of lines from `buildout.wheel`)

If you need any info regarding the proposed changes, let me know.

src/zc/buildout/easy_install.py
```diff
@@ -518,98 +570,76 @@ def _get_dist(self, requirement, ws, for_buildout_run=False):
...
-                        "uses namespace packages but the distribution "
-                        "does not require setuptools.",
-                        dist)
-                requirement = self._constrain(
-                    pkg_resources.Requirement.parse('setuptools')
-                    )
-                if ws.find(requirement) is None:
-                    for dist in self._get_dist(requirement, ws):
-                        ws.add(dist)
+    def _add_dependency_links_from_dists(self, dists):
```
jimfulton: I can live with this, but I wouldn't have split these out because the condition introducing the code made the intent as clear as a function name.

But the calling function and these aren't very small, so <shrug> :)

src/zc/buildout/easy_install.py
```diff
@@ -169,6 +169,60 @@ def _execute_permission():
...
+
+def dist_needs_pkg_resources(dist):
+    """
+    A distribution needs setuptools/pkg_resources added as requirement if:
+
+        * It has namespace packages declared with:
+        - `pkg_resources.declare_namespace()`
+        * Those namespace packages don't fall back to `pkgutil`
+        * It doesn't have `setuptools/pkg_resources` as requirement already
+    """
```
leorochael: > Oh gawd, I'm regretting ever providing this crutch. :)

Wanna get rid of it? It's fine by me. I actually thought of ripping it out myself. Ended up making it more complex instead so it wouldn't be so eager to add setuptools to everything now that it's not needed.

src/zc/buildout/easy_install.py
```diff
@@ -169,6 +169,60 @@ def _execute_permission():
...
+    def requires_pkg_resources_already(dist):
```
leorochael: > Why are this and the following function nested?

Because they're tiny functions that are only used once by another tiny function...?

I like that the final algorithm for deciding if `setuptools` is needed is expressed clearly in the `return` expression, so I inlined those functions that are never going to be used outside of `dist_needs_pkg_resources()` to make that nice looking `return` expression.

As for the rest of the tiny functions, I actually use `get_namespace_package_paths()` in `buildout.wheel` to fix up PEP 420 namespaces for Python 2.

If you'd rather we get rid of the crutch. I can move it over there (and maybe I'll even inline it!).

Otherwise, I can try to inline most of them and keep just this one out.

But in general, IMO Buildout is plagued by very long multi-screen functions and methods, which I think makes them hard to understand. The interaction between `Installer.install()`, `Installer._get_dist()` and `Installer._call_easy_install()` took many passes for me to see that there was a lot of redundant code between them and where I could add the smallest possible wheel hook.

So, between the context switching (which I try to alleviate using expressive function names), and multi-page functions, we need to strike some balance for maximum readability.

src/zc/buildout/easy_install.py
```diff
@@ -213,20 +267,51 @@ def __init__(self,
...
 
+    def _make_env(self):
+        full_path = self._get_dest_dist_paths() + self._path
+        env = pkg_resources.Environment(full_path)
+        # this needs to be called whenever self._env is modified (or we could
+        # make an Environment subclass):
+        self._eggify_env_dest_dists(env, self._dest)
+        return env
+
+    def _env_rescan_dest(self):
```
leorochael: > Only used once. Please inline.

I know you mentioned you don't think it's necessary, but `_env_rescan_dest()` and  `_make_env()` are the only methods that modify `self._env`.

I feel that if we just inline it we might forget to `_eggify_env_dest_dists()` in the future if we add another place that modifies `self._env`.

Or we could use a subclass of `Environment`. We would only need to override the `Environment.add()` method in this case.

But it's your call.

src/zc/buildout/easy_install.py
leorochael: > ... I wouldn't have split these out because the condition introducing the code made the intent as clear as a function name.
> 
> But the calling function and these aren't very small, so :)

Yes, it's one of those multi-page functions. Extracting these methods made the function almost fit in one screen in my editor.

src/zc/buildout/easy_install.py
```diff
@@ -330,48 +415,17 @@ def _satisfied(self, req, source=None):
             str(req))
         return best_we_have, None
 
-    def _load_dist(self, dist):
-        dists = pkg_resources.Environment(dist.location)[dist.project_name]
-        assert len(dists) == 1
-        return dists[0]
-
-    def _call_easy_install(self, spec, ws, dest, dist):
+    def _call_easy_install_for_dist(self, spec, dest, dist):
```
leorochael: > Please rename this back to `_call_easy_install`. It's just a needless change now.

Ok. Was half tempted to do it already.

jimfulton:
On Fri, Mar 3, 2017 at 1:59 PM, Leonardo Rochael Almeida <
notifications@github.com> wrote:

> *@leorochael* commented on this pull request.
>
> Thanks for the input. I'll try to combat the plague of little functions
> and come up with a better PR.
>
Thanks!

------------------------------
>
> In src/zc/buildout/easy_install.py
> <https://github.com/buildout/buildout/pull/352#discussion_r104210601>:
>
> > +    for namespace in dist.get_metadata_lines('namespace_packages.txt'):
> +        yield os.path.join(*(base + namespace.split('.') + init))
> +
> +def get_namespace_package_sources(dist):
> +    for path in get_namespace_package_paths(dist):
> +        yield _get_source(path)
> +
> +def dist_needs_pkg_resources(dist):
> +    """
> +    A distribution needs setuptools/pkg_resources added as requirement if:
> +
> +        * It has namespace packages declared with:
> +        - `pkg_resources.declare_namespace()`
> +        * Those namespace packages don't fall back to `pkgutil`
> +        * It doesn't have `setuptools/pkg_resources` as requirement already
> +    """
>
> Oh gawd, I'm regretting ever providing this crutch. :)
>
> Wanna get rid of it? It's fine by me.
>
It's sooooo tempting.  Maybe later.

https://github.com/buildout/buildout/issues/364

...

> I like that the final algorithm for deciding if setuptools is needed is
> expressed clearly in the return expression, so I inlined those functions
> that are never going to be used outside of dist_needs_pkg_resources() to
> make that nice looking return expression.
>
I understand, but...



> As for the rest of the tiny functions, I actually use
> get_namespace_package_paths() in buildout.wheel to fix up PEP 420
> namespaces for Python 2.
>
Ah, OK.
...

> Otherwise, I can try to inline most of them and keep just this one out.
>
+1



> But in general, IMO Buildout is plagued by very long multi-screen
> functions and methods, which I think makes them hard to understand.
>
There are a number of interacting factors.  WRT length and abstraction,
this is a tradeoff. Note that I didn't object (too much :)) to splitting
some functions out that are only called once.

> The interaction between Installer.install(), Installer._get_dist() and
> Installer._call_easy_install() took many passes for me to see that there
> was a lot of redundant code between them and where I could add the smallest
> possible wheel hook.
>
Yup.  This architecture has evolved a lot, often under battle and and with
many hands.

I suspect a total rewrite is warranted at some point, possible if we stop
using setuptools.


> So, between the context switching (which I try to alleviate using
> expressive function names), and multi-page functions, we need to strike
> some balance for maximum readability.
>

Yup.

...

> I know you mentioned you don't think it's necessary, but
> _env_rescan_dest() and _make_env() are the only methods that modify
> self._env.
>
> I feel that if we just inline it we might forget to
> _eggify_env_dest_dists() in the future if we add another place that
> modifies self._env.
>
OK, since you feel strongly, I defer.


> Or we could use a subclass of Environment.
>
Nooooooo.

Really subclassing across orgs is a bad idea unless the provider advertises
a subclassing API.

..


> Yes, it's one of those multi-page functions. Extracting these methods made
> the function almost fit in one screen in my editor.
>
You need a bigger screen. :)

leorochael:
> You need a bigger screen. :)

Hmm, I'd love a new 4K monitor... donations accepted ;-)

Should be mergeable now as soon as travis is green.

leorochael:
> @jimfulton restored the encapsulate-easy-install-env branch

This branch is OK to remove by me. I have no further use for it.

PR #353, opened 2017-02-24
easy_install: moar loggin' for fetchin'

leorochael:
Make it clear from where we're fetching a dist that we don't have installed, reporting the specific URL or the path from inside the download cache.


leorochael:
@jimfulton, others, this looks ok?

jimfulton:
Yup. :) I thought this was failing earlier.

leorochael:
It was. I fixed it before asking for approval earlier today.

PR #354, opened 2017-02-25
Added reference documentation for command-line options

jimfulton:
- Intentionally didn't document -o, -O and -n

  Offline mode is an attractive nuisance.

  -n is unlikely to be needed and can be spelled newest=true

- Removed -o, -O and -n from command help.

- Sorted options alphabetically in the reference doc and command help.

- Removed documentation of the log-level option.  It just confuses things.
  I seriously doubt that anyone uses it.

### DIFF
--- a/doc/reference.rst
+++ b/doc/reference.rst
@@ -11,15 +11,15 @@ A Buildout execution is of the form:
 
 .. code-block:: console
 
-  buildout [buildout-options] [settings] [subcommand [subcommand-arguments]]
+  buildout [buildout-options] [assignments] [command [command arguments]]
 
-Settings take the form ``section:option=value`` and override (or
-augment) settings in configuration files.  For example, to pin a
+Assignments take the form ``section:option=value`` and override (or
+augment) options in configuration files.  For example, to pin a
 version of ZEO you could use ``versions:ZEO=4.3.1``.  The section
 defaults to the ``buildout`` section.  So, for example: ``parts=test``
 sets the ``buildout`` section ``parts`` option.
 
-Command-line settings overrides can use ``+=`` and ``-=`` to
+Command-line assignments can use ``+=`` and ``-=`` to
 :ref:`merge values with existing values <merge-values-with-existing-values>`
 
 Buildout command-line options
@@ -42,7 +42,7 @@ Buildout command-line options
 
 ``-N``
   Run in :ref:`non-newest mode <non-newest-mode>`.  This is equivalent
-  to the command-line setting ``newest=false``.
+  to the command-line assignment ``newest=false``.
 
 ``-q``
    Decrease the level of verbosity.  This option can be used multiple
@@ -76,8 +76,8 @@ Buildout command-line options
 
    At the next, "debugging", level, ``-vv`` (or equivalently ``-v
    -v``), buildout prints low-level debugging information, including a
-   listing of all configuration options, including: default settings,
-   computed settings and the results of :ref:`value substitutions
+   listing of all configuration options, including: default options,
+   computed options and the results of :ref:`value substitutions
    <value-substitutions>` and :ref:`macros <macros-label>`.
 
    Using this option more than twice has no effect.
@@ -85,21 +85,21 @@ Buildout command-line options
 ``--version``
    Print buildout version number and exit.
 
-Buildout subcommands
---------------------
+Buildout commands
+-----------------
 
-.. _bootstrap-subcommand:
+.. _bootstrap-command:
 
 bootstrap
 _________
 
-Install a local ``bootstrap`` script.  The ``bootstrap`` subcommand
+Install a local ``bootstrap`` script.  The ``bootstrap`` command
 doesn't take any arguments.
 
 See :doc:`Bootstrapping <topics/bootstrapping>` for information on why
 you might want to do this.
 
-.. _init-subcommand:
+.. _init-command:
 
 init [requirements]
 ____________________
--- a/doc/reference.rst
+++ b/doc/reference.rst
@@ -326,11 +326,13 @@ offline, default: 'false'
 
   .. Warning:: Offline mode is deprecated.
 
-     Its purpose has evolved over time and the end result doesn't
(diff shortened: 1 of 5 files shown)


jimfulton:
I'll probably get a test failure because I changed the help text.

leorochael:
One thing we need to consider is that some recipes actually rely on the [`offline` switch instead of the `install-from-cache` one](https://github.com/anybox/anybox.recipe.odoo/blob/fdf58c00d5fdb799474ffaaea08d7159a458f8ea/anybox/recipe/odoo/base.py#L188):

In the case I linked above, specifically, the `offline` switch makes more sense than an `install-from-cache` one. It's used by a section of the recipe that checks out repositories for odoo add-ons (remember when Zope Products were still not proper Python packages? Something like that). The `offline` switch is used to tell the recipe not to try to fetch from the remote repos, but just checkout the configured branch/tag from the previously cloned repos already inside the buildout.

I myself use the -o switch by reflex. If we deprecate it, then perhaps we should make it an alias of `install-from-cache`. Or do the other way around, and make the offline switch the official one, and `install-from-cache` just an alias (my preference, really, but open to be convinced otherwise).

jimfulton:
On Sat, Feb 25, 2017 at 9:06 PM, Leonardo Rochael Almeida <
notifications@github.com> wrote:

> One thing we need to consider is that some recipes actually rely on the
> offline switch instead of the install-from-cache one
> <https://github.com/anybox/anybox.recipe.odoo/blob/fdf58c00d5fdb799474ffaaea08d7159a458f8ea/anybox/recipe/odoo/base.py#L188>
> :
>
> In the case I linked above, specifically, the offline switch makes more
> sense than an install-from-cache one. It's used by a section of the
> recipe that checks out repositories for odoo add-ons (remember when Zope
> Products were still not proper Python packages? Something like that). The
> offline switch is used to tell the recipe not to try to fetch from the
> remote repos, but just checkout the configured branch/tag from the
> previously cloned repos already inside the buildout.
>
It could as easily interpret install-from cache this way, where the
previous clones are viewed as cache.

> I myself use the -o switch by reflex.
>
Are you doing this for speed? (I assume so.) If so, you should train
yourself to use -N instead.

> If we deprecate it, then perhaps we should make it an alias of
> install-from-cache.
>
Perhaps, except I think the main use case for install-from-cache is
installing from a source distro (or similar situation).  I hardly think
this and similar use cases need shortcuts.

The original motivation for -o was speed, but -N is *much* better for that,
because it still downloads if there is a need to.

> Or do the other way around, and make the offline switch the official one,
> and install-from-cache just an alias (my preference, really, but open to
> be convinced otherwise).
>
That would be backward incompatible.

It would be interesting to know if anyone depends on the "no install" quirk
of offline mode. If we could convince ourselves that no one was, then I'd
pe happy to do away with offline mode and make install-from-cache the new
offline mode.  That's more or less what I'm doing.

But I think the people who have the habit of using -o *really* should
switch to -N.

jimfulton:
To be clear, I wasn't suggesting doing away with offline mode.  I was simply de-emphasizing it in the documentation because it's confusing.

mgedmin:
Hey, let's make `-o`/`offline=true` Do The Right Thing in buildout 3.0, so people who want to install stuff (from their shared cache) without hitting the network can use the intuitive option name.

src/zc/buildout/buildout.py
```diff
@@ -1777,27 +1777,32 @@ def _check_for_unused_options_in_section(buildout, section):
...
+     This defaults to the file named "buildout.cfg" in the current
+     working directory.
 
-  --version
+  -D
 
-     Print buildout version number and exit.
+    Debug errors.  If an error occurs, then the post-mortem debugger
+    will be started. This is especially useful for debuging recipe
+    problems.
```
mgedmin: Some of these text sections are indented with 4 spaces, and some are indented with 5 spaces.  This bugs me. ;)

doc/reference.rst
```diff
@@ -11,35 +11,95 @@ A Buildout execution is of the form:
...
-.. _bootstrap-subcommand:
+``-c config_file``
+   Specify the path (or URL) to the buildout configuration file to be used.
+   This defaults to the file named ``buildout.cfg`` in the current
+   working directory.
+
+``-D``
+  Debug errors.  If an error occurs, then the post-mortem debugger
+  will be started. This is especially useful for debugging recipe
+  problems.
```
mgedmin: Some of these text sections are indented with 3 spaces, and some are indented with 2 spaces.  This bugs me. ;)

mgedmin:
> -n is unlikely to be needed and can be spelled newest=true

I often use `newest = false` in my buildout configs to make it fast by default, and then manually run `bin/buildout -n` every now and then when I want to upgrade all the packages to latest versions.

(Actually, what happens is I run `bin/buildout --help` because I can never remember which of the `-n`/`-N` is newest and which is non-newest.  I've long wanted these options to have long aliases `--newest`/`--non-newest`.  I suppose `newest=true/false` are explicit enough, even if they annoy me by being longer to type.  But I digress.)

I guess the above means I think `-n` ought to be documented.

src/zc/buildout/buildout.py
jimfulton: As well it should!

jimfulton:
WRT -o, I'd be OK with changing the meaning of this for buildout 3. If we do that though we should start warning.

IMO, it would be better to think of a better single-character option.

jimfulton:
WRT ``-n``, I'd rather not encourage being non-repeatable by default. :) Seriously.

WRT to the remembering the option, you could just use: ``newest=true``.

Part of my reluctance to document ``-n`` is my distaste for the words I might write. :)

I'd yield to the majority if someone proposed documentation for ``-n`` in a separate PR and someone else approved it. :)

doc/reference.rst
jimfulton: Me too. fixed.

src/zc/buildout/buildout.py
jimfulton: fixed

PR #360, opened 2017-02-26
Added missing section on conditional sections

jimfulton:
Fortunately, I was able to move and update the not-bad doctest here,
because it was more doc than test.


tseaver:
LGTM.  You might want to `git rebase -i` to amend the "consitional" in the commit message. :)  I fixed it in the issue title.

jimfulton:
Thanks.

doc/topics/variables-extending-and-substitutions.rst
```diff
@@ -131,6 +131,91 @@ extends-cache buildout option <extends-cache-buildout-option>`.
...
+
+A conditional section has a colon and then a Python expression after
+the name.  If the Python expression result is true, the section
+options from the section are included.  If the value is false, the
+section is ignored.
+
+Some things to note:
+
+- If there is no exception, then options from the section are
+  included.
```
mgedmin: I don't understand this sentence.  It seems to repeat the previous paragraph somewhat (if the result is true), and contradict it somewhat (if there's no exception, but the value was false, there's no inclusion).  Also, does "exception" here mean a Python exception raised when evaluating the conditional expression?  Maybe it would be better to say what happens if there *is* an exception.

doc/topics/variables-extending-and-substitutions.rst
```diff
@@ -131,6 +131,91 @@ extends-cache buildout option <extends-cache-buildout-option>`.
...
+sys_version    ``sys.version.lower()``
+pypy           True if running PyPy
+jython         True if running Jython
+iron           True if running Iron Python
+cpython        True if not running PyPy, Jython, or Iron Python
+sys_platform   ``str(sys.platform).lower()``
+linux          True if running on linux
+windows        True if running on Windows
+cygwin         True if running on cygwin
+solaris        True if running on solaris
```
mgedmin: I'm a pedant and so I would prefer to see Linux, Cygwin and Solaris capitalized.

doc/topics/variables-extending-and-substitutions.rst
```diff
@@ -131,6 +131,91 @@ extends-cache buildout option <extends-cache-buildout-option>`.
...
+macosx         True if running on Mac OS X
+posix          True if running on a POSIX-compatible system
+bits32         True if running on a 32-bit system.
+bits64         True if running on a 64-bit system.
+little_endian  True if running on a little-endian system
+big_endian     True if running on a little-endian system
```
mgedmin: I think you meant "big-endian" here.

doc/topics/variables-extending-and-substitutions.rst
jimfulton: This, like most of this text was just moved from a doctest. :)

If there's an exception, the buildout will fail. I don't think that needs to be said.

I'm inclined to remove this bullet and reformat the remaining bullet as a paragraph, as it makes no sense to have one bullet.

Thanks.

doc/topics/variables-extending-and-substitutions.rst
jimfulton: "pendent"?

But I agree in any case.

I also wonder if these should be sorted.

doc/topics/variables-extending-and-substitutions.rst
jimfulton: Yes. (Bug copied from doctest.)  I'll fix.

Thanks.

doc/topics/variables-extending-and-substitutions.rst
jimfulton: fixed, except for sorting.

Thoughts on sorting?

doc/topics/variables-extending-and-substitutions.rst
jimfulton: Fixed

PR #367, opened 2017-03-05
Automatic dependency part selection

### LINKED ISSUE #366: Provide a way for a part to declare that it depends on other parts without using value interpolation.
See: https://groups.google.com/forum/#!topic/buildout-development/NIm9kDrnmE8

We didn't reach a conclusion there. In the mean time, I'm going to rework the implicit parts example, https://github.com/buildout/buildout/blob/implicit-parts/doc/topics/implicit-parts.rst to use the server example and just note this issue.

jimfulton:
This is both doc and functionality.

Fixed #366


src/zc/buildout/buildout.py
```diff
@@ -1237,6 +1237,10 @@ def _initialize(self):
         if name == 'buildout':
             return # buildout section can never be a part
 
+        for dname in self.get('<part-dependencies>', '').strip().split():
+            # force use of dependencies in buidout:
```
mgedmin: Typo: buidout -> buildout

src/zc/buildout/buildout.py
```diff
@@ -1237,6 +1237,10 @@ def _initialize(self):
         if name == 'buildout':
             return # buildout section can never be a part
 
+        for dname in self.get('<part-dependencies>', '').strip().split():
```
mgedmin: `.strip().split()` is equivalent to `.split()`.

leorochael: > `.strip().split()` is equivalent to `.split()`.

TIL!

In 1.5 decades of Python I've been expecting empty strings at the ends of the list in case of unstripped whitespace...

jimfulton: Yeah, me too.  I'm 98% sure this is because it behaved that way at one time.

jimfulton:
Thanks for the review!

PR #368, opened 2017-03-06
Testing recipes

jimfulton:
Document how to (and not to :)) test recipes.

Also moved some useful functions into testing.

Made some adjustments to ``zc.buildout.testing.Buildout`` to make it more useful for testing ordinary recipes.

### DIFF
--- a/doc/topics/writing-recipes.rst
+++ b/doc/topics/writing-recipes.rst
@@ -337,6 +337,155 @@ entry points, using the type ``zc.buildout.uninstall`` as can be seen
 in the `zc.recipe.rhrc setup script
 <https://github.com/zopefoundation/zc.recipe.rhrc/blob/master/setup.py#L23>`_.
 
+Testing recipes
+================
+
+The recipe API is fairly simple and standard unit-testing approaches
+can be used.  We'll use a helper class,
+``zc.buildout.testing.Buildout`` [#helper-class-refined-in-2.9]_ to
+provide a minimal buildout environment.
+
+Let's write a test for our configuration recipe.  We need to verify that:
+
+- The recipe generates a ``path`` option.
+
+- The recipe generates a file in the correct place.
+
+- The recipe returns the path it created from ``install``.
+
+We create a ``testdemoconfigrecipe.py`` file containing our tests::
+
+  import os
+  import shutil
+  import tempfile
+  import unittest
+  import zc.buildout.testing
+
+  class RecipeTests(unittest.TestCase):
+
+      def setUp(self):
+          self.here = os.getcwd()
+          self.tmp = tempfile.mkdtemp()
+          os.chdir(self.tmp)
+          self.buildout = buildout = zc.buildout.testing.Buildout()
+          self.config = 'some config text\n'
+          buildout['config'] = dict(contents=self.config)
+          import democonfigrecipe
+          self.recipe = democonfigrecipe.Recipe(
+              buildout, 'config', buildout['config'])
+
+      def tearDown(self):
+          os.chdir(self.here)
+          shutil.rmtree(self.tmp)
+
+      def test_path_option(self):
+          buildout = self.buildout
+          self.assertEqual(os.path.join(buildout['buildout']['parts-directory'],
+                                        'config'),
+                           buildout['config']['path'])
+
+      def test_install(self):
+          buildout = self.buildout
+          self.assertEqual(self.recipe.install(), [buildout['config']['path']])
+          with open(buildout['config']['path']) as f:
+              self.assertEqual(self.config, f.read())
+
+  if __name__ == '__main__':
+      unittest.main()
+
+.. -> src
+
+   >>> write(src, 'src', 'testdemoconfigrecipe.py')
+
+In the ``setUp`` method, we created a temporary directory and changed
+to it.  This is useful to make sure we have a clean working
+directory.  We clean it up in the ``tearDown`` method.
+
+Our test uses ``zc.buildout`` so that we can use the
+``zc.buildout.testing.Buildout`` helper class.  We did this so we'd
+have a more realistic environment, but of course, we could have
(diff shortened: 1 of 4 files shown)


doc/topics/writing-recipes.rst
```diff
@@ -337,6 +342,267 @@ entry points, using the type ``zc.buildout.uninstall`` as can be seen
...
+  import shutil
+  import tempfile
+  import unittest
+  import zc.buildout.testing
+
+  class RecipeTests(unittest.TestCase):
+
+      def setUp(self):
+          self.here = os.getcwd()
+          self.tmp = tempfile.mkdtemp()
```
mgedmin: I always recommend passing a prefix to `mkdtemp()`, to make it easier to track down the source in case some test forgets to clean up:
```
self.tmp = tempfile.mkdtemp(prefix='testdemoconfigrecipe-')
```

doc/topics/writing-recipes.rst
```diff
@@ -337,6 +342,267 @@ entry points, using the type ``zc.buildout.uninstall`` as can be seen
...
+``eqs(got, *expected)``
+    Compare a collection with a collection given as multiple
+    arguments.
+
+    Both collections are converted to and compared as sets.  If the
+    sets are the same, then no output is returned, otherwise a tuple
+    of extras is returned, so, for example::
+
+      >>> eqs([1, 2, 3], 3, 1, 2)
+      >>> eqs([1, 2, 3], 1, 2, 4) == (set([3]), set([4]))
```
mgedmin: We no longer support Python 2.6 so we can use the set syntax here for perhaps more clarity?  `== ({3}, {4})`.

OTOH that doesn't match the repr() of the sets returned, so maybe the current syntax is better.

doc/topics/writing-recipes.rst
```diff
@@ -337,6 +342,267 @@ entry points, using the type ``zc.buildout.uninstall`` as can be seen
...
+    more strings, to be joined with os.path.join.
+
+``mkdir(*path)``
+    Create a directory. The directory path is provided as one or
+    more strings, to be joined with os.path.join.
+
+``system(command, input='')``
+    Execute a system command with the given input passed to the
+    command's standard input.  The output (error and regular output)
+    from the command is returned.
```
mgedmin: It's a bit unclear to me, from this description, whether stdout and stderr are returned intermixed in one single string, or whether you get a tuple (error_output, regular_output).

doc/topics/writing-recipes.rst
```diff
@@ -337,6 +342,267 @@ entry points, using the type ``zc.buildout.uninstall`` as can be seen
...
+
+``read(*path)``
+    Read data from a file at the given path.  The file path is
+    provided as one or more strings, to be joined with os.path.join.
+
+    If no path is given, the ``'out'`` us used.
```
mgedmin: Typo: us -> is

doc/topics/writing-recipes.rst
```diff
@@ -337,6 +342,267 @@ entry points, using the type ``zc.buildout.uninstall`` as can be seen
...
+    Create a directory. The directory path is provided as one or
+    more strings, to be joined with os.path.join.
+
+``system(command, input='')``
+    Execute a system command with the given input passed to the
+    command's standard input.  The output (error and regular output)
+    from the command is returned.
+
+``read(*path)``
+    Read data from a file at the given path.  The file path is
```
mgedmin: Perhaps "Read text from a file", since it opens the file in text mode and returns a string, rather than bytes.

src/zc/buildout/testing.py
```diff
@@ -56,6 +61,17 @@ def cat(dir, *names):
     with open(path) as f:
         print_(f.read(), end='')
 
+def eqs(a, *b):
+    a = set(a); b = set(b)
+    return None if a == b else (a - b, b - a)
+
+def clear_here():
+    for name in os.listdir('.'):
+        if os.path.isfile(name):
```
mgedmin: I suggest adding `or os.path.islink(name)`, otherwise you can get an `OSError: Cannot call rmtree on a symbolic link` in the else branch.  (Which is much better than my $HOME getting accidentally erased, so kudos to Python stdlib developers!)

src/zc/buildout/testing.py
```diff
@@ -541,3 +564,37 @@ def _normalize_path(match):
     re.compile(
     'Not upgrading because not running a local buildout command.\n'
     ), '')
+
+def run_buildout(command):
+    # Make sure we don't get .buildout
+    os.environ['HOME'] = os.path.join(os.getcwd(), 'home')
+    args = command.strip().split()
```
mgedmin: Can be simplified to `command.split()`, since `.split()` with no arguments implicitly strips.

src/zc/buildout/tests.py
```diff
@@ -928,7 +927,7 @@ def add_setuptools_to_dependencies_when_namespace_packages():
 
 On the other hand, if the distribution uses ``pkgutil.extend_path()`` to
 implement its namespaces, even if just as fallback from the absence of
-``pkg_resources``, then ``setuptools`` shoudn't be added as requirement to
+``pkg_resources``, then ``setuptools`` shoud not be added as requirement to
```
mgedmin: Typo: shoud -> should

doc/topics/writing-recipes.rst
jimfulton: Fair enough. I was trying to keep things simple but I'll add this.

doc/topics/writing-recipes.rst
jimfulton: Whether it matches the repr depends on Python version.  I'll change it.

doc/topics/writing-recipes.rst
jimfulton: OK, added some words:

(error and regular output combined into a single string)

jimfulton:
Thanks for the review!

PR #369, opened 2017-03-06
Meta recipes

jimfulton:
This just moves the existing documentation from the buildout package into the manual.

The original test was written as documentation using manuel.


jimfulton:
Thanks!

PR #373, opened 2017-03-06
Always unpack what `wheel_to_egg()` generates

leorochael:
The `unpack_wheel` was blindly trusting the result of `wheel_to_egg()`
and possibly laying zipped eggs in `eggs-directory`.

Fix it to always unpack the result of `wheel_to_egg()` instead.


leorochael:
@jimfulton my bad, didn't properly test the backward compatibility shim. It was working somewhat, but laying zipped eggs in `eggs_directory`.

Issue #375, opened 2017-03-06
AttributeError: 'module' object has no attribute '_by_version_descending' on 2.9.0


mauritsvanrees:
I just saw this [Travis error on a Plone package](https://travis-ci.org/collective/collective.catalogcleanup/jobs/208286558):

```
While:
  Installing.
  Checking for upgrades.
An internal error occurred due to a bug in either zc.buildout or in a
recipe being used:
Traceback (most recent call last):
  File "/home/travis/build/collective/collective.catalogcleanup/eggs/zc.buildout-2.9.0-py2.7.egg/zc/buildout/buildout.py", line 1982, in main
    getattr(buildout, command)(args)
  File "/home/travis/build/collective/collective.catalogcleanup/eggs/zc.buildout-2.9.0-py2.7.egg/zc/buildout/buildout.py", line 517, in install
    self._maybe_upgrade()
  File "/home/travis/build/collective/collective.catalogcleanup/eggs/zc.buildout-2.9.0-py2.7.egg/zc/buildout/buildout.py", line 953, in _maybe_upgrade
    allow_hosts = self._allow_hosts
  File "/home/travis/build/collective/collective.catalogcleanup/eggs/zc.buildout-2.9.0-py2.7.egg/zc/buildout/easy_install.py", line 913, in install
    check_picked=check_picked)
  File "/home/travis/build/collective/collective.catalogcleanup/eggs/zc.buildout-2.9.0-py2.7.egg/zc/buildout/easy_install.py", line 265, in __init__
    self._env = self._make_env()
  File "/home/travis/build/collective/collective.catalogcleanup/eggs/zc.buildout-2.9.0-py2.7.egg/zc/buildout/easy_install.py", line 274, in _make_env
    full_path = self._get_dest_dist_paths() + self._path
  File "/home/travis/build/collective/collective.catalogcleanup/eggs/zc.buildout-2.9.0-py2.7.egg/zc/buildout/easy_install.py", line 293, in _get_dest_dist_paths
    return pkg_resources._by_version_descending(set(eggs + dists))
AttributeError: 'module' object has no attribute '_by_version_descending'
```

It fails in code that was introduced in the fresh `zc.buildout` 2.9.0.
The buildout works fine locally.
I haven't dived into it yet and I gotta run now.

jimfulton:
What version of setuptools is being used?

jimfulton:
I suspect we need to add a minimum setuptools version.

jimfulton:
[...]

PR #376, opened 2017-03-06
Drop use of pkg_resources._by_version_descending()

leorochael:
It's not present in older setuptools versions.

It seems not to be needed since `pkg_resources.Environment.add()` will
sort dists in each project_name according to their version anyway.

Fix: #375


jimfulton:
Thanks. /me waits for travis to go green

PR #386, opened 2017-05-04
Respect index url and links when installing setup-required dependencies

janwijbrand:
In our use-case we allow installing package only through our own package index (proxying pypi for public package and as a repository for internal private package). Reason is some of the production machines are not allowed to go out to the internet.

When installed a package that itself has a "setup requirement" (in our concrete use-case it is the "cryptography" package, i.e. having dependencies for compiling sources), buildout would not pass the index URL nor "find links" on to the easy install calls. 

This PR fixes that albeit maybe in an naive fashion. I'd like to get some feedback on:

a) If this is anywhere near an acceptable approach. If not, perhaps someone can suggest a better approach?
b) How to write a tests for this. I tested locally by adding an entry for `pypi.python.org` pointing to `127.0.0.1` in my `/etc/hosts`. How would we mimic not being able to access anything but the index URL from the tests?
c) If the PR is against the correct branch 

Help and feedback is highly appreciated.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,8 +4,8 @@ Change History
 2.9.4 (unreleased)
 ==================
 
-- Nothing changed yet.
-
+- Respect index-url and find-links when calling easy install for
+  setup-required dependencies.
 
 2.9.3 (2017-03-30)
 ==================
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -1593,11 +1593,11 @@ def call_easy_install(spec, dest, index=None, links=None):
         '-mZUNx']
 
     if index is not None:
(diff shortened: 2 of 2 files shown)


jcbrand:
Concerning tests, we could instantiate an installer and then call `installer.build` so that `_call_easy_install` is called and eventually `call_easy_install`.

Then one could use python's mock library to patch the `subprocess.call` method to check whether the right values were passed in.

Would that make sense?

janwijbrand:
Can someone help me with a hint for getting this PR in a "mergable" state?

idgserpro:
@janwijbrand this is the same as #238? It would be great! We had trouble with this:

https://github.com/Julian/jsonschema/issues/453#issuecomment-413936081

@jonathanultis FYI

PR #393, opened 2017-05-25
Sort distributions when computing a signature.

### LINKED ISSUE #392: Signature comparison broken when PYTHONHASHSEED=random
I discovered this when trying to port zc.recipe.cmmi to Python 3. I have the environmnet variable PYTHONHASHSEED=random set for the sake of Python 2 (because on Python 3, [hash randomization is on by default](https://docs.python.org/3/using/cmdline.html#cmdoption-R))  and was getting a test failure when the tests wanted to verify that re-installing an installed part did nothing:
```
Failed example:
    print(system('bin/buildout').strip())
Expected:
    Updating foo.
Got
    Uninstalling foo.
    Installing foo.
    foo: Downloading http://localhost/foo.tgz
    foo: Unpacking and configuring
    configuring foo --prefix=/sample_buildout/parts/foo -a -b c
    echo building foo
    building foo
    echo installing foo
    installing foo
```

Cranking up the verbosity of that command revealed this output:
```
Part foo, option __buildout_signature__ changed:
'zc.recipe.cmmi-c239cab0189d8edcfbbb196af19c987c setuptools-1b99ced40b029a77bcf7fde7d625fbd5 zc.buildout-1b99ced40b029a77bcf7fde7d625fbd5 packaging-1b99ced40b029a77bcf7fde7d625fbd5 appdirs-1b99ced40b029a77bcf7fde7d625fbd5 six-1b99ced40b029a77bcf7fde7d625fbd5 pyparsing-1b99ced40b029a77bcf7fde7d625fbd5' != 'zc.recipe.cmmi-1ae0794e23637cc24ee1587b1d972e9e setuptools-1b99ced40b029a77bcf7fde7d625fbd5 zc.buildout-1b99ced40b029a77bcf7fde7d625fbd5 appdirs-1b99ced40b029a77bcf7fde7d625fbd5 packaging-1b99ced40b029a77bcf7fde7d625fbd5 six-1b99ced40b029a77bcf7fde7d625fbd5 pyparsing-1b99ced40b029a77bcf7fde7d625fbd5'
```

Unsetting PYTHONHASHSEED solved the problem. Obviously the ordering is slightly different, but some of the hashes are different too.

I don't yet know if I'll run into the same problem on Python 3 or if something else changed that resolves it.

jamadden:
This fixes #392 (as far as I can test locally with zc.recipe.cmmi).

Previously under Python 3 (probably prior to 3.6, when dicts became ordered by default) or under Python 2 with ``-R`` or ``PYTHONHASHSEED=random`` the results were undefined because dict
iteration order is undefined.

The value being sorted is complex and defined by pkg_resources.Distribution:

```python
    @property
    def hashcmp(self):
        return (
            self.parsed_version,
            self.precedence,
            self.key,
            _remove_md5_fragment(self.location),
            self.py_version or '',
            self.platform or '',
        )

    def __hash__(self):
        return hash(self.hashcmp)

    def __lt__(self, other):
        return self.hashcmp < other.hashcmp
...
```

Note that there may be certain circumstances where this will cause existing `.installed.cfg` to be considered outdated spuriously, most likely under plain Python 2.

I had issues using tox to run the tests locally, so this is a fingers-crossed PR.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,14 @@ Change History
 2.9.4 (unreleased)
 ==================
 
-- Nothing changed yet.
+- Sort the distributions used to compute ``__buildout_signature__`` to
+  ensure reproducibility under Python 3 or under Python 2 when ``-R``
+  is used on ``PYTHONHASHSEED`` is set to ``random``. Fixes `issue 392
+  <https://github.com/buildout/buildout/issues/392>`_.
+
+  **NOTE**: This may cause existing ``.installed.cfg`` to be
+  considered outdated and lead to parts being reinstalled spuriously
+  under Python 2.
 
 
 2.9.3 (2017-03-30)
--- a/src/zc/buildout/buildout.py
+++ b/src/zc/buildout/buildout.py
@@ -1821,7 +1821,7 @@ def _dir_hash(dir):
 def _dists_sig(dists):
     seen = set()
     result = []
-    for dist in dists:
+    for dist in sorted(dists):
         if dist in seen:
             continue
         seen.add(dist)


jimfulton:
Interesting. Thanks!

jamadden:
Thank you! A PyPI release will be helpful for zopefoundation/zc.recipe.cmmi#2, when convenient.

Issue #398, opened 2017-06-16
Setuptools 36 once again vendors all its deps, breaks buildout tests


jamadden:
Setuptools 36 (May 30th) is now [vendoring everything it needs again](https://github.com/pypa/setuptools/issues/980). The [last build](https://travis-ci.org/buildout/buildout/builds/236023662) before that release was May 25th, and it passed. The next [build after that release](https://travis-ci.org/buildout/buildout/builds/242963450) was yesterday, and it fails with errors like:
```
File "/home/travis/build/buildout/buildout/src/zc/buildout/buildout.txt", line 57, in buildout.txt
Failed example:
    ls(sample_buildout, 'eggs')
Expected:
    d...
    setuptools.egg
    ...
    -  zc.buildout.egg
Got:
    setuptools.egg
    -  zc.buildout.egg
```

Unfortunately it looks like many of the changes in https://github.com/buildout/buildout/commit/e129e187c1cde18b80e4d1d70c4751dfd55dc071 to work with the non-vendored setuptools are now broken with the newly-re-vendored setuptools 😢

jamadden:
I'm working on a PR to fix these tests.

PR #399, opened 2017-06-16
Support coverage for subprocess.

### LINKED ISSUE #397: Add coverage support to zc.buildout.testing
Because zc.buildout.testing encourages running buildout in subprocesses in a temporary directory, there are a few boilerplate steps that have to be taken in order to accurately measure coverage across all the test cases:

1. Export `COVERAGE_PROCESS_START=/path/to/.coveragerc`. It needs to specify `parallel=True`.
2. Normalize `COVERAGE_PROCESS_START` to have an absolute path since we'll be working in a temporary directory.
2. Arrange for the generated `bin/buildout` to invoke `import coverage; coverage.process_startup()` at startup. 
3. Before destroying the temporary directory, copy `.coverage.*` back to the location holding the `.coveragerc`.

zc.buildout.testing provides setup and tear down functions that would be ideal places to do steps 2-4, meaning that for recipes to get coverage (if they use this test pattern) they would only need to write a `.coveragerc` and export the variable. (Or I suppose buildout could look for a .coveragerc in the current directory and automatically export that variable...but users may not always want coverage, so it seems like checking for the presence of this variable is a good way to gate the feature.)

I already have an implementation of these steps in [zc.recipe.cmmi#2](https://github.com/zopefoundation/zc.recipe.cmmi/pull/2/). 

I would be happy to extract them as a PR to move to buildout proper. Is there any interest in seeing this done?

It looks like many of buildout's own tests are also based on this test pattern, so this could facilitate getting coverage reports enabled for buildout too.

jamadden:
Fixes #397.

tox will produce the coverage reports by default.

Travis currently has the coverage disabled because it wants to work in a completely isolated python (can it be changed to invoke tox, maybe?).

Unfortunately because of #398 it's difficult to prove that it works. I have tested it with zopefoundation/zc.recipe.cmmi#2, though, and it does work there.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,9 @@
+[run]
+source = zc.buildout
+parallel = true
+
+[report]
+exclude_lines =
+    pragma: no cover
+    if __name__ == '__main__':
+    raise NotImplementedError
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ python*/
 *.egg-info
 dist
 .tox
+.coverage
+.coverage.*
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ notifications:
 
 install:
     - ls -al pythons
+    - export COVERAGE_PROCESS_START=
     - deactivate
     - make build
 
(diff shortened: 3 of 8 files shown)


jamadden:
I've rebased on current master.

jimfulton:
Thanks!

jamadden:
Thank you! A PyPI release would be helpful when convenient.

jimfulton:
Released

PR #402, opened 2017-07-26
Fix #153: buildout should cache working set environments

### LINKED ISSUE #153: buildout should cache working set environments
A while back @rpatterson worked on a patch in env-cache (https://github.com/buildout/buildout/commits/env-cache) which has a large performance improvement in buildouts with similar parts. I tested this patch for awhile and found no problems with it. 
Was there found to be a problem with it? otherwise I will update the branch with the latest master and submit a pull request?

rafaelbco:
This PR is meant to address the problem discussed here: https://community.plone.org/t/multiple-zope-instances-slow-buildout

Review is much appreciated because I'm new to buildout development (I'm a buildout heavy user, though).

### DIFF
--- a/zc.recipe.egg_/src/zc/recipe/egg/egg.py
+++ b/zc.recipe.egg_/src/zc/recipe/egg/egg.py
@@ -19,7 +19,7 @@ import os
 import re
 import sys
 import zc.buildout.easy_install
-import zipfile
+
 
 class Eggs(object):
 
@@ -43,13 +43,13 @@ class Eggs(object):
 
         allow_hosts = b_options['allow-hosts']
         allow_hosts = tuple([host.strip() for host in allow_hosts.split('\n')
-                               if host.strip()!=''])
+                            if host.strip() != ''])
         self.allow_hosts = allow_hosts
 
         options['eggs-directory'] = b_options['eggs-directory']
-        options['_e'] = options['eggs-directory'] # backward compat.
+        options['_e'] = options['eggs-directory']  # backward compat.
         options['develop-eggs-directory'] = b_options['develop-eggs-directory']
-        options['_d'] = options['develop-eggs-directory'] # backward compat.
+        options['_d'] = options['develop-eggs-directory']  # backward compat.
 
     def working_set(self, extra=()):
         """Separate method to just get the working set
@@ -57,7 +57,6 @@ class Eggs(object):
         This is intended for reuse by similar recipes.
         """
         options = self.options
-        b_options = self.buildout['buildout']
 
         # Backward compat. :(
         options['executable'] = sys.executable
@@ -91,13 +90,14 @@ class Eggs(object):
 
     update = install
 
+
 class Scripts(Eggs):
 
     def __init__(self, buildout, name, options):
         super(Scripts, self).__init__(buildout, name, options)
 
         options['bin-directory'] = buildout['buildout']['bin-directory']
-        options['_b'] = options['bin-directory'] # backward compat.
+        options['_b'] = options['bin-directory']  # backward compat.
 
         self.extra_paths = [
             os.path.join(buildout['buildout']['directory'], p.strip())
@@ -106,7 +106,6 @@ class Scripts(Eggs):
             ]
         if self.extra_paths:
             options['extra-paths'] = '\n'.join(self.extra_paths)
-
 
         relative_paths = options.get(
             'relative-paths',
@@ -122,6 +121,7 @@ class Scripts(Eggs):
     parse_entry_point = re.compile(
         '([^=]+)=(\w+(?:[.]\w+)*):(\w+(?:[.]\w+)*)$'
         ).match
+
     def install(self):
         reqs, ws = self.working_set()
         options = self.options
@@ -165,6 +165,7 @@ class Scripts(Eggs):
         return ()
 
     update = install
+
 
 def get_bool(options, name, default=False):
     value = options.get(name)
--- a/zc.recipe.egg_/src/zc/recipe/egg/egg.py
+++ b/zc.recipe.egg_/src/zc/recipe/egg/egg.py
@@ -14,6 +14,7 @@
 """Install packages as eggs
 """
 
+import copy
 import logging
 import os
 import re
@@ -22,6 +23,8 @@ import zc.buildout.easy_install
 
 
 class Eggs(object):
+
+    _WORKING_SET_CACHE_ATTR_NAME = '_zc_recipe_egg_working_set_cache'
 
     def __init__(self, buildout, name, options):
         self.buildout = buildout
@@ -57,30 +60,27 @@ class Eggs(object):
         This is intended for reuse by similar recipes.
         """
         options = self.options
+        buildout_section = self.buildout['buildout']
 
         # Backward compat. :(
         options['executable'] = sys.executable
 
-        distributions = [
+        orig_distributions = [
             r.strip()
             for r in options.get('eggs', self.name).split('\n')
-            if r.strip()]
-        orig_distributions = distributions[:]
-        distributions.extend(extra)
+            if r.strip()
+            ]
 
-        if self.buildout['buildout'].get('offline') == 'true':
-            ws = zc.buildout.easy_install.working_set(
-                distributions,
-                [options['develop-eggs-directory'], options['eggs-directory']]
-                )
-        else:
-            ws = zc.buildout.easy_install.install(
-                distributions, options['eggs-directory'],
-                links=self.links,
-                index=self.index,
-                path=[options['develop-eggs-directory']],
-                newest=self.buildout['buildout'].get('newest') == 'true',
-                allow_hosts=self.allow_hosts)
+        ws = self._working_set(
+            distributions=orig_distributions + list(extra),
+            develop_eggs_dir=options['develop-eggs-directory'],
+            eggs_dir=options['eggs-directory'],
+            offline=(buildout_section.get('offline') == 'true'),
+            newest=(buildout_section.get('newest') == 'true'),
+            links=self.links,
+            index=self.index,
+            allow_hosts=self.allow_hosts,
+            )
 
         return orig_distributions, ws
 
@@ -89,6 +89,73 @@ class Eggs(object):
         return ()
 
     update = install
+
+    def _working_set(
+        self,
+        distributions,
+        eggs_dir,
+        develop_eggs_dir,
+        offline=False,
+        newest=True,
+        links=(),
+        index=None,
+        allow_hosts=('*',),
+    ):
+        """Helper function to build a working set.
+
+        Return an instance of `pkg_resources.WorkingSet`.
+
+        Results are cached. The cache key is composed by all the arguments
+        passed to the function. See also `self._get_cache_storage()`.
+        """
+        cache_storage = self._get_cache_storage()
+        cache_key = (
+            tuple(distributions),
+            eggs_dir,
+            develop_eggs_dir,
+            offline,
+            newest,
+            tuple(links),
+            index,
+            tuple(allow_hosts),
+        )
+        if cache_key not in cache_storage:
+            if offline:
+                ws = zc.buildout.easy_install.working_set(
+                    distributions,
+                    [develop_eggs_dir, eggs_dir]
+                    )
+            else:
+                ws = zc.buildout.easy_install.install(
+                    distributions, eggs_dir,
+                    links=links,
+                    index=index,
+                    path=[develop_eggs_dir],
+                    newest=newest,
+                    allow_hosts=allow_hosts)
+            cache_storage[cache_key] = ws
+
+        # `pkg_resources.WorkingSet` instances are mutable, so we need to return
+        # a copy.
+        return copy.deepcopy(cache_storage[cache_key])
+
+    def _get_cache_storage(self):
+        """Return a mapping where to store generated working sets.
+
+        The cache storage is stored in an attribute of `self.buildout` with
+        name given by `self._WORKING_SET_CACHE_ATTR_NAME`.
+        """
+        cache_storage = getattr(
+            self.buildout,
+            self._WORKING_SET_CACHE_ATTR_NAME,
+            None)
+        if cache_storage is None:
+            cache_storage = {}
+            setattr(
+                self.buildout,
+                self._WORKING_SET_CACHE_ATTR_NAME,
+                cache_storage)
+        return cache_storage
 
 
 class Scripts(Eggs):
--- a/zc.recipe.egg_/CHANGES.rst
+++ b/zc.recipe.egg_/CHANGES.rst
@@ -4,7 +4,8 @@ Change History
 2.0.4 (unreleased)
 ==================
 
-- Nothing changed yet.
+- Fix #153: buildout should cache working set environments
+  [rafaelbco]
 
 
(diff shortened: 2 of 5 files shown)


zc.recipe.egg_/src/zc/recipe/egg/egg.py
```diff
@@ -57,40 +57,34 @@ def working_set(self, extra=()):
...
+            extra=extra,
+            )
 
     def install(self):
         reqs, ws = self.working_set()
         return ()
 
     update = install
 
+
```
rafaelbco: I agree with the suggestions. But how can I do that? My git-fu is weak. Perhaps close this PR and do a new one?

zc.recipe.egg_/src/zc/recipe/egg/egg.py
```diff
@@ -166,6 +159,60 @@ def install(self):
...
+    extra=(),
+):
+    """Helper function to build a working set.
+
+    Results are cached in the `_working_set_cache` dict at module level.
+    The cache key is composed by all the arguments passed to the function.
+    """
+    orig_distributions = list(distributions)
+    cache_key = (
+        tuple(distributions),
```
rafaelbco: Can the order of distributions affect the generated working set? If not then definitely the list should be sorted.

zc.recipe.egg_/src/zc/recipe/egg/egg.py
```diff
@@ -166,6 +159,60 @@ def install(self):
...
+            ws = zc.buildout.easy_install.install(
+                distributions, eggs_dir,
+                links=links,
+                index=index,
+                path=[develop_eggs_dir],
+                newest=newest,
+                allow_hosts=allow_hosts)
+        _working_set_cache[cache_key] = ws
+
+    return orig_distributions, _working_set_cache[cache_key]
```
rafaelbco: Understood, I'll do that. And thanks for the feedback!

zc.recipe.egg_/src/zc/recipe/egg/egg.py
leorochael: > I agree with the suggestions. But how can I do that? My git-fu is weak. Perhaps close this PR and do a new one?

It's not necessary to close this PR. As long as this PR is open, whatever is on `rafaelbco:master` will be what this PR represents, but you'll need to do `git rebase -i upstream/master` to reorder your history and squash commits, then `git push -f` to force push the rewritten branch on top of the one in your github repository.

This assumes you have a git "remote" called `upstream` pointing to the original buildout repo.

Check some tutorials on `git interactive rebase` or `git rewrite history`.

In any case it's better to do the history rewriting as the last step, after replying to all change requests with pure additional commits.

But in the end, if after making all changes you're still uncomfortable doing git history rewriting, just give me permission to your repo and I'll rewrite the history for you. The easiest way to do that is by [allowing changes to a pull request](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/).

zc.recipe.egg_/src/zc/recipe/egg/egg.py
leorochael: > Can the order of distributions affect the generated working set?

That is a good question. In theory, there should be no difference as long as packages don't try to place modules in the same namespace as other packages and a smart-enough dependency resolver should try to find a set of dependencies for the requested packages no matter the order the packages are requested.

In practice I think the order matters because the resolver in `setuptools` is **not** smart enough and the specific ordering can mean the difference between a dependency resolution that works and one that doesn't, so leave the `distributions` unsorted for now.

rafaelbco:
@leorochael I think I addressed all the issues in the code. Please review again.

The timing comparisons are still pending though.

rafaelbco:
"Allow edits from maintainers" is enabled so feel free to change whatever you want.

leorochael:
Hi Rafael,

I've just added a commit (to be squashed later, in case you agree), fixing some formatting and wording.

But I just had another idea: The `working_set()` top level function, doesn't actually do anything with `orig_distributions` and `extra` other than jamming them together and then returning `orig_distributions as is.

So, what if instead of taking `extras` and returning a pair of `orig_distributions` and a working set, the top level function accepted the list of distributions already concatenated, and returned just the working set? The method version would do the concatenation of `orig_distributions` and `extra` and would still return the `orig_distributions, ws` pair.

The documentation you added would need to be adjusted **not** to say it does the same thing as the `working_set()` method, but the top level function itself would be simpler, with a simpler signature and simpler return value and its name would match better what it does.

What do you think?

rafaelbco:
I think you're right. This way the module level function would make an even better addition to the API (besides being used for making the caching easier). 

I'll work on it.

rafaelbco:
@leorochael I made the changes requested to the `working_set` function and I'm working on the timing comparisons.

leorochael:
Looks good to me!

@jimfulton, any comments?

jamadden:
I wonder if instead of the module-global `_working_set_cache`, the cache could be on the buildout instance object itself? That way we have isolation of the cache to a particular buildout run. 

If I understand what is being cached, then in practice (running buildout on the command line) I wouldn't expect that to make any difference, but for third-party tests (or anything that may invoke buildout APIs programmatically more than once in a single process) there's no (or less) concern about cache pollution/corruption.

(E.g., for this reason, if a zopefoundation project has to use a module-global cache (because there is no scoped argument to hang it on) it tends to include cleanup code like so:
```python
_cache = {}
try:
   from zope.testing import cleanup
except ImportError:
   pass
else:
   cleanup.addCleanUp(_cache.clear())
```
)

rafaelbco:
Timing comparisons: https://github.com/rafaelbco/buildout_improvements/tree/master/timings_ws_caching

zc.recipe.egg_/src/zc/recipe/egg/egg.py
```diff
@@ -177,14 +185,21 @@ def working_set(
...
+    cache_storage=_global_working_set_cache,
 ):
     """Helper function to build a working set.
 
     Return an instance of `pkg_resources.WorkingSet`.
 
-    Results are cached in the `_working_set_cache` dict at module level.
+    Results are cached in the mapping provided in the `cache_storage` argument.
+    This defaults to a global dict at this module. If `None` is passed then
+    caching is disabled.
```
jamadden: Is there a reason to even have the option of a module global dict cache? That seems like a foot-gun to me.

rafaelbco:
I made the proposed changes. 

I chose to store the cache as an attribute of the buildout object. Storing it as an option seemed kind of strange to me.  I used `__zc_recipe_egg_working_set_cache__`  as the attribute name to avoid name clashes and make debugging easier. The `__XXX__` syntax makes it clear it is an special thing.

> This would imply that either the function that caches working sets goes back to being a method of the Egg recipe class or that it remains module global, but accepts an extra argument: the cache dictionary itself.

I like having `working_set()` as a module function with a simpler signature and return value. I think it's useful as a separate function.. So I kept the function at module level and added an optional `cache_storage` argument. If not provided then it uses the global module level dict as before. You can even pass `None` to disable caching.


> @rafaelbco, I hope the repeated request for changes doesn't discourage you.
> I think your proposed improvement is nice, and each additional opinion has made it better.

Thanks!

zc.recipe.egg_/src/zc/recipe/egg/egg.py
rafaelbco: The reason is I think the `working_set()` function at module level is useful on its own. And if people choose to use it then it's nice to have caching enabled by default, since building working sets is slow. 

I think it's better to do this than putting the burden of setting up the cache storage in the callers.

For example, check out this code: https://github.com/plone/plone.recipe.zope2instance/blob/master/src/plone/recipe/zope2instance/__init__.py#L620

I think it would be nice to use the `working_set()` function here, instead of instantiating the `Egg` class.

Remember caching can be disable if needed, passing `cache_storage=None`.

jamadden: > The reason is I think the working_set() function at module level is useful on its own.
> if people choose to use it then it's nice to have caching enabled by default, since building working sets is slow.

> I think it's better to do this than putting the burden of setting up the cache storage in the callers.

Personally, I happen to disagree, I think module level caches are dangerous by default. Moreover...:

> I think it would be nice to use the working_set() function here, instead of instantiating the Egg class.

That seems to disprove the point. That recipe *already* has an Egg instance, which has the buildout, etc, and so access to the scoped cache. That's been the only way to do it so far, and I don't see a reason to introduce another way to do it ("There should be one-- and preferably only one --obvious way to do it."), especially if that way is potentially dangerous.

rafaelbco: So, I see two different and independent issues brought by you here. 

First is: 

> module level caches are dangerous by default

It's fixable by not providing a default value for the `cache_storage` argument. I don't agree with this change because I think the benefits are greater than the danger.

Second is: 

> I don't see a reason to introduce another way to do it.

It's fixable by removing the `working_set()` module level function from the API. Probably making it a non-public method named `_working_set()` or `_build_working_set()`. 

I don't like this change because I'd rather use `working_set()` as a function than instantiate the `Egg` class just to call the method. I think the module level function is a better API.

So, if I'm right, and the new proposed API is better then we have to solve the problem you mentioned: now we have two ways of doing the same thing. We could go ahead and remove the `Egg.working_set()` method from the API. 

If we would adhere strictly to "There should be one-- and preferably only one --obvious way to do it." then we could as well go ahead and remove the `Egg.working_set()` method from the API. However it seems unwise to break backwards compatibility in a PR about performance. So I propose we violate "There should be one...".

Since:

1. I want the PR approved because of the performance gains. The API improvement is a small bonus for me.
2. I don't agree with the two changes above, but they don't impact the performance gains.

Then I'll ask kindly for @leorochael  and @jamadden decide together what changes needs to be done and I'll made them (provided they are in the scope I delimited above). 

It would be even better if you do the changes yourselves, after agreeing on what needs to be changed.

leorochael:
@rafaelbco, I've done a lot of thinking this last weekend on the problem we're trying to solve and the different options and opinions, and how to solve them not through a compromise that'll leave some or all of us only partially dissatisfied, but through a solution that is better overall.

The functionality we're trying to provide is comprised of an entry point (currently a module level function), and some state that exists outside of the scope of a single run of this entry point (the cache). The needed lifetime of this state is greater than the lifetime of the run of a single buildout part, so multiple parts can make use of the cache, but it doesn't need to be longer than the lifetime of the buildout run itself, so it doesn't require module-level state.

At this point the remaining reason for a module level state is simplification of the API, but as @jamadden, pointed out, by having `zc.recipe.egg.egg:Eggs` make use of it in its `.working_set()` method, no further changes are needed in any other recipe, so the benefits of a simplified API are restricted to its use inside `zc.recipe.egg` itself.

(Incidentally, `plone.recipe.zope2instance:Recipe` is a weird example in that it both inherits from `zc.recipe.egg:Eggs` (thru `zc.recipe.egg.Scripts`) and also makes use of an instance of `zc.recipe.egg:Eggs` (thru `zc.recipe.egg.Egg`) as an attribute. Only one of these would have been necessary, and again, no changes to it are necessary to make use of the cache, as long as it keeps using the `.working_set()` method).

Now the usually Pythonic way of having "some state that exists outside of the scope of a single function run" and "an entry point for a functionality" (preferably one that is simple to use) is to create a class (say, `WorkingSetCache`) that encapsulates both the state and the entry point.

And since the needed lifetime of the cache is contained in the lifetime of the buildout run, it seems to me the better place to put the cache is as an instance of the buildout object, instantiated by the buildout instance itself, so that it's simply used by recipes, instead of being instantiated by recipes on demand (it could perhaps even be used by buildout itself to cache its own working sets, but this is a massive scope creep).

I'm still weighing if the added complexity of the above paragraph is worth the effort (it might be if we could make the classes in `zc.recipe.egg` much simpler by simply passing all their options into the constructor or methods of the cache instance), but I didn't want @rafaelbco thinking that we just dropped the matter.

But if the additional complexity is not worth it, then I believe the simpler solution here is to get rid of the module level function and turn it into an internal `._working_set()` method in the recipe instance.

I will get around to one of these solutions sometime this week, but if any one is feeling impatient. I'd be willing to merge the `._working_set()` internal method solution if we have one other approval (@jimfulton?).

Regards,

rafaelbco:
@leorochael and @jamadden  

It's my first contribution to `zc.buildout` and I tried to touch as little as possible in order to make the caching work and not break anything. So far I could keep up with the requested changes. The last comment from @leorochael however mentions bigger changes that I feel I cannot tackle at the moment.

That said, I changed my mind about what I think needs to be done. Here's my new suggestion:

- I'll change the code to get rid of the module level function. The module level method will be turned into a `_working_set()` method.
- The cache will be stored in the `buildout` object, as a dict instantiated by the `_working_set()` method, named `__zc_recipe_egg_working_set_cache__`.

This won't change the API at all and achieve to goal of the PR, which is performance improvement.

So you merge the PR and do a new `zc.recipe.egg` release. After that you can improve the code, making the more complex proposed changes. Outside the scope of this PR.

What do you think?

leorochael:
> - I'll change the code to get rid of the module level function. The module level method will be turned into a `_working_set()` method.
> - The cache will be stored in the `buildout` object, as a dict instantiated by the `_working_set()` method, named `__zc_recipe_egg_working_set_cache__`.
> 

@rafaelbco, sounds good to me.

rafaelbco:
@leorochael Ready to be reviewed again!

leorochael:
> @leorochael Ready to be reviewed again!

@rafaelbco, looks good to me!

@jamadden, I expect your concerns have been met, right?

@jimfulton, are you ok, with merging this?

jamadden:
> @jamadden, I expect your concerns have been met, right?

Absolutely! Thanks @rafaelbco!

jimfulton:
I don't have time to look at this now. If @leorochael and @jamadden are good I'm good. :)

jimfulton:
Thanks @rafaelbco!

rafaelbco:
Thank you guys, it was a pleasure work with you.

rafaelbco:
Any plans on releasing on pypi?

leorochael:
> Any plans on releasing on pypi?

I tried today, but failed to use `zest.releaser` while trying to keep different sets of tags for `zc.buildout` and `zc.recipe.egg` due to a [ConfigParser issue](https://github.com/zestsoftware/zest.releaser/issues/212#issuecomment-322593853).

I'm afraid to try without `zest.releaser` and mess the release for now. If anyone else wants to have a go, fell free.

leorochael:
A new version of `zest.releaser` now gives better support for specifying the tag format.

@jimfulton, I don't think I have permission for adding `zc.buildout` our `zc.recipe.egg` releases to PyPI.

Could you please either create a new release or give me the permission to do so?

leorochael:
@jimfulton, thanks for the PyPI credentials.

@rafaelbco, after much wrangling with `zest.releaser` (including conjuring a new bugfix for it, and being forced to run it from python3), I managed to upload a new release of `zc.recipe.egg`.

It seems we're keeping the pace of releasing one every two years ;-)

jimfulton:
Thanks @leorochael !

rafaelbco:
Thanks @leorochael <https://github.com/leorochael> !


Rafael Oliveira

Issue #297, opened 2016-05-25
Drop Python 2.6 / 3.2 support


tseaver:
3.2 isn't supported by currrent setuptools;  2.6 is long out-of-maintenance.

do3cc:
Support for 2.6 has now de facto been removed, because the latest change and latest release has test failures with python 2.6. The fastest fix would be to remove python 2.6 support

do3cc:
I't seems 3.3 is also not supported any more. I've updated PR.

reinout:
I don't see it in the PR. Did you push it? Or is it perhaps because I already merged it?

And can you look at Tres' comment about updating the python 2.7 version?

do3cc:
I made a new PR with all patch versions updated. Python 3.3 still gets security fixes but it isn't listed on the plone releases page, so I kicked it.

mauritsvanrees:
The new PR #304 was merged too.
This issue can be closed.

PR #405, opened 2017-08-11
Update easy_install.py

rainp1ng:
line212: "{r.project_name for r in dist.requires()}" is an invalid syntax in Python 2.6, so (r.project_name for r in dist.requires()) would be better to make it compatible with Python 2.6


tseaver:
@rainp1ng Thanks for the thought you put into the patch!  However, we have [dropped support for Python 2.6](https://github.com/buildout/buildout/issues/297) as of version 2.6.0, and so don't need to use syntax which is compatible with it.

PR #417, opened 2017-09-21
Use HTTPS for default PyPI index

### LINKED ISSUE #114: Buildout should default to Pypi over HTTPS
ATM, the default in easy_install.py is http:

```
default_index_url = os.environ.get(
    'buildout-testing-index-url',
    'http://pypi.python.org/simple',
    )
```

davidjb:
PyPI redirects all requests from HTTP to HTTPS by default now so using HTTPS directly avoids the potential for that redirect being modified in flight, helping prevent MITM attacks.

Fixes #114.  The old comments over there aren't applicable any longer.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,9 @@ Change History
 2.9.5 (unreleased)
 ==================
 
-- Nothing changed yet.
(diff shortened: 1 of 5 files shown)


reinout:
Thanks for the fix!

reinout:
I've released 2.9.5 with the fix as it improves the security a bit.

PR #422, opened 2017-11-02
Update implicit-parts.rst

clarksun:
miss {


tseaver:
@clarksun Thank you for the patch!

PR #428, opened 2017-11-29
Install wheels via setuptools, fixes #425

### LINKED ISSUE #425: Error: Wheels are not supported with setuptools 38.2.0
Yesterday setuptools 38.2.0 came out with this interesting new feature:

- pypa/setuptools#1200: easy_install now support installing from wheels: they will be installed as standalone unzipped eggs.

Since then all my Jenkins jobs that run `bin/buildout` started failing with the following error

```
Getting distribution for 'zope.testrunner==4.7.0'.
Using unpack_wheel() shim over the deprecated wheel_to_egg() hook.
Please update your wheel extension implementation for one that installs a .whl
handler in zc.buildout.easy_install.UNPACKERS
While:
  Installing.
  Getting section test.
  Initializing section test.
  Installing recipe zc.recipe.testrunner.
  Getting distribution for 'zope.testrunner==4.7.0'.
Error: Wheels are not supported
```

Note: I'm not using buildout.wheel, and I've pinned the versions of every dependency and recipe, except for setuptools.

reinout:
Default to setuptools' wheel support instead of failing

Setuptools 38.2.0 suddenly started supporting and downloading wheels.
.whl files would end up in the non-functioning and deprecated unpack_wheel().
The new default ('just use setuptools') works fine.

- If you use an older setuptools, setuptools won't download wheels and the new
  code won't be used.

- If you use buildout.wheel, you'll overwrite the unpacker anyway.

Fixes #425

### DIFF
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -1624,7 +1624,7 @@ def unpack_egg(location, dest):
 
 WHEEL_WARNING = """
 *.whl file detected (%s), you'll need setuptools > 38.2.3 for that
-or an extension like buildout.wheel
+or an extension like buildout.wheel > 0.2.0.
 """
 
 
@@ -1633,7 +1633,7 @@ def unpack_wheel(location, dest):
         wheel = Wheel(location)
         wheel.install_as_egg(os.path.join(dest, wheel.egg_name()))
     else:
-        logger.warning(WHEEL_WARNING, location)
+        logger.error(WHEEL_WARNING, location)
 
 
 UNPACKERS = {
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -1637,7 +1637,7 @@ def unpack_wheel(location, dest):
         wheel = Wheel(location)
         wheel.install_as_egg(os.path.join(dest, wheel.egg_name()))
     else:
-        logger.error(WHEEL_WARNING, location)
+        raise zc.buildout.UserError(WHEEL_WARNING % location)
 
 
 UNPACKERS = {
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -1626,7 +1626,7 @@ def unpack_egg(location, dest):
 
 
 WHEEL_WARNING = """
-*.whl file detected (%s), you'll need setuptools > 38.2.0 for that
+*.whl file detected (%s), you'll need setuptools > 38.2.3 for that
 or an extension like buildout.wheel
 """
 
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -37,6 +37,12 @@ import tempfile
 import zc.buildout
 import warnings
 
+try:
+    from setuptools.wheel import Wheel
+    SETUPTOOLS_SUPPORTS_WHEELS = True
+except ImportError:
+    SETUPTOOLS_SUPPORTS_WHEELS = False
+
 warnings.filterwarnings(
     'ignore', '.+is being parsed as a legacy, non PEP 440, version')
 
@@ -1619,29 +1625,23 @@ def unpack_egg(location, dest):
     setuptools.archive_util.unpack_archive(location, dest)
 
 
-WHEEL_TO_EGG_WARNING = """
-Using unpack_wheel() shim over the deprecated wheel_to_egg() hook.
-Please update your wheel extension implementation for one that installs a .whl
-handler in %s.UNPACKERS
-""".strip() % (__name__,)
+WHEEL_WARNING = """
+*.whl file detected (%s), you'll need setuptools > 38.2.0 for that
+or an extension like buildout.wheel
+"""
+
 
 def unpack_wheel(location, dest):
-    # Deprecated backward compatibility shim. Please do not use.
-    logger.warning(WHEEL_TO_EGG_WARNING)
-    basename = os.path.basename(location)
-    dists = setuptools.package_index.distros_for_location(location, basename)
-    # `wheel_to_egg()` might generate zipped eggs, so we have to make sure we
-    # get unpacked eggs in the end:
-    tmp_dest = tempfile.mkdtemp(dir=dest)
-    wheel_to_egg(list(dists)[0], tmp_dest)
-    [egg] = glob.glob(os.path.join(tmp_dest, '*.egg'))
-    unpack_egg(egg, dest)
-    shutil.rmtree(tmp_dest)
+    if SETUPTOOLS_SUPPORTS_WHEELS:
+        wheel = Wheel(location)
+        wheel.install_as_egg(os.path.join(dest, wheel.egg_name()))
+    else:
+        logger.warning(WHEEL_WARNING, location)
 
 
 UNPACKERS = {
     '.egg': unpack_egg,
-    '.whl': unpack_egg,  # Since 38.2, setuptools handles wheels, too.
+    '.whl': unpack_wheel,
 }
 
 
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -1637,11 +1637,11 @@ def unpack_wheel(location, dest):
     [egg] = glob.glob(os.path.join(tmp_dest, '*.egg'))
     unpack_egg(egg, dest)
     shutil.rmtree(tmp_dest)
-    
+
 
 UNPACKERS = {
     '.egg': unpack_egg,
-    '.whl': unpack_wheel,
+    '.whl': unpack_egg,  # Since 38.2, setuptools handles wheels, too.
 }
 
 
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -89,9 +89,6 @@ setuptools_path = buildout_and_setuptools_path
 FILE_SCHEME = re.compile('file://', re.I).match
 DUNDER_FILE_PATTERN = re.compile(r"__file__ = '(?P<filename>.+)'$")
 
-def wheel_to_egg(dist, dest):
-    raise zc.buildout.UserError("Wheels are not supported")
-
 class _Monkey(object):
     def __init__(self, module, **kw):
         mdict = self._mdict = module.__dict__
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -1623,7 +1623,7 @@ def unpack_egg(location, dest):
 
 
 WHEEL_WARNING = """
-*.whl file detected (%s), you'll need setuptools > 38.2.3 for that
+*.whl file detected (%s), you'll need setuptools >= 38.2.3 for that
 or an extension like buildout.wheel > 0.2.0.
 """
 
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -38,8 +38,12 @@ import zc.buildout
 import warnings
 
 try:
-    from setuptools.wheel import Wheel
-    SETUPTOOLS_SUPPORTS_WHEELS = True
+    from setuptools.wheel import Wheel  # This is the important import
+    from setuptools import __version__ as setuptools_version
+    # Now we need to check if we have at least 38.2.3 for namespace support.
+    SETUPTOOLS_SUPPORTS_WHEELS = (
+        pkg_resources.SetuptoolsVersion(setuptools_version) >=
+        pkg_resources.SetuptoolsVersion('38.2.3'))
 except ImportError:
     SETUPTOOLS_SUPPORTS_WHEELS = False
 
--- a/src/zc/buildout/tests.py
+++ b/src/zc/buildout/tests.py
@@ -3144,53 +3144,6 @@ def test_buildout_doesnt_keep_adding_itself_to_versions():
 if sys.platform == 'win32':
     del buildout_honors_umask # umask on dohs is academic
 
-class UnitTests(unittest.TestCase):
-
-    @property
-    def globs(self):
-        return self.__dict__
-
-    def setUp(self):
-        easy_install_SetUp(self)
-        import setuptools.package_index
-        setuptools.package_index.EXTENSIONS.append('.whl')
-        import zc.buildout.easy_install
-        self.orig_wheel_to_egg = zc.buildout.easy_install.wheel_to_egg
-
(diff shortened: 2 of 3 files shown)


reinout:
I'm having all sorts of problems getting the tests to run locally.

On travis they're failing because they cannot find zope.testrunner. This *might* be a namespace package issue. Here's the python 3.5 output:

screen shot 2017-11-29 at 16 24 35

Note that there's a `zope.testrunner-4.8.1-py3.6-nspkg.pth` in the downloaded wheel. I find the "py3.6" a bit strange.

reinout:
Here's more debugging from the "make build" step of https://travis-ci.org/buildout/buildout/jobs/309039781

screen shot 2017-11-29 at 17 51 24

This is a `cat bin/test` with all the paths set OK. But still no working `import zope.testrunner`.
Might be that the namespace stuff isn't working OK yet? I'm not sure.

jamadden:
Hmm. I know that in general .whl *files* are not meant to be placed on sys.path. (Sometimes if you are lucky it works, though.) I see that the wheel file for zope.testrunner (which I probably created using python 3.6) appears to be unpacked to a directory, though. How does that compare to an unpacked egg? Or, is there any chance the .whl extension is confusing some part of the import machinery?

mgedmin:
I've tried running buildout's tests locally.  Master fails for me with that Error: wheels not supported.  Your branch gets stuck in buildout.txt forever (well, I ^Cd it after 5 minutes and 50 seconds; on Travis that particular test succeeded in 27 seconds).

(Curiously, I first forgot to `export COVERAGE_PROCESS_START=` before running `make test`, and buildout.txt did not get stuck, but rather showed a bunch of failures about `coverage` imports failing.)

reinout:
Ok, I got it to work locally in an almost-empty docker (that *might* be a nice way anyway to run the test).

What I found out: zope.exceptions and zope.interface are installed as eggs, only zope.testrunner is installed via a wheel. Zope.exceptions and zope.interface are importable, zope.testrunner not.

The problem is that zope.testrunner has no `__init__.py` in the `zope/` directory. If I create it, it is importable:

screen shot 2017-12-01 at 10 21 27

reinout:
Ah, the problem is a missing `__init__.py` in zope.testrunner's `zope/` dir.

See https://github.com/pypa/setuptools/issues/1214#issuecomment-348449705 for more info. It probably needs to be fixed in setuptools.

reinout:
Actually running the test also gives me all sorts of "failed to import coverage" errors. Doing `export COVERAGE_PROCESS_START=` first solves that.

But.... Why is that? And shouldn't that be in `DEVELOPERS.txt` if it is needed?

reinout:
I've hacked the `__init__.py` into my hacky temporary test-via-travis branch: https://github.com/buildout/buildout/tree/reinout-temp-travis-test

That looks much better.
https://travis-ci.org/buildout/buildout/jobs/309963146

Only one error, setuptools complaining about "invalid wheel name: 'demo-0.3.whl'".

benoit-pierre:
@reinout: I have not looked closely at your test, but `demo-0.3.whl` is indeed an invalid wheel name, so it will get ignored by setuptools, try `demo-0.3-py2.py3-none-any.whl` instead.

reinout:
Yes, I've corrected the name now: https://github.com/buildout/buildout/pull/428/commits/f603b11d1cbc132d6644f59ec26723e86941d74a

Something else goes wrong now, but that's probably regular test stuff.

src/zc/buildout/tests.py
```diff
@@ -3168,7 +3168,8 @@ def test_wheel_to_egg(self):
         [egg_name] = [n for n in os.listdir(self.sample_eggs)
                   if n.startswith('demo-0.3-')]
         path = os.path.join(self.sample_eggs, egg_name)
-        os.rename(path, os.path.join(self.sample_eggs, 'demo-0.3.whl'))
+        os.rename(path, os.path.join(self.sample_eggs,
+                                     'demo-0.3-py2.py3-none-any.whl'))
```
benoit-pierre: That's completely wrong! You can't create a wheel by just renaming an egg file! And even if the wheel is valid, you can't install it by just uncompressing it.

reinout: You're right.

The original os.rename() that I replaced was probably only used to test if some code path was being executed.

And that unpack method that I'm calling: that needs to be another one.

benoit-pierre:
Something like that:
```diff
 src/zc/buildout/easy_install.py | 10 ++++++++++
 1 file changed, 10 insertions(+)

diff --git i/src/zc/buildout/easy_install.py w/src/zc/buildout/easy_install.py
index aadf670..4f0592d 100644
--- i/src/zc/buildout/easy_install.py
+++ w/src/zc/buildout/easy_install.py
@@ -1639,6 +1639,16 @@ def unpack_wheel(location, dest):
     shutil.rmtree(tmp_dest)
     
 
+# Use setuptools>=38.2.0 support for wheels if available.
+try:
+    from setuptools.wheel import Wheel
+except ImportError:
+    pass
+else:
+    def unpack_wheel(location, dest):
+        wheel = Wheel(location)
+        wheel.install_as_egg(os.path.join(dest, wheel.egg_name()))
+
 UNPACKERS = {
     '.egg': unpack_egg,
     '.whl': unpack_wheel,
```

reinout:
Thanks, I'm trying that out right now. Will take some time as buildout's tests run quite slow :-)

src/zc/buildout/easy_install.py
```diff
@@ -37,6 +37,12 @@
 import zc.buildout
 import warnings
 
+try:
+    from setuptools.wheel import Wheel
+    SETUPTOOLS_SUPPORTS_WHEELS = True
+except ImportError:
+    SETUPTOOLS_SUPPORTS_WHEELS = False
```
benoit-pierre: Since v38.2.0 does not support namespace packages correctly, maybe it would be best to check the version of setuptools?
```python
from pkg_resources import SetuptoolsVersion
from setuptools import __version__ as setuptools_version
SETUPTOOLS_SUPPORTS_WHEELS = SetuptoolsVersion(setuptools_version) >= SetuptoolsVersion('38.2.3')
```

reinout:
@benoit-pierre: I've done it more-or-less like you suggested (https://github.com/buildout/buildout/pull/428/commits/4904f968629259ec0a4648c11868020b76a0725c). The tests are actually running on travis now, which means that zope.testrunner got installed correctly!

The weird demo-0.3.whl test code is actually part of an older attempt at supporting wheels, which I can remove now. So that'll be cleaned up :-)

src/zc/buildout/easy_install.py
```diff
@@ -1619,29 +1625,23 @@ def unpack_egg(location, dest):
     setuptools.archive_util.unpack_archive(location, dest)
 
 
-WHEEL_TO_EGG_WARNING = """
-Using unpack_wheel() shim over the deprecated wheel_to_egg() hook.
-Please update your wheel extension implementation for one that installs a .whl
-handler in %s.UNPACKERS
-""".strip() % (__name__,)
+WHEEL_WARNING = """
+*.whl file detected (%s), you'll need setuptools > 38.2.0 for that
```
benoit-pierre: So maybe suggest 38.2.3?

src/zc/buildout/tests.py
```diff
@@ -3704,7 +3704,6 @@ def test_suite():
                ])
             ),
         doctest.DocFileSuite('testing_bugfix.txt'),
-        unittest.makeSuite(UnitTests),
```
mgedmin: I recommend `unittest.defaultTestLoader.loadTestsFromName(__name__)` instead of manually calling `unittest.makeSuite()` for every TestCase subclass.

(Assuming there are any unittest.TestCase subclasses left.)

reinout: Nope :-) I just removed it :-)

reinout:
Yeah! The tests pass now!
Some questions/thoughts:

- That the tests run is proof that this buildout version cooperates well with the latest setuptools. Reason: zope.testrunner is installed as a wheel before running the tests.

- Testing if it works with an old setuptools is hard. 99.9% certain that it will work just fine as it doesn't download wheels so the new code won't be hit.

- I'm detecting setuptools support with a try/except import as I don't want to try comparing versions at import time. Is that OK? And we need the try/except anyway.

- There might have to be a test for this functionality (though it is tested by definition if you can actually run the tests).

- We *could* go all the way..... require setuptools 38.2.3 at a minimum.... Raise our own version number to 3.0 (it would have to be 2.10 otherwise, which I don't mind).

src/zc/buildout/easy_install.py
```diff
@@ -1619,25 +1622,19 @@ def unpack_egg(location, dest):
...
 
 
-WHEEL_TO_EGG_WARNING = """
-Using unpack_wheel() shim over the deprecated wheel_to_egg() hook.
-Please update your wheel extension implementation for one that installs a .whl
-handler in %s.UNPACKERS
-""".strip() % (__name__,)
+WHEEL_WARNING = """
+*.whl file detected (%s), you'll need setuptools > 38.2.3 for that
+or an extension like buildout.wheel
```
leorochael: Can you please make it?:

    or an extension like buildout.wheel > 0.2.0

src/zc/buildout/easy_install.py
```diff
@@ -1619,25 +1622,19 @@ def unpack_egg(location, dest):
...
-    wheel_to_egg(list(dists)[0], tmp_dest)
-    [egg] = glob.glob(os.path.join(tmp_dest, '*.egg'))
-    unpack_egg(egg, dest)
-    shutil.rmtree(tmp_dest)
-    
+    if SETUPTOOLS_SUPPORTS_WHEELS:
+        wheel = Wheel(location)
+        wheel.install_as_egg(os.path.join(dest, wheel.egg_name()))
+    else:
+        logger.warning(WHEEL_WARNING, location)
```
leorochael: > I don't like putting in the old shim as it was quite a horrid piece of code that only did anything meaningful if you had an old version of buildout.wheel. In that case you're probably also using an older buildout, I'd say (or you need to upgrade). So I'd rather keep the shim out.

You're right.

Still, this corner case in the else clause means something that is guaranteed to cause an error condition later on that is harder to track to the cause (user has old `setuptools`, with old `buildout.wheel`).

And `wheel_to_egg()` is still dead code, so would you mind turning this warning into an error and removing `wheel_to_egg()`?

src/zc/buildout/easy_install.py
reinout: Good idea. Done!

src/zc/buildout/easy_install.py
reinout: `wheel_to_egg()` is gone

src/zc/buildout/easy_install.py
mgedmin: Should these be `>` or `>=`?

benoit-pierre: Definitely `>= 38.2.3` for setuptools, good catch!

reinout: Whoopsie! Fixed in f76788c06b7652cca7213044eb0e142e0612fd57 :-)

src/zc/buildout/easy_install.py
reinout: And the warning is an error now.

src/zc/buildout/easy_install.py
reinout: I added your version check *inside* the try/except in 5a79e918c73830cc3994402cc683c21bee334731 .
I cannot do that on the top level as there are even checks there if setuptools is even installed.

reinout:
@leorochael, you're happy with the changes you requested, I assume?

VuongN:
For what it's worth, I am eagerly awaiting this change. Thank you @reinout for the great work and 'et al' for the support.

reinout:
OK to merge it? It *should* not hurt existing behaviour and it actually fixes lots of broken travis/jenkins jobs, I think :-)

leorochael:
> @leorochael, you're happy with the changes you requested, I assume?

Almost. I realize I wasn't explicit in what changes I requested.

What I meant was not changing `logger.warning(...)` into `logger.error(...)`.

Rather, I wanted to replace `logger.warning(...)` with `raise zc.buildout.UserError(...)`.

I see no point in allowing the execution to continue after this point when it's clearly going to break later on with a hard-to-debug error message.

reinout:
@leorochael: good idea, I've changed it in https://github.com/buildout/buildout/pull/428/commits/f3207f9e4438438fddc2246ebd286cc92f117a65

reinout:
I'll make a new release monday morning (morning EU time). That way if something goes really wrong I've got an entire day to fix it :-)

jimfulton:
Thank you @reinout!!!  And thank y'all for the reviews!

reinout:
I've released 2.10.0.
Lets hope it all goes OK!

mgedmin:
I'm having great success with most of my builds.

One project is still failing, specifically
```
Getting distribution for 'zc.buildout==2.10.0'.
While:
  Installing.
  Loading extensions.
  Getting distribution for 'zc.buildout==2.10.0'.
Error: Couldn't find a distribution for 'zc.buildout==2.10.0'.
```
This is a situation where an older zc.buildout (2.9.6, I think) is trying and failing to upgrade itself, after I changed the zc.buildout version pin in my versions.cfg.

Perhaps uploading a zc.buildout 2.10 wheel to PyPI wasn't the best idea?  I think an older zc.buildout could manage to upgrade itself if all it saw was an sdist.

reinout:
There have been .whl releases of buildout for quite some time now. Refraining from building buildout wheels is probably going to be hard.

Can you run the failing one with `-vvvvv` or so? 

I wonder if it would work in a two-stage approach. First upgrading buildout and then upgrading setuptools. There's an interaction between those versions now.

`pip install -U zc.buildout setuptools` and then just running `buildout` could help, too. I hope.

mgedmin:
`pip install -U zc.buildout setuptools` is what I did in the end, and it worked perfectly.

reinout:
I secretly hope that #432 also helps because it makes the paths more explicit/dependable. Sadly that'll only work for newer versions. Retroactively fixing old versions is too hard :-)

Can you perhaps take a quick look at #432? Releasing it might be important if we need to keep setuptools versions sorted out.

reinout:
I've released zc.recipe.egg 2.0.5 with the hopefully-better egg order handling.

PR #430, opened 2017-12-01
Remove no longer existing winbot

icemac:


reinout:
winbot.zope.org is permanently down?

reinout:
Note: the test errors are because of a recent setuptools change.

There's a patch in the works (#428), but that can't be checked before https://github.com/pypa/setuptools/issues/1214 is fixed.

agroszer:
winbot RIP... decomissioned

PR #431, opened 2017-12-01
Added sorting of working set used by zc.recipe.egg

Cykooz:
It could fix #429 issue.

PR #432, opened 2017-12-04
Added sorting of working set used by zc.recipe.egg

### LINKED ISSUE #429: Order of paths in scripts (eg. paster) is arbitrary
There seems to be no consistency to the paths generated by zc.recipe.egg between different machines.  

This is not generally a problem, except for /usr/lib/python2.7/dist-packages; when it arbitrarily appears further up in the list, there is a potential for conflicts with system packages being preferred and detected before buildout eggs. 

Manually editing the bin/paster script and moving the system path to the bottom ignores system packages in favor of buildout eggs.

It is confusing, since a buildout that works perfectly on one machine fails on another for no apparent reason.

Cykooz:
It could fix #429 issue.

PS: This PR is a fixed version of https://github.com/buildout/buildout/pull/431

### DIFF
--- a/zc.recipe.egg_/CHANGES.rst
+++ b/zc.recipe.egg_/CHANGES.rst
@@ -4,7 +4,8 @@ Change History
 2.0.5 (unreleased)
 ==================
 
-- Nothing changed yet.
+- Fix #429: added sorting of working set by priority of different
+  type of paths (develop-eggs-directory, eggs-directory, other paths).
 
 
 2.0.4 (2017-08-17)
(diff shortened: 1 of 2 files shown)


reinout:
Looks good to me. Could you add a quick `CHANGES.rst` entry?

Cykooz:
I've been added an entry into `CHANGES.rst`.

PS: sorry for my English

reinout:
Thanks for the fix! Released as zc.recipe.egg 2.0.5.
Let's hope it leads to better-running buildouts!

prsephton:
Thanks for the fix, guys, most appreciated.

PR #435, opened 2018-01-12
Fix race condition with working set not reflecting reality during integrity check.

### LINKED ISSUE #423: VersionConflict with globally installed packages
The following relates to buildout 2.9.5

If there is a package installed globally, eg. zope.interface 4.1.1  in /usr/lib/python2.7/dist-packages, and buildout tries to install a lower version, eg. zope.interface==3.7.0, then buildout fails with a VersionConflict.  Re-running buildout directly afterwards succeeds, as the 3.7.0 egg is installed before failure.

    File "/home/paul/git/grokproject/eggs/zc.buildout-2.9.5-py2.7.egg/zc/buildout/easy_install.py", line 714, in install
    File "/home/paul/git/grokproject/eggs/zc.buildout-2.9.5-py2.7.egg/zc/buildout/easy_install.py", line 570, in _get_dist
      dist = self._env.best_match(requirement, ws)
    File "build/bdist.linux-x86_64/egg/pkg_resources/__init__.py", line 1111, in best_match
    dist = working_set.find(req)
    File "build/bdist.linux-x86_64/egg/pkg_resources/__init__.py", line 715, in find
    raise VersionConflict(dist, req)

Virtualenv does not help in resolving this, even with the --no-site-packages.

Buildout no longer has a -s flag to ignore site-packages.

I would expect buildout to ignore the conflict and continue, in fact not report a conflict at all, as the potential conflict is directly resolved by the action of installing the egg.

prsephton:
In easy_install.Installer._get_dist, a package is retrieved and installed in the eggs folder (eg. ~/.buildout/eggs).

Previously, the program would re-scan the environment for changes, check that the newly downloaded package was accessible and return a list of packages (allowing for dependencies) which were installed.  These packages would then be added to the working set.

The problem was that the accessibility check was done against a working set that did not yet reflect the newly downloaded packages.  This was not generally a problem, except where a conflicting package was already installed (for example in sdist).  In such a case, buildout would fail, reporting a version conflict, but running buildout directly afterward would succeed because the environment and working set would contain the packages installed in the eggs folder, and there would be no conflict.

This patch moves the "ws.add(dist)" calls from where they previously were, to inside the _get_dist() method, and adds the "replace=True" flag which effectively replaces any conflicting working set packages with the newly downloaded ones.  Now when we do:

            self._env_rescan_dest()
            dist = self._env.best_match(requirement, ws)

we are testing against a working set that reflects the true file system state, and buildout runs consistently.

Fixes #423

### DIFF
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -568,6 +568,9 @@ class Installer:
                         "Couldn't download distribution %s." % avail)
 
                 dists = [_move_to_eggs_dir_and_compile(dist, self._dest)]
+                for _d in dists:
+                    if _d not in ws:
+                        ws.add(_d, replace=True)
 
             finally:
                 if tmp != self._download_cache:
@@ -575,10 +578,14 @@ class Installer:
 
             self._env_rescan_dest()
             dist = self._env.best_match(requirement, ws)
+
             logger.info("Got %s.", dist)
 
         else:
             dists = [dist]
+            if dist not in ws:
+                ws.add(dist)
+
 
         if not self._install_from_cache and self._use_dependency_links:
             self._add_dependency_links_from_dists(dists)
@@ -635,8 +642,7 @@ class Installer:
                 pkg_resources.Requirement.parse('setuptools')
                 )
             if ws.find(requirement) is None:
-                for dist in self._get_dist(requirement, ws):
-                    ws.add(dist)
+                self._get_dist(requirement, ws)
 
     def _constrain(self, requirement):
         """Return requirement with optional [versions] constraint added."""
@@ -670,7 +676,6 @@ class Installer:
 
         for requirement in requirements:
             for dist in self._get_dist(requirement, ws):
-                ws.add(dist)
                 self._maybe_add_setuptools(ws, dist)
 
         # OK, we have the requested distributions and they're in the working
@@ -719,7 +724,6 @@ class Installer:
                     logger.debug('Adding required %r', str(req))
                 self._log_requirement(ws, req)
                 for dist in self._get_dist(req, ws):
-                    ws.add(dist)
                     self._maybe_add_setuptools(ws, dist)
             if dist not in req:
                 # Oops, the "best" so far conflicts with a dependency.
--- a/src/zc/buildout/easy_install.txt
+++ b/src/zc/buildout/easy_install.txt
@@ -125,8 +125,8 @@ The working set contains the distributions we retrieved.
 
     >>> for dist in ws:
     ...     print_(dist)
-    demo 0.2
     demoneeded 1.1
+    demo 0.2
 
 We got demoneeded because it was a dependency of demo.
 
@@ -169,8 +169,8 @@ The old setting is returned.
     ...     ['demo'], dest, links=[link_server], index=link_server+'index/')
     >>> for dist in ws:
     ...     print_(dist)
-    demo 0.4rc1
     demoneeded 1.2rc1
+    demo 0.4rc1
 
     >>> ls(dest)
     d  demo-0.2-py2.4.egg
@@ -195,9 +195,9 @@ dependencies.  We might do this to specify a specific version.
 
     >>> for dist in ws:
     ...     print_(dist)
-    demo 0.3
-    other 1.0
     demoneeded 1.0
+    other 1.0
+    demo 0.3
 
     >>> ls(dest)
     d  demo-0.2-py2.4.egg
@@ -231,8 +231,8 @@ Let's check that the uppercase dist is installed.
 
     >>> for dist in ws:
     ...     print_(dist)
-    MIXEDCASE 0.5
     demoneeded 1.1
+    MIXEDCASE 0.5
     >>> ls(dest)
     d  MIXEDCASE-0.5-pyN.N.egg
     d  demoneeded-1.1-py2.4.egg
@@ -256,7 +256,7 @@ then the versions numbers will be used.
     ...     ['demo'], dest, links=[link_server], index=link_server+'index/',
     ...     versions = dict(demo='0.2', demoneeded='1.0'))
     >>> [d.version for d in ws]
-    ['0.2', '1.0']
+    ['1.0', '0.2']
 
 In this example, we specified a version for demoneeded, even though we
 didn't define a requirement for it.  The versions specified apply to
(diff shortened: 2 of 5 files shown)


reinout:
Your description of the problem is not totally clear to me. At least, I'm a bit confused by your terminology:

- An "sdist" is a source distribution, a .tgz you download from pypi. It is a package. So if you say "except where a conflicting package was already installed (for example in sdist)", it doesn't make much sense.

- A "working set" is an in-memory list of packages that buildout explicity adds packages to. So it isn't the case that everything in the `eggs/` directory is automatically added to it. So it isn't automatically in sync when you run buildout again. 

- File system state: what's in `eggs/` isn't automatically what's in the working set. You can set the eggs dir to `~/buildout-cache/eggs/`, for instance, sharing/caching eggs between many different buildouts.

Can you clarify a bit?

prsephton:
Hi, @reinout, 

> An "sdist" is a source distribution, a .tgz you download from pypi

Sorry, I misspoke.  What I meant to say, is "excepting where an existing package was installed, for example as an installed system package".  

In my case for example, my Debian distro has an unrelated app which depends on zope.interface, and this gets installed as a system package.  My buildout-installed zope.interface has a pinned version which pre-dates the one Debian installs.

For the sake of discussion, lets stick to zope.interface as an example.

> A "working set" is an in-memory list of packages that buildout explicity adds packages to.

This is true, but `pkg_resources.WorkingSet.__init__()` (line 608) initialises itself with:

        if entries is None:
            entries = sys.path

        for entry in entries:
            self.add_entry(entry)

WorkingSet's `__iter__()` comment reads as follows:

    def __iter__(self):
        """Yield distributions for non-duplicate projects in the working set

        The yield order is the order in which the items' path entries were
        added to the working set.
        """

`pkg_resources._initialize_master_working_set()` also explicitly adds all known packages to the working set on startup, one at a time.

Therefore,  WorkingSet already knows about the zope.interface (and all other packages) installed at a system level.

On the first run, buildout downloads our pinned version of zope.interface and installs it in the eggs dir.  At line 569 of existing code for `easy_install.Installer._get_dist()`, it then calls:

            self._env_rescan_dest()
            dist = self._env.best_match(requirement, ws)

which reports a version conflict, as the newly downloaded and installed zope.interface conflicts with the already known system package.  At this point, our new egg is not known to us in `Installer._env` or working set, although it has been installed in the eggs dir.  We currently add the egg to the working set only *after* `_get_dist()` returns.

> File system state: what's in eggs/ isn't automatically what's in the working set.

On the second run, when installing zope,interface, easy_install.install() creates an easy_install.Installer() with the eggs dir as dest, and `easy_install.__init__()` creates a `self._env` that contains all of the known eggs in the dest.  In `Installer._get_dist()` (line 532), `Installer._satisfied()` returns the dist from the eggs dir (self._env), and there is no version conflict.  When `_get_dist()` returns, the dist is added to the working set.  

So by the time all is done, the working set contains the *relevant* eggs from the eggs dir.

So in conclusion, there is inconsistent behavior for buildout between the first and subsequent runs if a pinned package version conflicts with an already installed system package.

To get buildout to succeed on the first try, after downloading and installing the package at easy_install.Installer._get_dist(), line 557, we need to first add the package to the working set, and *then* do `dist = self._env.best_match(requirement, ws)` which will no longer report a version conflict (unless there really is one).

I hope this helps.

reinout:
I don't totally understand the full implications of the change yet, but it *smells* like a good one. If I'm right, it might especially solve some setuptools-from-the-main-sitepackages issues. Which is a very good thing :-)

There are some errors in the tests. All of them seem related to the path order of setuptools. @prsephton, can you check whether those changes in the test output are what you're expecting? (And if they are: change the tests to reflect the new situation).

prsephton:
Apologies for taking a while over this update, but I have been doing, redoing, exploring and going over things with a fine toothcomb, and reducing the footprint of the change as far as possible.

@leorochael Thanks for your feedback. For now, just to avoid confusion I would like to make as few changes as possible.  If everyone is happy eventually, perhaps we can do the rename of _get_dist() to _get_dist_and_update_working_set() (or whatever).  Since _maybe_add_setuptools() calls _get_dist() it would probably not be a good idea to inline this function within _get_dist().

@reinout The reversal of package lists inside of working set all have to do with the use of the "replace=True" argument to ws.add().  By first checking for the existence of a package in the ws prior to calling ws.add(), most of those additions which would have affected the ordering are eliminated, and this minimised the number of tests which broke.  Indeed, the PR without the "replace=True" argument, all of the original tests succeed without any changes to the test cases at all.

As for the implications of this change, I think it addresses a very specific behaviour, where the working set may already contain a package which we are trying to install as an egg, and the new egg conflicts with the system package. Other than that edge case, I believe behaviour is identical, or near as dammit to existing implementation.

Regards

leorochael:
> @leorochael Thanks for your feedback. For now, just to avoid confusion I would like to make as few changes as possible.  If everyone is happy eventually, perhaps we can do the rename of _get_dist() to _get_dist_and_update_working_set() (or whatever).  Since _maybe_add_setuptools() calls _get_dist() it would probably not be a good idea to inline this function within _get_dist().

I'm fine with `_get_dist_and_update_working_set()`.

The idea of inlining `_maybe_add_setuptools()` would be to also change it to not call `_get_dist()` recursively but to detect whether adding setuptools is necessary as soon as possible so as to interpret the parameters as not only requesting the original dist, but setuptools as well.

That said, I'm fine leaving the inlining change for later.

> As for the implications of this change, I think it addresses a very specific behaviour, where the working set may already contain a package which we are trying to install as an egg, and the new egg conflicts with the system package. Other than that edge case, I believe behaviour is identical, or near as dammit to existing implementation.

Seems fine to me. I would only ask that you collapse all your commits into the first.

@jimfulton, any opinion?

reinout:
I'm getting the impression that this will really help with system packages that conflict with versions we've specified ourselves.
I must say I'm not 100% sure that there won't be any side effects. But... as there were only minimal changes in the tests, I assume we're safe here.
So... I'm merging it and releasing it. Then we'll know for sure :-)

reinout:
@prsephton, I've released 2.11.0: can you double-check if it fixes your issue?
Thanks for the pull request!

prsephton:
Thanks, @reinout , its working quite sweetly!  Thanks for the review and release.

PR #439, opened 2018-02-02
Clarify intent plus spelling/grammar fixes

nlsdfnbch:


doc/getting-started.rst
```diff
@@ -349,7 +349,7 @@ Buildout environment
 --------------------
 
 A Buildout environment includes the operating system and the Python
-installation it's run with. The more a buildout depends on its
+installation Buildout is executed with. The more a buildout depends on its
```
reinout: "executed" is perhaps a bit more technically correct, but "run" seems fine enough to me.

doc/getting-started.rst
```diff
@@ -405,7 +405,7 @@ run without this option.
 Pinned versions
 _______________
 
-You can also pin required versions in two ways.  You can specify them
+You can also pin required versions, and so so in two ways.  You can specify them
```
reinout: "and so so"? That doesn't look OK.

doc/getting-started.rst
```diff
@@ -486,9 +486,7 @@ versions in their own file:
    >>> write(src, 'buildout.cfg')
 
 Here, we've used the Buildout ``extends`` option to say that
-configurations should be read from the named file (or files) and that
-configuration in the current file should override configuration in the
-extended files.  To continue the example, our ``versions.cfg`` file
+configurations should be read from the named file (or files). These configurations will override and/or extend the current file from which these are loaded (in this case, ``buildout.cfg``).  To continue the example, our ``versions.cfg`` file
```
reinout: a) Please add linebreaks (to keep within 80 characters per line).

b) The "loading" config file wins over the config files it extends: I believe your sentence says it the wrong way around.

doc/getting-started.rst
```diff
@@ -539,7 +537,7 @@ With ``update-versions-file``, whenever Buildout gets the newest
 version for a requirement (subject to requirement constraints), it
 appends the version to the named file, along with a comment saying
 when and why the requirement is installed.  If you later want to
-upgrade a dependency, just edit this file with the new version, or to
+upgrade a dependency, manually update this file with the new version. Alternatively,
```
reinout: "just edit" is fine. "Manually update" suggests that your can also do it automatically.

doc/getting-started.rst
nlsdfnbch: >that configuration in the current file should override configuration in the extended files

This states that the included files are overwritten by the file that includes them

>These configurations will override and/or extend the current file from which these are loaded (in this case, ``buildout.cfg``)

States that the included files will overwrite the settings in the file that includes them (so the other way around as above)

So, which one is it ? :D 

I assumed it simply works the way other software uses `include`. And from what I understand, the following holds true for `Buildout`:

```
>>conf_a.conf
    setting_a = True
    include = ./conf_b.conf
>>conf_b.conf
    setting_a = False

>>Resulting config
    setting_a = False
```

Which is what my edit describes. Is this correct? If it isn't, your original sentence is correct.

doc/getting-started.rst
nlsdfnbch: Indeed, this should be ".., and do so in two ways." !

doc/getting-started.rst
nlsdfnbch: Okidoks!

doc/getting-started.rst
reinout: "extends" means that buildout reads that config as a base config. Everything in the current buildout config adds/overrides/changes the extended config.

If you have a base.cfg:

    ....
    [django]
    settings = myproject.developmentsettings

You could have a production.cfg like this:

    [buildout]
    extends = base.cfg

    ....
    [django]
    settings = myproject.productionsettings

So in production.cfg, you'd use myproject.productionsettings.

nlsdfnbch: Ah, alright! Thanks for clearing that up. I misunderstood this, then.

reinout:
Thanks!

PR #441, opened 2018-03-01
Correctly detect upgrade with extension that depends on setuptools

reinout:
Use case: buildout that uses mr.developer. When you bootstrap it with ye olde bootstrap.py, it starts up with setuptools 33.1.1. 

When installing the mr.developer extension, it upgrades setuptools to 38.5.2 (as we've pinned that version). With the new buildout 2.11 functionality, installed packages are immediately added to the working set.

After installing the extension, the `maybe_upgrade()` function it called. This checked the working set if it was up to date with the currently installed packages. With buildout 2.11, this was always the case.

I've changed the check to look at the buildout/setuptools install locations that were detected when first loading the `easy_install.py` module.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,9 @@ Change History
 2.11.1 (unreleased)
 ===================
 
-- Nothing changed yet.
+- Made upgrade check more robust. When using extensions, the improvement
+  introduced in 2.11 could prevent buildout from restarting itself when it
+  upgraded setuptools.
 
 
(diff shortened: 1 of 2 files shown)


reinout:
Seems to work like a charm.
OK to merge it? I'll release it afterwards.

byrman:
Yes, this fixes my problem, thanks! Looking forward to the new release.

reinout:
@byrman, can you check if 2.11.1 works for you?

Issue #442, opened 2018-03-19
attribute error looking for pkg_resources.SetuptoolsVersion


arterrey:
`pkg_resources.SetuptoolsVersion` is now gone (see https://github.com/pypa/setuptools/commit/eeeb9b27fa48fccf2b5d52919eff1c75c4ad1718) and has been renamed  to `SetuptoolsLegacyVersion`

Using setuptools 39.0.1 and zc.buildout 2.11.1
buildout now breaks with an attribute error...

```
bin/buildout
Traceback (most recent call last):
  File "bin/buildout", line 11, in <module>
    load_entry_point('zc.buildout==2.11.1', 'console_scripts', 'buildout')()
  File "/a/lib/python3.6/site-packages/pkg_resources/__init__.py", line 480, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/a/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2693, in load_entry_point
    return ep.load()
  File "/a/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2324, in load
    return self.resolve()
  File "/a/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2330, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/a4/lib/python3.6/site-packages/zc.buildout-2.11.1-py3.6.egg/zc/buildout/buildout.py", line 18, in <module>
    import zc.buildout.easy_install
  File "/a/lib/python3.6/site-packages/zc.buildout-2.11.1-py3.6.egg/zc/buildout/easy_install.py", line 45, in <module>
    pkg_resources.SetuptoolsVersion(setuptools_version) >=
AttributeError: module 'pkg_resources' has no attribute 'SetuptoolsVersion'
```

enkidulan:
same here, pinnig buildout version to `2.9.6` fix the problem. Issue appers starting with version `2.10.0`

gforcada:
@enkidulan I think that this is rather a problem with newer setuptools. On version 39.0.0 ``SetuptoolsVersion`` got removed https://setuptools.readthedocs.io/en/latest/history.html#v39-0-0

enkidulan:
@gforcada in my case buildout version `2.9.6` works with the newest `setuptools` version fine. It seems like a combination :) Yep, here are new changes introduced in `2.10.0` that depend on `SetuptoolsVersion` -
[...]

PR #443, opened 2018-03-19
Fix for #442 issue

enkidulan:
`pkg_resources.SetuptoolsVersion` got deprecated in `setuptools` `v39.0.0`.
Moved on using stable and recommended `pkg_resources.parse_version`
function that does the same. For more details on the topic
see http://setuptools.readthedocs.io/en/latest/history.html#v39-0-0.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Change History
 2.11.2 (unreleased)
 ===================
 
-- Nothing changed yet.
+- Fix #442 issue. [enkidulan]
 
 
 2.11.1 (2018-03-01)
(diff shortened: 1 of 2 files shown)


reinout:
Thanks for the fix! I've released 2.11.2, can you double-check that it works?

enkidulan:
It seems to work fine, I just have tested the new buildout release against my current project.

reinout:
Nice. Thanks for the work, again.

jimfulton:
Seriously. Thanks very much!

enkidulan:
You are very welcome :) It's the least I can do to show my appreciation for `buildout` :)

PR #445, opened 2018-04-12
Use the new PyPI at https://pypi.org/.

icemac:
According to a blog post of [Sumana Harihareswara](https://www.harihareswara.net/sumana/2018/04/11/0) https://pypi.python.org/ will be a redirect to https://pypi.org by April 16th.

I changed all occurrences (besides the ones in `HISTORY.rst`) of `pypi.python` to the new URLs.

### DIFF
--- a/src/zc/buildout/easy_install.txt
+++ b/src/zc/buildout/easy_install.txt
@@ -46,7 +46,7 @@ index
    The URL of an index server, or almost any other valid URL. :)
 
    If not specified, the Python Package Index,
-   https://pypi.python.org/simple/, is used.  You can specify an
+   https://pypi.org/simple/, is used.  You can specify an
    alternate index with this option.  If you use the links option and
    if the links point to the needed distributions, then the index can
    be anything and will be largely ignored.  In the examples, here,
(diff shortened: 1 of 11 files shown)


src/zc/buildout/easy_install.txt
```diff
@@ -46,7 +46,7 @@ index
    The URL of an index server, or almost any other valid URL. :)
 
    If not specified, the Python Package Index,
-   https://pypi.python.org/simple/, is used.  You can specify an
+   https://pypi.org/simple/, is used.  You can specify an
```
mgedmin: I see you're changing the default index URL in the documentation, but I don't see you changing it in the code itself.

(Is the default index URL hardcoded in zc.buildout itself, or does it come from setuptools?)

mauritsvanrees: @mgedmin He changed the `default_index_url` in `easy_install.py`.

icemac: It's here: https://github.com/buildout/buildout/pull/445/files#diff-784f9bb654bbcee96dcbc547bbddd493

mgedmin: Whoops!

reinout:
Thanks, I've released it as 2.11.3.

PR #458, opened 2018-06-29
Add 'allow-unknown-extras'

### LINKED ISSUE #457: UnknownExtra: html5lib 0.9999999 has no such extra feature 'datrie'
Hello,

I am getting the following error while installing eggs

UnknownExtra: html5lib 0.9999999 has no such extra feature 'datrie'

I am using two libraries nti.contentfragments which has in its setpy.py/install_requires
[
'setuptools',
'html5lib[datrie]', # > 1.0.1
...

and google tensorflow which has a hard requirement for 
html5lib = 0.9999999

I am pinning html5lib to 0.9999999 in my version.cfg; however; instead of getting a warning 
(i.e. html5lib 10.9999999 does not provide the extra 'datrie') like when installing a package in pip, I get the exception mentioned above.

Thanks

Carlos

jamadden:
Fixes #457

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,20 +4,23 @@ Change History
 2.11.6 (unreleased)
 ===================
 
-- Nothing changed yet.
+- Add a new buildout option ``allow-unknown-extras`` to enable
+  installing requirements that specify extras that do not exist. This
+  needs a corresponding update to zc.recipe.egg. See `issue 457
+  <https://github.com/buildout/buildout/issues/457>`_.
 
 
 2.11.5 (2018-06-19)
 ===================
 
-- Fix for `issue 295 <https://github.com/buildout/buildout/issues/295>`. On
+- Fix for `issue 295 <https://github.com/buildout/buildout/issues/295>`_. On
   windows, deletion of temporary egg files is more robust now.
 
 
 2.11.4 (2018-05-14)
 ===================
 
-- Fix for `issue 451 <https://github.com/buildout/buildout/issues/451>`:
+- Fix for `issue 451 <https://github.com/buildout/buildout/issues/451>`_:
   distributions with a version number that normalizes to a shorter version
   number (3.3.0 to 3.3, for instance) can be installed now.
 
--- a/src/zc/buildout/allow-unknown-extras.txt
+++ b/src/zc/buildout/allow-unknown-extras.txt
@@ -40,7 +40,7 @@ Now we can run the buildout and see that it fails:
     ...
     While:
       Installing eggs.
-    Error: Couldn't find the required extras. Add 'allow-unknown-extras=true' to the [buildout] configuration if this is acceptable.
+    Error: Couldn't find the required extra...
 
 If we flip the option on, the buildout succeeds
 
@@ -57,7 +57,7 @@ If we flip the option on, the buildout succeeds
     ... ''')
 
 
-Now we can run the buildout and make sure all attempts to dist.plone.org fails::
+Now we can run the buildout and only get a warning::
 
     >>> print_(system(buildout), end='') # doctest: +ELLIPSIS
     Develop: '/sample-buildout/allowdemo'
--- a/src/zc/buildout/allow-unknown-extras.txt
+++ b/src/zc/buildout/allow-unknown-extras.txt
@@ -1,0 +1,65 @@
+======================
+ Allow Unknown Extras
+======================
+
+Sometimes we need to allow unknown extras.
+
+The ``allow-unknown-extras`` option lets us do that in a buildout
+configuration, just as we can directly calling ``easy_install``
+works exactly like the one provided in ``easy_install``.
+
+Let's create a develop egg that requires a bogus extra.
+
+    >>> mkdir(sample_buildout, 'allowdemo')
+    >>> write(sample_buildout, 'allowdemo', 'dependencydemo.py',
+    ...       'import eggrecipekss.core')
+    >>> write(sample_buildout, 'allowdemo', 'setup.py',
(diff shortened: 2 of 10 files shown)


jamadden:
It took a few iterations, but CI is now green.

CHANGES.rst
```diff
@@ -4,20 +4,23 @@ Change History
...
+  installing requirements that specify extras that do not exist. This
+  needs a corresponding update to zc.recipe.egg. See `issue 457
+  <https://github.com/buildout/buildout/issues/457>`_.
 
 
 2.11.5 (2018-06-19)
 ===================
 
-- Fix for `issue 295 <https://github.com/buildout/buildout/issues/295>`. On
+- Fix for `issue 295 <https://github.com/buildout/buildout/issues/295>`_. On
```
reinout: Oopsie... that missing underscore is my error. Thanks for fixing it :-)

src/zc/buildout/easy_install.py
```diff
@@ -733,7 +736,20 @@ def install(self, specs, working_set=None):
...
 
             best[req.key] = dist
-            extra_requirements = dist.requires(req.extras)[::-1]
+            if self._allow_unknown_extras:
+                missing_requested = sorted(
+                    set(req.extras) - set(dist.extras)
+                )
+                for missing in missing_requested:
+                    logger.warning(
+                        '%s does not provide the extra \'%s\'',
```
reinout: It is slightly more readable if you quote it with double quotes, then you can omit the quote escapes.

src/zc/buildout/easy_install.py
```diff
@@ -733,7 +736,20 @@ def install(self, specs, working_set=None):
                     pkg_resources.VersionConflict(dist, req), ws)
 
             best[req.key] = dist
-            extra_requirements = dist.requires(req.extras)[::-1]
+            if self._allow_unknown_extras:
```
reinout: I'd add a comment on the line below like this:

    # Note: by default, self._allow_unknown_extras is False

Reason: the "if" part is normally the normal path, but in this case it is the "else" part that's almost always executed.

reinout: Another idea, which just means shuffling some code around: there is now a good warning, but the error is the same.

You could use the `missing_requested` list in both cases.

    if missing_requested:
        if self._allow_unkown_extras:
            logger.warn(...)
        else:
            logger.error(..., you can use --allow-unkown-extras if that is not a problem)
            sys.exit(1)   # or however that happens in the rest of the code

src/zc/buildout/easy_install.py
jamadden: I admit I lifted that straight out of pip, quotes and all :)

CHANGES.rst
```diff
@@ -4,20 +4,23 @@ Change History
 2.11.6 (unreleased)
 ===================
 
-- Nothing changed yet.
+- Add a new buildout option ``allow-unknown-extras`` to enable
+  installing requirements that specify extras that do not exist. This
+  needs a corresponding update to zc.recipe.egg. See `issue 457
```
reinout: I've missed what needs updating in zc.recipe.egg.
If that's the case, there should be a change in the requirements in buildout's setup.py (after the zc.recipe.egg release).

jamadden: Yikes! Those changes didn't make it into the commit! They wound up in the installed copy.

I didn't add any tests specifically for zc.recipe.egg because it was just trivial copying arguments around.

reinout: As long as the original tests didn't break... :-)

Note: if you needed to add an argument in zc.recipe.egg, I really want @jimfulton to have a quick look, as zc.recipe.egg rarely need changing.

jamadden: > As long as the original tests didn't break... :-)

They didn't, but the fact that I missed committing that file and everything still passed proves to me that it needed an integration test too. Added. :)

reinout:
LGTM.
@jimfulton, do you have any comments? Otherwise I'll release it.

(Note to self: add a requirement on the newer zc.recipe.egg release in zc.buildout's setup.py)

src/zc/buildout/allow-unknown-extras.txt
```diff
@@ -0,0 +1,65 @@
+======================
```
jimfulton: This is fine, however, the experiment to reuse doctests as documentation is over.   A regular unit test would have been just fine. :)

src/zc/buildout/easy_install.py
```diff
@@ -733,7 +736,29 @@ def install(self, specs, working_set=None):
...
+                logger.warning(
+                    '%s does not provide the extra \'%s\'',
+                    dist, missing
+                )
+
+            if missing_requested:
+                if not self._allow_unknown_extras:
+                    raise zc.buildout.UserError(
+                        "Couldn't find the required extras. Add 'allow-unknown-extras=true' "
+                        "to the [buildout] configuration if this is acceptable."
```
jimfulton: I'll prefer some different wording.

"Couldn't find the required extra.  This means the requirement is incorrect.  If the requirement is itself from software you requested, then there might be a bug in requested software.  You can ignore this by using  'allow-unknown-extras=true', however that may simply cause needed software to be omitted."

jimfulton: After all, we're not doing this to be mean and make people say the magic word.

reinout:
Ok, yes, there's a little bit more work to do. Hopefully #461 sorts it all out.

reinout:
With zc.recipe.egg 2.0.7 and zc.buildout 2.12.1 the dependency between the both of them ought to be explicitly set.

Issue #460, opened 2018-07-02
KeyError: 'allow-unknown-extras'


mrichar1:
Since the latest release I'm now getting the following error trying to run buildout:

```
Getting distribution for 'zc.recipe.egg>=2.0.0a3'.
Got zc.recipe.egg 2.0.6.
Installing 3rdparty.
While:
  Installing 3rdparty.

An internal error occurred due to a bug in either zc.buildout or in a
recipe being used:
Traceback (most recent call last):
  File "/home/mrichar1/git/ros/eggs/zc.buildout-2.11.5-py3.6.egg/zc/buildout/buildout.py", line 2127, in main
    getattr(buildout, command)(args)
  File "/home/mrichar1/git/ros/eggs/zc.buildout-2.11.5-py3.6.egg/zc/buildout/buildout.py", line 797, in install
    installed_files = self[part]._call(recipe.install)
  File "/home/mrichar1/git/ros/eggs/zc.buildout-2.11.5-py3.6.egg/zc/buildout/buildout.py", line 1557, in _call
    return f()
  File "/home/mrichar1/git/ros/eggs/zc.recipe.egg-2.0.6-py3.6.egg/zc/recipe/egg/egg.py", line 227, in install
    reqs, ws = self.working_set()
  File "/home/mrichar1/git/ros/eggs/zc.recipe.egg-2.0.6-py3.6.egg/zc/recipe/egg/egg.py", line 87, in working_set
    allow_unknown_extras=bool_option(buildout_section, 'allow-unknown-extras')
  File "/home/mrichar1/git/ros/eggs/zc.buildout-2.11.5-py3.6.egg/zc/buildout/buildout.py", line 2167, in bool_option
    raise KeyError(name)
KeyError: 'allow-unknown-extras'
```
Assume this is linked to #458 ?

reinout:
I first released zc.recipe.egg. Then I got distracted for half an hour and released zc.buildout (2.12).

So if you try it now, it should be OK.

reinout:
zc.buildout 2.12.0, that is.

Looks like I should update the minimum requirement on zc.buildout in zc.recipe.egg. That's at `1.2.0` at the moment.

mrichar1:
Yep - that's it working again with 2.0.6/2.12.0 - thanks!

reinout:
And with zc.recipe.egg 2.0.7 and zc.buildout 2.12.1 the problem should be fixed even better as they now each explicitly require the correct minimal version of the other.

PR #461, opened 2018-07-02
Fix dependencies between zc.buildout and zc.recipe.egg after zc.recipe.egg 2.0.6 release

reinout:


reinout:
This ought to be a fix for #460

reinout:
... and I also made buildout depend explicitly on 2.0.6+

jamadden:
Thanks for dealing with the bookkeeping on this!

reinout:
So, with zc.recipe.egg 2.0.7 and zc.buildout 2.12.1 it *really* should be fixed :-)

PR #465, opened 2018-08-29
debugger: Exit with error code on debugged error

jerome-nexedi:
When buildout encounter an error, program exists with error code 1, this
should also be the case when exiting from a debugger session when
running with -D option.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Change History
 2.12.2 (unreleased)
 ===================
 
-- Nothing changed yet.
+- Exit program with error code on error when running with ``-D``.
 
 
 2.12.1 (2018-07-02)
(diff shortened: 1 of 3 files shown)


jerome-nexedi:
Hello, `buildout -D` does not exits with error code when there's an error in the profile, I believe this is a bug.

jerome-nexedi:
Thanks @leorochael , thanks @reinout 

I pushed a new commit with a changelog entry.

reinout:
I've merged it, but I can't make a release right now as the guest wifi network I'm on doesn't allow ssh connections.

If I haven't made a release within 24 hours, I've forgotten about it. Please @mention me if so and I'll make the release :-)

jerome-nexedi:
Thanks ! I don't need a release, as far as I am concerned, there is no rush and this can go in next release if you prefer.

reinout:
I've released it as 2.12.2.
Thanks for the fix, again.

PR #471, opened 2019-01-15
add a `query` command

gotcha:
for continuous integration

I am pretty happy with the code. I have some doubt about what to do in case of multiline values.

Feedback appreciated.

### DIFF
--- a/src/zc/buildout/buildout.py
+++ b/src/zc/buildout/buildout.py
@@ -50,6 +50,19 @@ if PY3:
 else:
     text_type = unicode
 
+
+def command(method):
+    method.buildout_command = True
+    return method
+
+
+def commands(cls):
+    for name, method in cls.__dict__.items():
+        if hasattr(method, "buildout_command"):
+            cls.COMMANDS.add(name)
+    return cls
+
+
 def _print_options(sep=' ', end='\n', file=None):
     return sep, end, file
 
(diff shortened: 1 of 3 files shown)


gotcha:
@mgedmin Thanks for the review... and reminding me the changelog.

gotcha:
@reinout Thanks for the review.

src/zc/buildout/buildout.py
```diff
@@ -2115,7 +2140,7 @@ def main(args=None):
         command = args.pop(0)
         if command not in (
             'install', 'bootstrap', 'runsetup', 'setup', 'init',
-            'annotate',
+            'annotate', 'query',
```
leorochael: I wish there was a way not to have to repeat command names here.

Perhaps a simple `@command` decorator that appends method names to a list of commands...

Anyway, just a thought, not an actual request for change.

gotcha: Nice idea !

gotcha:
@leorochael Thanks for the review.

gotcha:
@leorochael Would you review the last commit ? following your suggestion...

leorochael:
> @leorochael Would you review the last commit ? following your suggestion...

It's perfect!

arterrey:
Can I suggest returning a non-zero status after handling the KeyError?

gotcha:
@arterrey Good point !

gotcha:
released at https://pypi.org/project/zc.buildout/2.13.0/

PR #473, opened 2019-02-07
Fix typo.

jugmac00:


src/zc/buildout/testing_bugfix.txt
```diff
@@ -9,7 +9,7 @@ The buildout testing set up runs a buildout which adds a
 remove it. This can disturb other tests of packages reusing
 zc.buildout.testing.
 
-The handers before calling set up are:
+The handlers before calling set up are:
```
reinout:
```suggestion
The handlers before calling our own testing set-up are:
```

This would probably take care of the rest of the sentence?

reinout: Thanks for the typo fix, btw :-)

jugmac00: A few lines above "setup" (as I would probably write it as a non native :-) ) is written as "set up" - see https://github.com/buildout/buildout/blob/d6c13b3edfb61b46a2a129104c3c23f282700c86/src/zc/buildout/testing_bugfix.txt#L7 - maybe fix both? or keep  "set up"?

Btw... I just came here by coincidence, when my ill configured py.test collected not only my apps' tests, but also these 10 year old tests - which then failed (got 2 logger instead of 1) - but this is yet another story.

mgedmin: English is a funny language:

- when you need a verb, you have to use "set up"
- when you need a noun, you have to use "setup"
- when you need an adjective, you have to use "set-up"

Examples:

- be ready to set up the thing
- make sure your setup was done correctly
- the set-up routine is long and dangerous

(But now that I look at it, the last example feels wrong!  I am not a native English speaker.)

mgedmin: Ah, here it looks like it was used as a noun in both places, so you're absolutely right, it should be "setup".

jamadden: As a native speaker, I agree that that last example looks funny, whether or not it is correct. I think 'set-up' is rarely used. I would probably just write "the setup routine is long and dangerous"  (but I admit to some past sloppiness around "setup" and "set up").

jugmac00:
And that all just because of a minor typo :-)

Btw - I only have write access to the zopefoundation repositories - so somebody else has to merge this pull request.

icemac:
This PR was approved by two persons. Who is able to merge it?

mgedmin:
Ah sorry I keep forgetting that buildout is not part of the zopefoundation org and most people can't merge their own PRs.

PR #491, opened 2019-11-14
Fix simple typo: expecially -> especially

### LINKED ISSUE #490: Small typo in src/zc/buildout/testing.py
Should read especially rather than expecially.

timgates42:
Closes #490


reinout:
The tests fail, but I'm merging anyway.
No way a typo change in a comment could cause whatever is going wrong in the test :-)

Issue #493, opened 2019-11-25
setuptools is removing easy_install, deprecation warning in 42.0.0


k4ml:
Refer to pypa/setuptools/pull/1830.

Buildout still calling out to setuptools.command.easy_install in `zc/buildout/easy_install.py:call_easy_install(spec, dest)`.

So what our take and future direction on this?

42.0.0 changelog also mentioned about eggs are not supported anymore.

> #1830, #1909: Mark the easy_install script and setuptools command as deprecated, and use pip when available to fetch/build wheels for missing setup_requires/tests_require requirements, with the following differences in behavior:
> 
>         support for python_requires
>         better support for wheels (proper handling of priority with respect to PEP 425 tags)
>         PEP 517/518 support
>         eggs are not supported

mauritsvanrees:
This also worries me. For now it seems to still work though.
I tried with Python 3.7.4 and this `requirements.txt`:

```
setuptools==42.0.1
zc.buildout==2.13.2
```

and this `buildout.cfg`:

```
[buildout]
parts = test
eggs-directory = eggs
download-cache = downloads
abi-tag-eggs = false

[test]
recipe = zc.recipe.egg
eggs = zest.releaser
```

Result is:

```
$ bin/buildout 
Creating directory '/Users/maurits/tmp/s42/downloads'.
Creating directory '/Users/maurits/tmp/s42/eggs'.
Creating directory '/Users/maurits/tmp/s42/parts'.
Creating directory '/Users/maurits/tmp/s42/develop-eggs'.
Getting distribution for 'zc.recipe.egg>=2.0.6'.
WARNING: The easy_install command is deprecated and will be removed in a future version.
Got zc.recipe.egg 2.0.7.
Installing test.
Getting distribution for 'zest.releaser'.
WARNING: The easy_install command is deprecated and will be removed in a future version.
warning: no previously-included files found matching '.installed.cfg'
warning: no previously-included files found matching '.gitattributes'
warning: no previously-included files matching '*.pyc' found anywhere in distribution
Got zest.releaser 6.19.1.
Getting distribution for 'twine>=1.6.0'.
Got twine 3.1.0.
Getting distribution for 'six'.
Got six 1.13.0.
...
Generated script '/Users/maurits/tmp/s42/bin/release'.
```

The generated scripts seem to work.
The `eggs` directory contains 23 eggs.
The `downloads` directory contains 21 wheels and two tar balls.

[...]

PR #494, opened 2019-11-26
Use Github Actions for smoke tests

gotcha:
Travis builds are broken. So we are not warned, among others, about changes in setuptools.

`setuptools 42.0.1` breaks our tests. See #493 

I also made smoke tests.

### DIFF
--- a/.github/workflows/scripts.yml
+++ b/.github/workflows/scripts.yml
@@ -1,0 +1,33 @@
+name: Test generating scripts
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 4
+      matrix:
+        python-version: [2.7, 3.5, 3.6, 3.7]
+        package: [zest.releaser, pyspf]
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Setup buildout virtualenv
+      run: |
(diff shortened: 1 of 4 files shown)


gotcha:
Building Python 3.4 is broken on GHA.
They provide support for Python 3.4 to 3.8. Do we really want to build our own python ?
Or was it because of previous limitations of Travis ?

mauritsvanrees:
I think we build our own version of Python so we have a clean Python without setuptools.

gotcha:
> I think we build our own version of Python so we have a clean Python without setuptools.

Someone with some expertise might then want to check why Python3.4 builds withouts https support in `urllib/urlopen`.

reinout:
@mauritsvanrees is right: we use a setuptools-less plain python build so that we can test installing setuptools. There's quite some hairy code in there...

Python 3.4: we could also perhaps drop (explicit) support for it?

gotcha:
@reinout Good idea, just did so #495

reinout:
Note: I can't figure out where/why travis is hanging. 

I've got the same problem locally (which is theoretically good news). But even with "strace bin/test -vvvv" I can't readily figure it out.

.github/workflows/scripts.yml
```diff
@@ -0,0 +1,33 @@
...
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 4
+      matrix:
+        python-version: [2.7, 3.5, 3.6, 3.7]
```
mgedmin: No 3.8?

gotcha: Would make sense ! I did it in #495

PR #495, opened 2019-11-28
Tests on gha

gotcha:
Run tests with Github Actions

### DIFF
--- a/.github/workflows/pythons.yml
+++ b/.github/workflows/pythons.yml
@@ -17,11 +17,16 @@ jobs:
 
     steps:
     - uses: actions/checkout@v1
+    - name: Python minor version
+      env:
+        PYTHON_VER: ${{matrix.python-version}}
+      run: |
+        make python_version
     - name: Cache pythons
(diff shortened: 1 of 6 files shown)


gotcha:
@mauritsvanrees I did setup python cache properly. And stopped to run 3.4, I think supporting 4 major versions is enough.

mgedmin:
Where can one see the test output?  It's not here: https://github.com/buildout/buildout/pull/495/checks

gotcha:
@mgedmin I am new to GHA. I see the same as you when following the link above. Did you check https://github.com/buildout/buildout/actions ?

gotcha:
@gedmin It seems that the actions UI does not allow to point and click to some of the previous runs.

Nevertheless, I found that, via the following URL,

https://github.com/gotcha/buildout/commit/914f8054a5826d2bd0261e0be50ba4535e0863a6/checks

I can access previous runs by specifying both the proper fork and the proper git hash.

gotcha:
Included in #502

PR #502, opened 2020-04-29
WIP Use `pip install` in place of `setuptools easy_install`

gotcha:
This PR includes 4 features/fixes:

- replace deprecated `setuptools.easy_install` by `pip install`
- make tests run again on Travis and Github Actions.
- add PEP345 `Requires-Python` support.
- resurrect `buildout` (and `setuptools` and `pip`) auto-update feature.

It might look a lot but I think `buildout` needed some deeper work.

I would appreciate feedback that I would integrate asap as I already did with very early @mgedmin review.
I think there would be a need for an alpha release that could be tested by some members of the `buildout` community.

Please do not merge this PR !

I will rather go through a series of smaller PRs.


buildout.cfg
```diff
@@ -23,10 +28,10 @@ eggs =
...
   zc.recipe.egg
   coverage
-initialization =
-  import os
-  if 'COVERAGE_PROCESS_START' not in os.environ: os.environ['COVERAGE_PROCESS_START'] = os.path.abspath('../../.coveragerc')
-  if os.getenv('COVERAGE_PROCESS_START'): import coverage; coverage.process_startup()
+#initialization =
+#  import os
+#  if ('TOX_ENV_NAME' in os.environ) and ('COVERAGE_PROCESS_START' not in os.environ): os.environ['COVERAGE_PROCESS_START'] = os.path.abspath('../../.coveragerc')
+#  if os.getenv('COVERAGE_PROCESS_START'): import coverage; coverage.process_startup()
```
mgedmin, line 34: Remove entirely?

dev.py
```diff
@@ -16,63 +16,115 @@
...
+    print('Install pip')
+    print('')
+    try:
+        from urllib.request import urlopen
+    except ImportError:
+        from urllib2 import urlopen
 
-import pkg_resources, setuptools
-setuptools_path = os.path.dirname(os.path.dirname(setuptools.__file__))
+    tmp = tempfile.mkdtemp()
```
mgedmin, line 82:
```suggestion
    tmp = tempfile.mkdtemp(prefix='buildout-dev-')
```

dev.py
```diff
@@ -16,63 +16,115 @@
...
+    try:
+        import pip
+        print('Remove pip, setuptools, wheel')
+        print('')
+        if subprocess.call(
+            [sys.executable] + ['-m', 'pip', 'uninstall', '-y', 'setuptools', 'pip',  'wheel']
+        ):
+            raise SystemError(
+                "Buildout development with pre-installed pip and setuptools"
+                "Could not uninstall with pip"
```
mgedmin, line 54: These two literals get concatenated without any whitespace, resulting in "... and setuptoolsCould not uninstall ...".


```suggestion
                "Buildout development with pre-installed pip and setuptools\n"
                "Could not uninstall with pip"
```

src/zc/buildout/__init__.py
```diff
@@ -13,9 +13,108 @@
...
+from setuptools.package_index import URL_SCHEME
+from setuptools.package_index import HREF
+from setuptools.package_index import distros_for_url
+from setuptools.package_index import htmldecode
+
+from pip._internal.index.collector import HTMLPage
+from pip._internal.index.collector import parse_links
+from pip._internal.index.package_finder import _check_link_requires_python
+from pip._internal.models.target_python import TargetPython
+from pip._vendor import six
```
mgedmin, line 27: Oh no.  Oh no no no.  pip does not want anyone to be relying on pip's internals.  This is going to be an unmaintainable nightmare.

src/zc/buildout/__init__.py
```diff
@@ -13,9 +13,108 @@
 ##############################################################################
 """Buildout package
 """
+from setuptools.extern.six.moves import urllib
+from setuptools.package_index import PackageIndex
+from setuptools.package_index import URL_SCHEME
+from setuptools.package_index import HREF
+from setuptools.package_index import distros_for_url
+from setuptools.package_index import htmldecode
```
mgedmin, line 21: We have a requirement of setuptools >= 8.0.

setuptools 8.0 does not have setuptools.extern.

src/zc/buildout/__init__.py
```diff
@@ -13,9 +13,108 @@
...
+        return
+    else:
+        dists = list(distros_for_url(url))
+        if dists:
+            if not self.url_ok(url):
+                return
+            self.debug("Found link: %s", url)
+
+    if dists or not retrieve or url in self.fetched_urls:
+        list(map(self.add, dists))
```
mgedmin, line 52:
```suggestion
        for dist in dists:
            self.add(dist)
```

src/zc/buildout/__init__.py
```diff
@@ -13,9 +13,108 @@
...
+        # In Python 3 and got bytes but want str.
+        page = page.decode(charset, "ignore")
+    f.close()
+
+    links = []
+    for match in HREF.finditer(page):
+        link = urllib.parse.urljoin(base, htmldecode(match.group(1)))
+        links.append(link)
+
+    assert not set(pip_links) ^ set(links)
```
mgedmin, line 103: This is just an obfuscated way of saying
```suggestion
    assert set(pip_links) == set(links)
```

buildout.cfg
gotcha: I kept it for now until I manage to run coverage properly again.

dev.py
gotcha: I am fine with this change but do not know why it is better to specify the prefix.
I am eager to learn why.

src/zc/buildout/__init__.py
gotcha: Right, I forgot to update minimal setuptools requirement.
I am eager to move to recent setuptools like 41.3.0 (chosen because it is the last one that does not complain about 2.7 if I remember well). What would be your opinion ?

src/zc/buildout/__init__.py
gotcha: This is code stolen from setuptools as I am monkeypatching.

https://github.com/pypa/setuptools/blob/v44.1.0/setuptools/package_index.py#L338

I decided to not touch what does not need to be touched. What is your take on this ?

src/zc/buildout/__init__.py
gotcha: Good catch :-)

src/zc/buildout/__init__.py
gotcha: I totally agree this could be a big risk.
This is only needed for PEP345 support. This is one of the questions I wanted to tackle with the buildout community. Would we vendor include what is neded by `parse_links`, `_check_link_requires_python` and `TargetPython`.

src/zc/buildout/easy_install.py
```diff
@@ -269,6 +273,8 @@ def __init__(self,
 
         self._index_url = index
         path = (path and path[:] or []) + buildout_and_setuptools_path
+        if force_eggs:
+            path = [p for p in path if not p.endswith('site-packages')]
```
gotcha, line 277: I am uneasy with this way of checking `site-packages`. Would anyone have a better solution ?

dev.py
mgedmin: I have scars from when I used to maintain a buildbot for the Zope 3.4 KGS.  My /tmp would fill up with a bunch of random temporary directories from tests that forgot to clean up after themselves, and I had a hard time tracking which package's test suite was responsible.

Asking `mkdtemp()` to tag the created temporary directory with a recognizable prefix (or suffix) is basic courtesy in my book.

src/zc/buildout/__init__.py
mgedmin: From what I've seen on distutils-sig@, the pip developers are open to people extracting bits of reusable logic from pip into standalone libraries like `packaging` or `distlib` or `pep517` (that they then will vendor).  That would be the most future-proof way of sharing pip's logic.

It might be a good idea to discuss this on distutils-sig@.

src/zc/buildout/__init__.py
mgedmin: I don't do many buildout-based deployments nowadays, but in the past I'd decided that the strategy of always upgrading setuptools to the latest version was the best way to avoid problems.  So, I'm fine with depending on whatever.

Other buildout users might have different opinions.

src/zc/buildout/__init__.py
mgedmin: If there's reason to believe in the future someone may want to merge the copy-pasted code with latest upstream changes, then keeping our changes to the minimum is the right thing.

I didn't realize this was copypasta.

src/zc/buildout/easy_install.py
mgedmin: No ideas on solutions, I haven't even looked at the problem, but I'm now wondering about Debian's `dist-packages`.

k4ml:
It surprising this actually works with our buildout config ;)

I'll look more into this PR.

gotcha:
@k4ml I will be pleased to get your feedback. In which context do you use buildout ?

icemac:
I tried this branch in union.cms on Python 3.7 (branch installed in a `venv` using `bin/pip install -e "git+https://github.com/gotcha/buildout.git@use-pip#egg=zc.buildout"`) and get the following traceback right away:

```python
$ bin/buildout
While:
  Installing.
  Loading extensions.
  Getting distribution for 'pip==20.0.2'.

recipe being used:
  File "/.../src/zc.buildout/src/zc/buildout/buildout.py", line 2213, in main
  File "/.../src/zc.buildout/src/zc/buildout/buildout.py", line 659, in install
    self._load_extensions()
  File "/.../src/zc.buildout/src/zc/buildout/buildout.py", line 1226, in _load_extensions
    newest=self.newest, allow_hosts=self._allow_hosts)
  File "/.../src/zc.buildout/src/zc/buildout/easy_install.py", line 964, in install
  File "/.../src/zc.buildout/src/zc/buildout/easy_install.py", line 734, in install
  File "/.../src/zc.buildout/src/zc/buildout/easy_install.py", line 549, in _get_dist
    dist, avail = self._satisfied(requirement)
  File "/.../src/zc.buildout/src/zc/buildout/easy_install.py", line 345, in _satisfied
    return None, self._obtain(req, source)
  File "/.../src/zc.buildout/src/zc/buildout/easy_install.py", line 472, in _obtain
    if index.obtain(requirement) is None:
  File "/.../lib/python3.7/site-packages/setuptools/package_index.py", line 503, in obtain
    self.find_packages(requirement)
  File "/.../lib/python3.7/site-packages/setuptools/package_index.py", line 487, in find_packages
    self.scan_url(self.index_url + requirement.unsafe_name + '/')
  File "/.../lib/python3.7/site-packages/setuptools/package_index.py", line 827, in scan_url
    self.process_url(url, True)
  File "/.../src/zc.buildout/src/zc/buildout/__init__.py", line 123, in process_url
    assert set(pip_links) == set(links)
```

`set(pip_links)` `set(links)` are disjuct sets because the URLs differ in the URL encoding:

`https://development.verdi.de/devpi/root/pypi/%2Bf/070/e4bf493c7c2c9/pip-18.0-...` vs.
`https://development.verdi.de/devpi/root/pypi/+f/070/e4bf493c7c2c9/pip-18.0-...`
(pip_links vs. links)

If I remove this assertion it runs the buildout successfully. In the `buildout.cfg` we use `mr.developer` and a bunch of recipes. At least the generated test script seems to be correct, it is able to run the tests.

icemac:
I did not look into the code changes in this PR.

gotcha:
@icemac Thanks for trying the code ! And debugging what went wrong with the assertion.

Do I understand correctly that, outside the assertion issue, your trial has gone well ?

Is the union.cms buildout which you used public ?

(The assertion is there in the short run while testing the PR. It should go away as there are no need to parse the index pages twice. Nevertheless, I'll check if there is a way to improve the assertion in order to keep it for now.)

src/zc/buildout/__init__.py
mauritsvanrees: Note that `assert` statements may get optimised away when compiled into a `.pyo` file.
See for example:
https://stackoverflow.com/questions/2055557/python-optimized-mode#2167639
So a `raise ValueError` or similar would be better, unless you are fine with ignoring this assert statement in optimised code.

src/zc/buildout/__init__.py
mauritsvanrees: I think only `pip` complains about Python 2.7, not `setuptools`.
Version 42 deprecates easy_install.
But 41.3 sounds like roughly a fine number to have as minimum.

mauritsvanrees:
- I do not have time currently to look at this properly, but I support the direction.
- This would need to be a buildout 3.0.0.
- Like you suggest, this could use an alpha release (probably several) so people can more easily try it out.
- I think support/documentation for installing with `bootstrap.py` can be removed, if this has not been done already.
- I would not mind if this is for Python 3 only, and more recent setuptools (and pip), especially if this makes testing easier. Testing buildout is hard, as can clearly be seen on Travis the past year. Downside for Plone would be that we cannot use it in Plone 5.2, because that still also has to support 2.7. But more likely is that Plone 5.2 will keep to buildout 2.13 even if a newer buildout still supports 2.7: the change is too big. (I would totally use newer buildout myself on 5.2 in customer projects.)

icemac:
@gotcha Yes, outside the assertion everything went fine.
The `buildout` for union.cms is not publicly available.
The URLs come from a devpi server we use as PyPI mirror.

gotcha:
@icemac Would you mind testing again with the fix I did to the broken assertion ?

gotcha:
@mauritsvanrees 
- Thanks for your comment
- I still have customers under Plone 4. I plan to use this updated `buildout` to help with the migrations. IOW, I'll support Python 2 for now.
- I still need to confirm that `bootstrap.py`can go away. In this case, you are right that the documentation needs to go away.

k4ml:
I created dummy buildout that resemble our internal buildout to test this. So far everything works.

https://github.com/k4ml/cutecat

src/zc/buildout/__init__.py
gotcha: As a first step, I made it less fragile by making it optional and warning if it fails.

src/zc/buildout/__init__.py
gotcha: You are right, I mixed complaining about 2.7 and `easy_install` deprecation.

I will go for last before 42: 41.6.0

icemac:
@gotcha I updated and now the buildout successfully ran through.

ale-rt:
Thanks @gotcha for your amazing and clever work!
I really think this is needed even if relying on pip internals has its drawbacks.

I tested this branch in a virtualenv with a plone buildout.

I did:
```
python3.8 -m venv .
./bin/pip3.8 install -e git+ssh://git@github.com/gotcha/buildout.git@use-pip#egg=zc.buildout
```

When I ran buildout I got this error:
```
  File "/home/ale/Code/plone/projects/ipdb/src/zc.buildout/src/zc/buildout/easy_install.py", line 1886, in _move_to_eggs_dir_and_compile
    [tmp_loc] = call_pip_install(dist.location, tmp_dest)
  File "/home/ale/Code/plone/projects/ipdb/src/zc.buildout/src/zc/buildout/easy_install.py", line 1690, in call_pip_install
    distinfo_dir = [
IndexError: list index out of range
```

In my case `split_entries` was:
```
(Pdb) split_entries
[('yafowil.widget.datetime-1.11-nspkg', '.pth'), ('yafowil.widget.datetime-1.11', '.egg-info'), ('yafowil', '')]
```

So this code:
```
        distinfo_dir = [base + ext for base, ext in split_entries if ext == ".dist-info"][0]
```
was raising the `IndexError`.
The error disappeared after I installed the wheel package in the virtualenv:
```
./bin/pip3.8 install wheel
```
Should we add `wheel` to the installation requirements?

Beside that small issue I thank you again for the amazing job!

gotcha:
@ale-rt Thanks for testing ! And thanks pointing that the PR has an implicit dependency on `wheel`.

gotcha:
#506 has less code changed and cleaner history.

PR #504, opened 2020-05-08
run Travis tests in CentOS container

gotcha:
Now that they pass on GitHub Actions


gotcha:
Tests under Python 2.7 are disabled temporarily.

PR #505, opened 2020-05-09
make install of unpinned setuptools much quicker

gotcha:
by patching `pkg_resources.Distribution`.


mgedmin:
It would be interesting to see some numbers.

gotcha:
see https://github.com/pypa/setuptools/issues/2089#issuecomment-627662988

PR #506, opened 2020-05-11
Use pip install instead of setuptools easy_install

gotcha:
The meat of the change is in src/zc/buildout/easy_install.py

Some tests had to be refactored a bit to make them more robust.

`tox` is removed to avoid issues due to sharing the same directory
between different virtual environments; `Makefile` enriched to make
per python virtual envs depend on source files.

No need to be afraid of default Python install as we depend on `pip`.
`bootstrap.py` is gone as not needed anymore.


Makefile
```diff
@@ -74,12 +74,48 @@ python: $(PYTHON_PATH)/bin/$(PYTHON_EXE)
...
+SRC_FILES := $(shell find $(HERE)/src ! -path '*egg-info*' \( -name '*.py' -o -name '*.txt' -o -name '*.test' \) )
+RCP_FILES := $(shell find $(HERE)/zc.recipe.egg_ ! -path '*egg-info*' \( -name '*.py' -o -name '*.txt' -o -name '*.rst' -o -name '*.cfg' -o -name '*.in' \) )
+DOC_FILES := $(shell find $(HERE)/doc -name '*.rst' -o -name '*.txt')
+
+ALL_COPY := $(subst $(HERE),$(VENV),$(SRC_FILES) $(DOC_FILES) $(RCP_FILES) $(ROOT_FILES))
+# Generate rules to map sources into targets
+$(foreach s,$(ALL_COPY),$(eval $s: $(VENV)/bin/$(PYTHON_EXE) $(subst $(VENV),$(HERE),$s)))
+
+$(ALL_COPY):
+	@test -d "$(dir $@)" || mkdir -p $(dir $@)
```
mgedmin, line 88: `mkdir -p` silently does nothing if the directory already exists, so you could omit the `test -d ... ||` part.

Makefile.configure
```diff
@@ -2,15 +2,15 @@ ifeq ($(PYTHON_VER),2.7)
...
+	PYTHON_CONFIGURE_ARGS ?=
 endif
 ifeq ($(PYTHON_VER),3.7)
-	PYTHON_CONFIGURE_ARGS ?= --without-ensurepip
+	PYTHON_CONFIGURE_ARGS ?=
 endif
 ifeq ($(PYTHON_VER),3.8)
-	PYTHON_CONFIGURE_ARGS ?= --without-ensurepip
+	PYTHON_CONFIGURE_ARGS ?=
 endif
```
mgedmin, line 15: Maybe simplify the entire series of python version checks into a single unconditional

```
PYTHON_CONFIGURE_ARGS ?=
```

dev.py
```diff
@@ -16,63 +16,94 @@
...
+    try:
+        get_pip = os.path.join(tmp, 'get-pip.py')
+        with open(get_pip, 'wb') as f:
+           f.write(urlopen('https://bootstrap.pypa.io/get-pip.py').read())
 
-# XXX use a more permanent ez_setup.py URL when available.
-exec(urlopen('https://bootstrap.pypa.io/ez_setup.py').read(), ez)
-ez['use_setuptools'](to_dir='eggs', download_delay=0)
+        if subprocess.call([sys.executable, get_pip]):
+            raise RuntimeError("pip failed.")
```
mgedmin, line 59:
```suggestion
            raise RuntimeError("Failed to install pip")
```

dev.py
```diff
@@ -16,63 +16,94 @@
...
+    ):
     raise RuntimeError("buildout build failed.")
 
+import pkg_resources
+
 pkg_resources.working_set.add_entry('src')
 
 import zc.buildout.easy_install
 zc.buildout.easy_install.scripts(
     ['zc.buildout'], pkg_resources.working_set , sys.executable, 'bin')
```
mgedmin, line 97: Not part of your changes, but I can't notice a space in front of a comma and not mention it.

setup.cfg
```diff
@@ -11,7 +11,6 @@ ignore =
     Makefile.configure.Darwin
     zc.recipe.egg_/*
     zc.recipe.egg_
-    tox.ini
     .github/*
     .github
```
mgedmin, line 15: With check-manifest >= 0.42 you may have to change `.github/*` to `.github/**`, to ignore files in subdirectories recursively.  (The two-* version also works with check-manifest < 0.42.)

src/zc/buildout/easy_install.py
```diff
@@ -1658,7 +1666,117 @@ def call_easy_install(spec, dest):
...
+                    "__import__('pkg_resources').declare_namespace(__name__)"
+                )
+
+    # Remove `bin` directory if needed
+    # as there is no way to avoid script installation
+    # when running `pip install`
+    ep_file = os.path.join(dest, distinfo_dir, 'entry_points.txt')
+    if os.path.exists(ep_file):
+        with open(ep_file) as f:
+            if "console_scripts" in f.read():
```
mgedmin, line 1711: Should we also check for `gui_scripts`?

src/zc/buildout/easy_install.py
```diff
@@ -1658,7 +1666,117 @@ def call_easy_install(spec, dest):
...
+                bin_dir = os.path.join(dest, 'bin')
```
mgedmin, line 1712: I'm now thinking about Windows and how the virtualenv `bin` directory is named `Scripts` there.

src/zc/buildout/easy_install.py
```diff
@@ -1658,7 +1666,117 @@ def call_easy_install(spec, dest):
...
+            shutil.move(top_level_dir, egg_dir)
+            continue
+        # as module
+        top_level_py = top_level_dir + '.py'
+        if os.path.exists(top_level_py):
+            shutil.move(top_level_py, egg_dir)
+            top_level_pyc = top_level_dir + '.pyc'
+            if os.path.exists(top_level_pyc):
+                shutil.move(top_level_pyc, egg_dir)
+            continue
```
mgedmin, line 1748: What about `__pycache__` subdirectories?

src/zc/buildout/easy_install.py
```diff
@@ -1658,7 +1666,117 @@ def call_easy_install(spec, dest):
...
+
+
+SLASH_BIN_SLASH = os.path.join('', 'bin', '')
+
+
+def _do_not_keep_leftover(filename):
+    return (
+        filename.endswith('.pyc') or
+        filename.endswith('.pyo') or
+        SLASH_BIN_SLASH in filename
```
mgedmin, line 1778: Ooh this could bite us if there's ever a subpackage named `bin`.

(Also, what about Scripts on Windows?)

src/zc/buildout/easy_install.py
```diff
@@ -37,6 +37,8 @@
 import zc.buildout
 import zc.buildout.rmtree
 import warnings
+import csv
+from pip._internal.utils.misc import FakeFile
```
mgedmin, line 41: No protection against pip internal changes here.

(FakeFile seems like a trivial StringIO-ish class used only to feed the .dist-info/RECORD file into csv.reader(), and I wonder why can't we feed the file directly?)

src/zc/buildout/easy_install.py
```diff
@@ -175,7 +178,7 @@ def _execute_permission():
     return 0o777 - current_umask
 
 
-_easy_install_cmd = 'from setuptools.command.easy_install import main; main()'
+_pip_install_cmd = 'from pip._internal.cli.main import main; main()'
```
mgedmin, line 181: No protection against pip internal changes here.

src/zc/buildout/patches.py
```diff
@@ -56,3 +56,113 @@ def set_location(self, l):
...
+                page = self.process_index(url, page)
+
+        setattr(PackageIndex, 'process_url', process_url)
+
+    except ImportError:
+        import logging
+        logger = logging.getLogger('zc.buildout.patches')
+        logger.warning('Requires-Python support missing. \n\n',
+            exc_info=True
+        )
```
mgedmin, line 164: :+1:

src/zc/buildout/patches.py
```diff
@@ -56,3 +56,113 @@ def set_location(self, l):
...
+        from pip._internal.index.collector import parse_links
+        from pip._internal.index.collector import _clean_link
+        from pip._internal.index.package_finder import _check_link_requires_python
+        from pip._internal.models.target_python import TargetPython
+        from pip._vendor import six
+        from pip._vendor.six.moves import urllib
+
+        PY_VERSION_INFO = TargetPython().py_version_info
+
+        def process_url(self, url, retrieve=False):
```
mgedmin, line 79: Might be a good idea to comment that this method was copied and pasted from setuptools.package_index.PackageIndex.process_url from setuptools version XX.YY, and then add box comments around the places where it was modified.

src/zc/buildout/patches.py
```diff
@@ -56,3 +56,113 @@ def set_location(self, l):
...
+            pip_links = [l.url for l in plinks]
+            if not isinstance(page, str):
+                # In Python 3 and got bytes but want str.
+                page = page.decode(charset, "ignore")
+            f.close()
+
+            links = []
+            for match in HREF.finditer(page):
+                link = urllib.parse.urljoin(base, htmldecode(match.group(1)))
+                links.append(_clean_link(link))
```
mgedmin, line 145: So if this is how setuptools is parsing indexes, I wonder why they haven't added `data-requires-python` parsing upstream.  Is there an open bug?  Are the maintainers opposed to having this logic check in setuptools itself?

src/zc/buildout/tests.py
```diff
@@ -3549,9 +3558,10 @@ def updateSetup(test):
     test.globs['new_releases'] = new_releases
     ws = getWorkingSetWithBuildoutEgg(test)
     # now let's make the new releases
-    for dist in zc.buildout.easy_install.buildout_and_setuptools_dists:
-        makeNewRelease(dist.key, ws, new_releases)
-        os.mkdir(os.path.join(new_releases, dist.key))
+    # TOD0 enable new releases of pip wheel setuptools
```
mgedmin, line 3561:
```suggestion
    # TODO enable new releases of pip wheel setuptools
```

tox.ini
```diff
@@ -1,20 +0,0 @@
...
-    {toxinidir}/bin/test -1 -vvv -c
-    coverage combine
-    coverage report
-# since we're removing setuptools we can't possibly reuse the virtualenv
-recreate = true
-# if the user has ccache installed, PATH may contain /usr/lib/ccache that has a
-# gcc wrapper that fails to build anything when buildout's tests set HOME to a
-# non-existent directory under /tmp
-setenv =
-    PATH=/usr/bin:/bin
```
mgedmin, line 20: I'm going to miss tox.

> tox is removed to avoid issues due to sharing the same directory between different virtual environments;

Which directory would that be?

Does it affect parallel tox runs only, or all tox runs?

CHANGES.rst
```diff
@@ -4,6 +4,13 @@ Change History
 2.13.4 (unreleased)
 ===================
 
+- Patch ``setuptools.PackageIndex`` to support ``Requires-Python``.
+  Fragile because it relies on ``pip._internal``.
+  Reason why imports are protected to avoid breaking buildout in case
+  ``pip`` changes too much.
```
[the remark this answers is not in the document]
mgedmin: > Will it emit a warning about pip incompatibility meaning that Requires-Python is ignored in that case?

Yes.  I saw the warning, but forgot to go back and delete this question.

src/zc/buildout/easy_install.py
mgedmin: On second thought, it's fine to not copy those.

Now I'm thinking about filenames embedded in pyc files.  Do we want to copy pyc files?  Maybe, if there happens to be a binary-only module where there are pyc files without matching py files (a use case heavily frowned upon, but supported, even on Python 3).

src/zc/buildout/patches.py
mgedmin: I was unable to find a bug in the upstream issue tracker.  The closest is https://github.com/pypa/setuptools/issues/1633

tox.ini
gotcha: The way `tox` was configured, the repository root was shared between all `tox` environments.
Concretely, eggs for multiple Python versions would be installed in the `eggs` dir in your checkout.

I did not manage to find a way to solve that issue. Anyone with `tox` experience is welcome to revive it.

.gitignore
```diff
@@ -6,10 +6,10 @@ develop-eggs/
 eggs/
 parts/
 python*/
+venvs/
 *.pyc
 *.egg
 *.egg-info
 dist
-.tox
```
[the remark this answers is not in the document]
gotcha: Good point

CHANGES.rst
```diff
@@ -4,6 +4,13 @@ Change History
...
+
+- Use ``pip install`` instead of deprecated ``setuptools.easy_install``.
```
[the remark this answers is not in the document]
gotcha: I think I will go for major.

CHANGES.rst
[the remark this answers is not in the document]
gotcha: I'll improve the changelog.

Makefile
gotcha: Yup

Makefile.configure
gotcha: And yup again

dev.py
gotcha: Better

src/zc/buildout/patches.py
gotcha: What do you mean by box comments ?

mgedmin:
```python
... upstream code ...
# --- LOCAL CHANGES MADE HERE: ---
... modified code ...
# --- END OF LOCAL CHANGES ---
... upstream code ...
```

tox.ini
mgedmin: Thanks for clarifying!

src/zc/buildout/easy_install.py
gotcha: Right

src/zc/buildout/easy_install.py
gotcha: Good point, will go away

src/zc/buildout/tests.py
gotcha: Thx

gotcha:
@mgedmin Thanks for a great review again !

About the checks, I am not sure I understand how/where the GitHub actions are shown. You can find them in https://github.com/gotcha/buildout/actions which is not really convenient.

I'll try to add a workflow on Windows.

src/zc/buildout/easy_install.py
mgedmin: I should clarify that this is not criticism.  I wanted to note where the dependencies are to see how exposed we would be to pip changing its implementation details.

The index parsing code is protected somewhat (with slight regression in behaviour), the FakeFile import can be replaced, and this one seems very unlikely to change/very easy to fix if changed.

So yay!

mgedmin:
Maybe actions are simply misconfigured?  I see `on: [push]`and maybe PRs need their own separate trigger?

Maybe we'd see checks _here_ if you'd pushed to a branch in this repo instead of your fork?

I don't know.  From what I've seen so far, GH actions are terribly underbaked.

gotcha:
@mgedmin Would you mind to review the commits I pushed where I made fixes according to your comments ? 

BTW, it seems you were right about missing `pull_request` in GH workflow.

Makefile
```diff
@@ -85,7 +85,7 @@ ALL_COPY := $(subst $(HERE),$(VENV),$(SRC_FILES) $(DOC_FILES) $(RCP_FILES) $(ROO
 $(foreach s,$(ALL_COPY),$(eval $s: $(VENV)/bin/$(PYTHON_EXE) $(subst $(VENV),$(HERE),$s)))
 
 $(ALL_COPY):
-	@test -d "$(dir $@)" || mkdir -p $(dir $@)
+	@mkdir -p $(dir $@)
```
mgedmin, line 88: JSYK there was more than one instance of `test -d || mkdir -p` in the Makefile, but I didn't tag them all.

src/zc/buildout/easy_install.py
```diff
@@ -178,7 +182,7 @@ def _execute_permission():
     return 0o777 - current_umask
 
 
-_pip_install_cmd = 'from pip._internal.cli.main import main; main()'
+_pip_install_cmd = 'from pip.__main__ import _main; _main()'
```
mgedmin, line 185: :100:

src/zc/buildout/easy_install.py
```diff
@@ -1747,18 +1752,16 @@ def make_egg_after_pip_install(dest, distinfo_dir):
...
+        with open(os.path.join(egg_dir, distinfo_dir, 'RECORD'), newline='') as f:
+            all_files = [row[0] for row in csv.reader(f)]
+    else:
+        with open(os.path.join(egg_dir, distinfo_dir, 'RECORD'), 'rb') as f:
+            all_files = [row[0] for row in csv.reader(f)]
 
     # There might be some c extensions left over
     for entry in all_files:
-        if _do_not_keep_leftover(entry):
+        if entry.endswith('.pyc') or entry.endswith('.pyo'):
```
mgedmin, line 1764:
```suggestion
        if entry.endswith(('.pyc', '.pyo')):
```

src/zc/buildout/easy_install.py
```diff
@@ -1768,17 +1771,6 @@ def get_all_files(it):
...
-
-
-def _do_not_keep_leftover(filename):
-    return (
-        filename.endswith('.pyc') or
-        filename.endswith('.pyo') or
-        SLASH_BIN_SLASH in filename
-    )
-
-
```
mgedmin, line 1663: Now that you've removed that logic, I wonder if scripts installed via the distutils scripts mechanism (as opposed to sentry-points) will be left over?

If you want an example of a package that uses scripts, try [ansible-cmdb](https://github.com/fboender/ansible-cmdb).

src/zc/buildout/easy_install.py
```diff
@@ -1625,17 +1637,17 @@ class IncompatibleConstraintError(zc.buildout.UserError):
...
     distribution specified by `spec` into `dest`.
     Returns all the paths inside `dest` created by the above.
     """
-    path = setuptools_path
+    path = pip_path
 
     args = [sys.executable, '-c',
             ('import sys; sys.path[0:0] = %r; ' % path) +
-            _easy_install_cmd, '-mZUNxd', dest]
+            _pip_install_cmd, 'install', '--no-deps', '-t', dest]
```
mgedmin, line 1650: I was thinking the other day (a bad habit, I know).  pip vendors a bunch of libraries (requests, urllib3, certifi, etc).  Debian doesn't allow multiple copies of a library in the distribution, so Debian patches pip to unvendor all these and makes it use their packaged versions.

I wonder, if you happen to use Debian's /usr/bin/python3 and Debian's python3-pip, will pip_path contain everything needed to run pip?

I'm guessing it will: if we use Debian's pip, then pip_path ought to contain /usr/lib/python3.X/dist-packages, which has all the needed bits.  If we don't use Debian's pip, then pip'll use its own vendored copies.

I wonder if it would be a good idea to set up a Debian-based build in the CI.  (I'm not going to insist.)

src/zc/buildout/easy_install.py
gotcha: `zc.buildout 2.13.3` does not know how to install the `ansible-cmdb` scripts.
Have you already used both combined ?

src/zc/buildout/easy_install.py
gotcha: I guess the easiest would be to run the tests in a Debian based container. is that what you mean ?

src/zc/buildout/easy_install.py
mgedmin: No, it's just the one package I know that still uses scripts instead of entry points.

(I know because I submitted a PR to make it use entry points a long time ago, but it was never merged.)

Curious that old zc.buildout can't handle scripts.  I thought easy_install supported them?

mgedmin: Oh, right, zc.buildout took over the role of generating scripts by itself.

Anyway I don't care whether scripts work, I just wondered if they cause complications for this copy-files-except-ones-we-dont-want piece of logic.

src/zc/buildout/easy_install.py
mgedmin: Yes.

doc/topics/bootstrapping.rst
```diff
@@ -27,15 +27,25 @@ Avoid modifying the python environment.
   existed, it might not have been possible to modify the environment
   without building Python from source.
 
-  Unfortunately, doing this requires :ref:`using a bootstrap script
-  <using-a-bootstrap-script>`.
+Installing from scratch
+=======================
+
+  We recommended to install ``buildout`` via ``pip install``
+ inside a ``virtualenv``:
```
mgedmin, line 34: Indentation here creates unnecessary nested blockquotes.
```suggestion
We recommended to install ``buildout`` via ``pip install``
inside a ``virtualenv``:
```

gotcha:
Thanks a lot to those that helped with review and trial. In particular to @mgedmin !

PR #527, opened 2020-06-19
Fix issue with incrementing options from .buildout/default.cfg

### LINKED ISSUE #525: '+=' syntax not working for buildout:extensions when setup in default.cfg
Hi,

when I globally activate extension(s) in the ~/.buildout/default.cfg file like this:
```
[buildout]
extensions=some.extension.a
```

..but also use project specific extension(s) in some buildout.cfg files like this:
```
[buildout]
extensions+=some.extension.b
```

..then only some.extension.b gets loaded. Thus some.extension.a only gets loaded when buildout.cfg files do not contain any buildout:extensions values at all. Tested with zc.buildout 2.8.0 and 2.13.3. Is it a known restriction or maybe just an unknown bug and simple to fix?
Thanks a lot!

Best regards

gotcha:
with some cleanup to help to understand code base.

fixes #525


mgedmin:
I don't know if I'm the best person to review this.

I'll try to find the time to look at it some time this week.

src/zc/buildout/buildout.py
```diff
@@ -251,15 +253,17 @@ def _print_annotate(data, verbose, chosen_sections, basedir):
 
 
 def _unannotate_section(section):
+    raw = {}
     for key in section:
-        section[key] = section[key].value
-    return section
+        raw[key] = section[key].value
+    return raw
```
mgedmin, line 259: Why not use a dict comprehension?

```
    return {key: value.value for key, value in section.items()}
```

src/zc/buildout/buildout.py
```diff
@@ -110,6 +110,8 @@ def overrideValue(self, sectionkey):
         self.value = sectionkey.value
         if sectionkey.history[-1].operation not in ['ADD', 'REMOVE']:
             self.addToHistory("OVERRIDE", sectionkey.value, sectionkey.source)
+        else:
+            self.history = copy.deepcopy(sectionkey.history)
```
mgedmin, line 114: deepcopy() always scares me.  I did not investigate whether its use is safe here.

src/zc/buildout/buildout.py
```diff
@@ -251,15 +253,17 @@ def _print_annotate(data, verbose, chosen_sections, basedir):
...
+    return raw
 
 
 def _unannotate(data):
+    raw = {}
     for key in data:
-        data[key] = _unannotate_section(data[key])
-    return data
+        raw[key] = _unannotate_section(data[key])
+    return raw
```
mgedmin, line 266: Why not use a dict comprehension?

```
    return {key: _unannotate_section(value) for key, value in data.items()}
```

gotcha:
@mgedmin Thanks a lot for taking a look.

The change is actually quite big. Reason why I tried to provide meaningful commits. Might make sense to review commit by commit.

src/zc/buildout/buildout.py
gotcha: I did not use them often enough. Thanks for pointing this.

src/zc/buildout/buildout.py
gotcha: AFAIU, the issues with `deepcopy` do not apply here as `history` is a list of `HistoryItem` instances which only holds strings.

gotcha:
@jamadden Would you have some time to review this PR ? It required some cleanup, reason why it is maybe best to review it commit by commit.

jamadden:
I'm taking a look. I've never had this problem though so I can't check the functionality myself...

fschulze:
I also took a look. Judging from the added tests and my understanding of the original issue these changes look good to me. It really helped to look at the changes commit by commit as suggested.

jamadden (commented):
I really appreciate trying to make more datastructures immutable! I do worry if there might be compatibility problems with extensions that might assume otherwise, though.

Overall, it looks fine to me. The tests pass, and there are new tests, so that's good.

gotcha:
@jamadden @fschulze Thx a lot for your reviews.

PR #533, opened 2020-09-17
Evaluate environment markers for requirement strings

hbrunn:
support [environment markers](https://www.python.org/dev/peps/pep-0496), this way we can say things like

```
eggs =
    lib == version1; python_version >= '3.0'
    lib == version2; python_version < '3.0'
```

### DIFF
--- a/doc/getting-started.rst
+++ b/doc/getting-started.rst
@@ -439,6 +439,39 @@ where you list them, as in:
 
 In this example, we've requested a version of bobo less than 5.0.
 
+You can also add `environment markers <https://python.org/dev/peps/pep-0496>`_
+to restrict some requirements to i.e. a certain platform or python version:
+
+.. code-block:: ini
+
+  [bobo]
+  recipe = zc.recipe.egg
+  eggs =
+      bobo ==2.2.0; python_version < '3.0'
+      bobo ==2.3.0; python_version >= '3.0'
+
+.. -> src
+
+   >>> prefix = """
+   ... [buildout]
+   ... parts = bobo
+   ... """
+   >>> with open('buildout.cfg', 'w') as f:
+   ...     _ = f.write(prefix)
+   ...     _ = f.write(src)
+
+   >>> import shutil
+   >>> import sys
+   >>> v = sys.version_info
+   >>> shutil.rmtree('eggs')
+   >>> run_buildout('buildout show-picked-versions=true')
+   >>> yup([n for n in ls('eggs') if n.startswith('bobo-2.3.0-')])\
+   ... if v.major >= 3 else\
+   ... yup([n for n in ls('eggs') if n.startswith('bobo-2.2.0-')])
+   >>> yup('bobo==2.3.0' in read('out'))\
+   ... if v.major >= 3 else\
+   ... yup('bobo==2.2.0' in read('out'))
+
 The more common way to pin a version is using a ``versions`` section:
 
 .. code-block:: ini
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -678,8 +678,14 @@ class Installer(object):
 
         for_buildout_run = bool(working_set)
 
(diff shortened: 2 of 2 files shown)


mgedmin:
This could be a good idea, but it needs documentation at the very least.

BTW did you know what buildout already supports an [alternative syntax](http://www.buildout.org/en/latest/topics/variables-extending-and-substitutions.html#conditional-configuration-sections) for this?

```ini
eggs =
    lib

[versions:python2]
lib = version1

[versions:python3]
lib = version2
```

hbrunn:
@mgedmin thank you, I did not know about the conditional sections. Just added a sentence about the marker strings.

mgedmin:
Are we getting build failures because modern pip emits

> DEPRECATION: Python 3.5 reached the end of its life on September 13th, 2020. Please upgrade your Python as Python 3.5 is no longer maintained. pip 21.0 will drop support for Python 3.5 in January 2021. pip 21.0 will remove support for this functionality.

and such?  We ought to fix that in master, maybe by adding a renormalizer that removes anything starting with `^DEPRECATION:`.

hbrunn:
or maybe patch pip after installation to not print the depreciation message?

jamadden:
It's possible to set an environment variable to disable that warning (here in a .travis.yml):

```yaml
env:
    global:
        # Don't get warnings about Python 2 support being deprecated. We
        # know. The env var works for pip 20+.
        - PIP_NO_PYTHON_VERSION_WARNING=1
```

hbrunn:
good idea, I added a commit for that and another to support python 3.7 for the conditional sections

mgedmin:
We don't use just Travis, we also use GitHub Actions.  It would be best to stick it into `os.environ['PIP_NO_DEPRECATION_THINGY'] = '1'` somewhere in a test `setUp()`.

I don't think fixing tests and adding Python version support fall into the scope of this PR; it'd be better to do those separately.  (Also, as long as we're adding 3.7, let's not forget that 3.8 exists too!).

hbrunn:
my hope was to just rebase-merge all of this in one go, but indeed one PR per feature is cleaner

gotcha:
@hbrunn Can you rebase on latest master where your fix for tests has been merged ?

hbrunn:
@gotcha done

gotcha:
@hbrunn Can you also add a test and a changelog entry ?

gotcha:
@hbrunn to avoid conflicts, forget the changelog.

I'll do it myself when releasing to avoid conflicts.

And I'll do a PR to use ``towncrier`` to avoid the issue later.

jamadden:
FWIW, while I have used (and liked!) towncrier in a project, it seems to be a bit of a barrier to contributors, especially "drive-by" contributors. I suggest it has the most value when there are a large number of active pull requests, or when there's lots of backporting or other multi-branch development going on. I'm not sure those cases apply to buildout at this time.

gotcha:
Well, in this PR, @mgedmin asked to separate fixes in different PRs. This by itself generates conflicts.

I wonder which is worse; conflicts or towncrier...

mgedmin:
The fix for deprecation warnings breaking CI is #534, and I don't think it should conflict with this PR?

gotcha:
@hbrunn Do you feel like adding a test ?

hbrunn:
@gotcha I added somethwat nonsensical markers in https://github.com/buildout/buildout/pull/533/commits/1622803ed1e7a75e19e12f964e0b94673420c1fa#diff-cb79b0dbf1ef0af92832aade7ac08c174b06629ffb52a7c845d06cf6025ae8b1R15

gotcha:
@hbrunn I do not understand your intent with the nonsensical markers.

hbrunn:
@gotcha and rightly so, one year younger me was too lazy to read up on doctest it seems. Now I think I did it the right way.

For all I can tell the CI failures are unrelated.

gotcha:
Feature merged via #602

Issue #444, opened 2018-04-03
Is a Zope Foundation contributor's agreement needed to contribute to this repository?


idgserpro:
Didn't see anything in https://github.com/buildout/buildout/blob/master/DEVELOPERS.txt. Such agreement in needed in Plone world, not so sure it's here as well...

idgserpro:
@jimfulton @leorochael @reinout

idgserpro:
We have a small addition we think could be useful in https://github.com/buildout/buildout/blob/master/bootstrap/bootstrap.py. 

We work with Plone, but for legal reasons we can't sign the contributor's agreement. That's why we're asking here.

You can close this issue after answering.

Thanks in advance!

@jimfulton @leorochael @reinout

gotcha:
@idgserpro If it is a small addition, just provide the diff in a gist that you reference here. And one of us will include it.

idgserpro:
Will you follow up https://community.plone.org/t/zope-contributor-agreement-process/8495 in the future?

idgserpro:
@gotcha 

RIght after https://github.com/buildout/buildout/blob/9ba1e3170a474e6139c9a7eadd32a498c22cb3bb/bootstrap/bootstrap.py#L77:

```
  74   │ options, args = parser.parse_args()
  75   │ if options.version:
  76   │     print("bootstrap.py version %s" % __version__)
  77   │     sys.exit(0)
```

It would be nice to add:

```python
python26 = (sys.version_info.major == 2 and sys.version_info.minor == 6)
if not options.buildout_version and python26:
    print("You're running bootstrap-buildout.py with python 2.6.")
    print("You need to specify an older buildout version for running this script.")
    print("Use the --buildout-version option. The last buildout version compatible with python 2.6 is 2.8.0.")
    sys.exit(0)
```

Don't forget to change `__version__` at the top of the file.

@mauritsvanrees since you were the last one missing with that file.

icemac:
@gotcha Is this a decision valid for the whole project? Should it be documented in the repository so the question is not asked over and over again?

gotcha:
@icemac I do not have any authority to decide. My answer above seemed pragmatic and constructive.

@idgserpro I am sorry I failed to commit those changes I suggested I would apply myself. Do you still need the changes to be applied ?

idgserpro:
[...]

PR #534, opened 2020-09-17
Set PIP_NO_PYTHON_VERSION_WARNING during tests

hbrunn:


idgserpro:
@hbrunn I think you can now remove this:

https://github.com/buildout/buildout/blob/16ea8e6a8dccde21b853ca2ce5cf5c9c9e5b59de/src/zc/buildout/testing.py#L597-L609

hbrunn:
without this, all 2.7 builds fail, see ie #533 - but it seems we need to set the flag on other places too to fix ie https://github.com/buildout/buildout/pull/534/checks?check_run_id=1128529685#step:7:54

idgserpro:
but in #533 it doesn't have `PIP_NO_PYTHON_VERSION_WARNING`

idgserpro:
Maybe put here?

https://github.com/buildout/buildout/blob/911dfb3f6eddb73d0884062fa5ba5320f8576361/.github/workflows/python-runners.yml

hbrunn:
@idgserpro thanks for the hint, added it, but this doesn't seem the issue:

https://github.com/buildout/buildout/pull/534/checks?check_run_id=1129346525#step:3:590

idgserpro:
@hbrunn At least it's one less problem :smile: 

Still on the issue of warnings, it seems that it will be necessary to add PIP_NO_PYTHON_VERSION_WARNING to some more files here:

https://github.com/buildout/buildout/tree/master/.github/workflows

hbrunn:
we get the same problem in travis where we're sure to have the variable set: https://travis-ci.org/github/buildout/buildout/jobs/728057034#L709 - as I'm not acquainted to the tests, is there a place where the tests create a new environment?

idgserpro:
Travis uses a docker. You can pass `PIP_NO_PYTHON_VERSION_WARNING` to docker here:

https://github.com/buildout/buildout/blob/master/.github/workflows/Dockerfile

idgserpro:
@hbrunn I remembered that zc.recipe.testrunner recipe has the environment option. For example:

https://github.com/zopefoundation/z3c.form/blob/b4bd899bd71e2eeb6e927059bb8f947608ae6b98/buildout.cfg#L13-L20

You can try to use this here:

https://github.com/buildout/buildout/blob/a82a8025d701a4efebb3a3345d91bad8a6bc9658/buildout.cfg#L21

If it works, it is the only place where you will need to set PIP_NO_PYTHON_VERSION_WARNING

idgserpro:
@hbrunn the environment option worked. You can remove PIP_NO_PYTHON_VERSION_WARNING from other places. You can also remove:

https://github.com/buildout/buildout/blob/16ea8e6a8dccde21b853ca2ce5cf5c9c9e5b59de/src/zc/buildout/testing.py#L597-L609

And their references in:

https://github.com/buildout/buildout/blob/86818e5b3e0745a28ac8b3842419744711437de2/src/zc/buildout/tests.py

hbrunn:
sure, I'll drop the commits before, and do the removal. Any idea about the encoding issue?

idgserpro:
I have no idea now. But this problem can be fix in another PR.

src/zc/buildout/tests.py
```diff
@@ -3445,6 +3445,7 @@ def add_source_dist(test, version=1.4):
         shutil.rmtree(tmp)
 
 def easy_install_SetUp(test):
+    os.environ['PIP_NO_PYTHON_VERSION_WARNING'] = '1'
```
mgedmin, line 3448: Given this, is it necessary to also sprinkle PIP_NO_PYTHON_VERSION_WARNING in all the dockerfiles/travis/github workflows/buildouts?

(And if this is insufficient, since there are many test suites using different setup functions, what about moving it into global scope so it happens as soon as the test runner imports tests.py?)

.github/workflows/Dockerfile
```diff
@@ -2,6 +2,8 @@ FROM centos as centos-python
 RUN yum install -y make gcc openssl-devel libffi-devel sqlite-devel
 ARG PYTHON_VER
 ENV LC_CTYPE="en_US.UTF-8"
+ENV LANG C.UTF-8
```
mgedmin, line 5: I'm curious about this one.  What does it do?  I mean, there already was a LC_CTYPE=en_US.UTF-8, so locale charset was UTF-8. Did you find some translated messages messing things up?

mgedmin (approved):
Not a great fan of sprinkling PIP_NO_PYTHON_VERSION_WARNING everywhere rather than once in `tests.py`, so -0.

Even less of a fan of having failing tests in git master, which changes that to +0.5.

The strange unicode traceback is probably https://github.com/zopefoundation/zope.testrunner/issues/18 and it hides the real error, annoying.

(I tried to re-run the cancelled MacOS job but the only option GitHub Actions gave me was "re-run all jobs", so now it's wasting CPU cycles and making polar bears sad.  I hate GitHub Actions.)

src/zc/buildout/tests.py
hbrunn: we concluded yesterday that the last [commit](https://github.com/buildout/buildout/pull/534/commits/d245de499cea38818c5397d994d1577124b858f5) is the actual fix, I'll remove everything else

.github/workflows/Dockerfile
hbrunn: this was just a hunch trying to fix the [last issue](https://travis-ci.org/github/buildout/buildout/jobs/728115069#L712) which looks to me like the terminal here isn't set to unicode. If you have a better idea I'll be happy to try it

mgedmin: Actually, it was a good try!  The `centos` docker image doesn't support `en_US.UTF-8`, but `C.UTF-8` works.

The only problem is LC_CTYPE overrides LANG, so your workaround wasn't actually being used.  Try setting LC_ALL instead of LANG and LC_CTYPE.

hbrunn:
@idgserpro @mgedmin thanks for your input, now we achieved greenness

mgedmin (approved):
:100:

icemac:
Is this repository under the zopefoundation policy so signing a contributor agreement is required or should one of the members of the buildout organization merge this PR?

idgserpro:
@icemac This was asked in https://github.com/buildout/buildout/issues/444 as well. Maybe with the definitive answer here we can close it.

gotcha:
@hbrunn Thanks for working on this.

 Do you plan to sign the zope foundation contributor agreement ? Or do you want me to redo your proposed changes ?

.github/workflows/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM centos as centos-python
 RUN yum install -y make gcc openssl-devel libffi-devel sqlite-devel
 ARG PYTHON_VER
-ENV LC_CTYPE="en_US.UTF-8"
+ENV LC_ALL C.UTF-8
```
gotcha, line 4: @hbrunn I never really understand what I am doing when I touch those LC_ variables. Would you mind explaining why this is better or could you point me to a good explanation ?

hbrunn: I find https://www.gnu.org/software/libc/manual/html_node/Locale-Categories.html quite illuminating

hbrunn:
@gotcha it was not clear to me we're waiting for an action from my side, I thought it was still being discussed if the signature is necessary. If it it speeds up things, feel free to pick up the code, or point me where to sign.

icemac:
@hbrunn The contributor agreement is at: https://www.zope.org/developer/becoming-a-committer.html

gotcha:
@hbrunn If you feel like signing the agreement, it's maybe the easiest. If you do not feel like it, can you let us know ?

gotcha:
@icemac asked about the Contributor Agreement on community.plone.org.

@jensens said he thought it was not needed. I contacted @jimfulton that confirmed.

@hbrunn You can thus forget my uninformed questions.

.github/workflows/Dockerfile
gotcha: @hbrunn Thanks !

Issue #515, opened 2020-05-19
mr.developer issue with zc.buildout 3.0.0a1


ale-rt:
I tested zc.buildout with an existent virtualenv:
```
$ ./bin/pip3.8 install -IU zc.buildout==3.0.0a1
Looking in links: file:///home/ale/.pip-wheelhouse
Collecting zc.buildout==3.0.0a1
  Using cached zc.buildout-3.0.0a1-py2.py3-none-any.whl (158 kB)
Collecting pip
  Downloading pip-20.1.1-py2.py3-none-any.whl (1.5 MB)
     |████████████████████████████████| 1.5 MB 1.1 MB/s 
Collecting setuptools>=8.0
  Using cached setuptools-46.4.0-py3-none-any.whl (583 kB)
Collecting wheel
  Using cached wheel-0.34.2-py2.py3-none-any.whl (26 kB)
Installing collected packages: pip, setuptools, wheel, zc.buildout
Successfully installed pip-20.1.1 setuptools-46.4.0 wheel-0.34.2 zc.buildout-3.0.0a1
```
Then running buildout:
```
./bin/buildout
Setting socket time out to 60 seconds.
mr.developer: Queued ... for checkout.
Upgraded:
  pip version 19.2.3,
  wheel version 0.33.6;
Restarting.
Generated script '/home/ale/Code/plone/projects/pi52-py3/bin/buildout'.
Requires-Python support missing. 

Traceback (most recent call last):
  File "/home/ale/Code/plone/projects/pi52-py3/lib/python3.8/site-packages/zc/buildout/patches.py", line 69, in patch_PackageIndex
    from pip._internal.index.collector import HTMLPage
ModuleNotFoundError: No module named 'pip._internal.index.collector'; 'pip._internal.index' is not a package
Setting socket time out to 60 seconds.
mr.developer: Queued '...' for checkout.
While:
  Installing.
  Loading extensions.

An internal error occurred due to a bug in either zc.buildout or in a
recipe being used:
Traceback (most recent call last):
  File "/home/ale/Code/plone/projects/pi52-py3/lib/python3.8/site-packages/zc/buildout/buildout.py", line 2182, in main
    getattr(buildout, command)(args)
  File "/home/ale/Code/plone/projects/pi52-py3/lib/python3.8/site-packages/zc/buildout/buildout.py", line 660, in install
    self._load_extensions()
[...]

PR #536, opened 2020-09-29
Problem: mr.developer broken by CLI arg to avoid endless loop

gotcha:
Solution: use environment variable instead

See #515

### DIFF
--- a/src/zc/buildout/buildout.py
+++ b/src/zc/buildout/buildout.py
@@ -1099,7 +1099,7 @@ class Buildout(DictMixin):
         # If they do, do the upgrade and restart the buildout process.
         __doing__ = 'Checking for upgrades.'
 
-        if self['buildout'].get('restart-after-upgrade', '') == 'true':
+        if 'BUILDOUT_RESTART_AFTER_UPGRADE' in os.environ:
             return
 
         if not self.newest:
@@ -1172,11 +1172,11 @@ class Buildout(DictMixin):
 
         # Restart
         args = sys.argv[:]
-        args.insert(1, '--restart-after-upgrade')
         if not __debug__:
             args.insert(0, '-O')
         args.insert(0, sys.executable)
-        sys.exit(subprocess.call(args))
+        env=dict(os.environ, BUILDOUT_RESTART_AFTER_UPGRADE='1')
+        sys.exit(subprocess.call(args, env=env))
 
     def _load_extensions(self):
         __doing__ = 'Loading extensions.'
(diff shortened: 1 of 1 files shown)


src/zc/buildout/buildout.py
```diff
@@ -1172,11 +1172,11 @@ def _maybe_upgrade(self):
 
         # Restart
         args = sys.argv[:]
-        args.insert(1, '--restart-after-upgrade')
         if not __debug__:
             args.insert(0, '-O')
         args.insert(0, sys.executable)
-        sys.exit(subprocess.call(args))
+        env = {'BUILDOUT_RESTART_AFTER_UPGRADE': '1'}
```
perrinjerome, line 1178: Shouldn't this be something like 
```python
        env = dict(os.environ, BUILDOUT_RESTART_AFTER_UPGRADE='1')
```
to also preserve environment from current process variables ?

src/zc/buildout/buildout.py
```diff
@@ -2161,8 +2161,6 @@ def main(args=None):
                     _help()
                 elif orig_op == '--version':
                     _version()
-                elif orig_op == '--restart-after-upgrade':
-                    options.append(('buildout', 'restart-after-upgrade', "true"))
```
mgedmin, line 2165: I'm curious as to what happens if a zc.buildout before this patch upgrades itself to a version after this patch.  Will the old buildout add --restart-after-upgrade to the command line, causing the new buildout to fail with an unknown command line argument error?

mgedmin:
Do you have a link to a mr.developer bug, or a traceback, or something for us to see how mr. developer fails?

mgedmin:
The CI failures I looked at seem to be caused by Python version deprecation warnings.  Didn't we have a fix for that in some other PR?  Perhaps, since the fix consisted of putting an environment variable asking pip not to print those warnings, the fix may have broken for the reason @perrinjerome noticed: that environment variable is now lost when you replace the entire environment.

mgedmin:
> Didn't we have a fix for that in some other PR? 

Ah, not yet: #534.  Still, the CI fix will need the environment to be preserved.

src/zc/buildout/buildout.py
gotcha: @perrinjerome Right, I missed that.

src/zc/buildout/buildout.py
gotcha: @mgedmin  ``--restart-after-upgrade`` only exist in 3.0.0a1 and 3.0.0a2

I feel we do not need to cover those alpha versions. What is your take on that ?

gotcha:
I added the link to #515 in the description.

src/zc/buildout/buildout.py
mgedmin: Good point!  I forgot when it was introduced; seemed like ages ago.

icemac:
Is there still something needed to be done before merging this PR?

ZachArani:
> 
> 
> Is there still something needed to be done before merging this PR?

I second this. Just found this issue rearing its ugly head in my workflows.

Issue #510, opened 2020-05-18
Requires-Python in 3.0.0a1 does double parsing of index pages.


gotcha:
Should be removed before final release.

PR #547, opened 2021-02-17
Fix bootstrapping for Python 2.7

### LINKED ISSUE #546: bootstrap process is broken for Python 2.7
The problem seems to be that when you download https://bootstrap.pypa.io/get-pip.py and run `python2.7 get-pip.py`, this results in

```
python ~/Downloads/get-pip.py 
  File "/home/jugmac00/Downloads/get-pip.py", line 24244, in <module>
    main()
  File "/home/jugmac00/Downloads/get-pip.py", line 199, in main
    bootstrap(tmpdir=tmpdir)
  File "/home/jugmac00/Downloads/get-pip.py", line 82, in bootstrap
    from pip._internal.cli.main import main as pip_entry_point
  File "/tmp/tmpApPNaE/pip.zip/pip/_internal/cli/main.py", line 60
    sys.stderr.write(f"ERROR: {exc}")
                                   ^
SyntaxError: invalid syntax
```

_Originally posted by @jugmac00 in https://github.com/buildout/buildout/issues/545#issuecomment-780755469_

jugmac00:
The bootstrap process uses `get-pip.py`, which in its latest version,
tries to bootstrap a Python 3 only version of pip.

We need to pass in a separate URL for a Python 2.7 compatible version of get-pip.py.

This should fix #546

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,16 +1,12 @@
 Change History
 **************
 
-2.13.x (unreleased)
-===================
-
-- Fix bootstrapping for python27
-
-
 3.0.0a3 (unreleased)
 ====================
 
 - Support python37, python38 and python39 in conditional section expressions
+
+- Fix bootstrapping for python27 and python35
 
 
 3.0.0a2 (2020-05-25)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,12 @@
 Change History
 **************
 
+2.13.x (unreleased)
+===================
+
+- Fix bootstrapping for python27
+
+
 3.0.0a3 (unreleased)
 ====================
 
--- a/dev.py
+++ b/dev.py
@@ -52,8 +52,11 @@ def install_pip():
     tmp = tempfile.mkdtemp(prefix='buildout-dev-')
     try:
         get_pip = os.path.join(tmp, 'get-pip.py')
+        GET_PIP_URL = 'https://bootstrap.pypa.io/get-pip.py'
+        if sys.version_info < (3, ):
+            GET_PIP_URL = 'https://bootstrap.pypa.io/2.7/get-pip.py'
         with open(get_pip, 'wb') as f:
-           f.write(urlopen('https://bootstrap.pypa.io/get-pip.py').read())
+            f.write(urlopen(GET_PIP_URL).read())
 
         sys.stdout.flush()
         if subprocess.call([sys.executable, get_pip]):
--- a/dev.py
+++ b/dev.py
@@ -86,7 +86,7 @@ def check_upgrade(package):
         output = subprocess.check_output(
             [sys.executable] + ['-m', 'pip', 'install', '--upgrade', package],
         )
-        was_up_to_date = b"up-to-date" in output
+        was_up_to_date = b"up-to-date" in output or b"already satisfied" in output
         if not was_up_to_date:
             print(output.decode('utf8'))
         return not was_up_to_date
--- a/dev.py
+++ b/dev.py
@@ -59,8 +59,7 @@ def install_pip():
         else:
             GET_PIP_URL = 'https://bootstrap.pypa.io/get-pip.py'
         with open(get_pip, 'wb') as f:
-            with urlopen(GET_PIP_URL) as r:
-                f.write(r.read())
+            f.write(urlopen(GET_PIP_URL).read())
 
         sys.stdout.flush()
         if subprocess.call([sys.executable, get_pip]):
--- a/dev.py
+++ b/dev.py
@@ -52,11 +52,15 @@ def install_pip():
     tmp = tempfile.mkdtemp(prefix='buildout-dev-')
     try:
         get_pip = os.path.join(tmp, 'get-pip.py')
-        GET_PIP_URL = 'https://bootstrap.pypa.io/get-pip.py'
         if sys.version_info < (3, ):
             GET_PIP_URL = 'https://bootstrap.pypa.io/2.7/get-pip.py'
+        elif (sys.version_info.major, sys.version_info.minor) == (3, 5):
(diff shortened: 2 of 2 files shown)


jugmac00:
Test failures where due to the bootstrap process did not account for pip's changed message for "nothing updated" which has changed from pip 20 to 21.

jugmac00:
pinging @mgedmin - you seem to the only active "people" of this project ( https://github.com/orgs/buildout/people )

CHANGES.rst
```diff
@@ -1,6 +1,12 @@
 Change History
 **************
 
+2.13.x (unreleased)
```
jamadden, line 4: This code is targeting master, so this should be 3.0.0a4, shouldn't it? 

If we want a 2.x release (IMO, we don't—3.0 is the way forward) this would need to be cherry-picked back to a new branch.

jamadden (approved):
LGTM, with the question about version targeting.

CHANGES.rst
jugmac00: I had a look around the branches and the developer documentation and I found no clue how the development will go forward concerning the version 2 and the version 3 development line.

Just last week @icemac released a new 2.13 version. @icemac Can you give us a hint? On which branch did you develop?

The version 3 line seems to be stalled for almost a complete year now, and also there seem to be at least a couple of reported problems, see #510 and #515

Also, I am not sure whether the alpha releases are already well tested in the field, e.g. by Plone and maybe some Zope projects - usually a lot of edge cases only surface when tested against real projects.

Also, will version 3 still support Python 2.7 or is this the moment for the cut? I have no clue.

As I use mr. developer and there seems to be a problem in the version 3 (see above #515), I'd be happy if there would be a new 2.13 release - except that would be too much effort.

jamadden: I use buildout 3.0 with a very large Zope 3 project, mr.developer, and a relatively complicated buildout, and haven't run into those specific problems. (Granted, that's just my local dev environment; we're not using it in production yet.)

I think 3.0 should/will support 2.7: 2.7 users need a path forward (and 2.7 is what I've been using 3.0 with).

The problem with pre-releases is that almost nobody tests them, since pip stopped installing them. The only way to get real testing, aside from begging the community, is to release a final version.

dev.py
```diff
@@ -52,8 +52,11 @@ def install_pip():
     tmp = tempfile.mkdtemp(prefix='buildout-dev-')
     try:
         get_pip = os.path.join(tmp, 'get-pip.py')
+        GET_PIP_URL = 'https://bootstrap.pypa.io/get-pip.py'
+        if sys.version_info < (3, ):
+            GET_PIP_URL = 'https://bootstrap.pypa.io/2.7/get-pip.py'
```
mgedmin, line 57: I think we also need to special-case Python 3.5 in a similar way.

dev.py
```diff
@@ -52,8 +52,11 @@ def install_pip():
...
         with open(get_pip, 'wb') as f:
-           f.write(urlopen('https://bootstrap.pypa.io/get-pip.py').read())
+            f.write(urlopen(GET_PIP_URL).read())
```
mgedmin, line 59: Minor:

```suggestion
            with urlopen(GET_PIP_URL) as r:
                f.write(r.read())
```

CHANGES.rst
```diff
@@ -1,6 +1,12 @@
...
+===================
+
+- Fix bootstrapping for python27
+
+
 3.0.0a3 (unreleased)
```
mgedmin, line 10: I object to non-monotonic version numbers in the changelog.  If we want to backport this to the 2.x release, we probably want to do that n a branch with a changelog that doesn't have all these 3.0.0 alpha versions.

mgedmin (commented):
Thing is, dev.py is only used for the development of buildout itself.  End-users of zc.buildout use `pip install zc.buildout && buildout bootstrap`, which doesn't execute any code of `dev.py`.  I can see this PR affecting CI.  I don't quite see the need for a 2.x release that won't affect any end users.

dev.py
jugmac00: Thanks! I learned something new.

dev.py
jugmac00: Now that you say it... I just wonder why both CI and my local test run did not show any problems. But let's do it anyway. Looks like we have to support 3.5 until the end of 2022 ( https://www.zope.org/developer/roadmap.html ). Updated with next commit.

CHANGES.rst
jugmac00: Very well. Makes sense. Will be fixed with the next commit.

CHANGES.rst
jugmac00: Thanks for your time - will be fixed with the next commit.

jugmac00:
@mgedmin and @jamadden Thank you for the reviews. I have applied all changes.

Unlike for the zopefoundation repos, here I have no right to merge.

Please do the merge if everything is ok and/or accept me as a developer for this project.

Cheers!

dev.py
jugmac00: Uhm, this made boom on Python 2.7 - I just pushed the old version.

mgedmin: Ahh oops.

CHANGES.rst
icemac: The last `zc.buildout` release is about a year old: https://pypi.org/project/zc.buildout/2.13.3/
There is the branch https://github.com/buildout/buildout/tree/2.x – I think the 2.13.x development shout happen there. – but probably we also need your changes on `master`, too.

jugmac00: Thanks for the correction - seems I still live in 2020 :-D Oh this pandemic!

master was def. broken, so imho yes, we need the changes here.

I will have a look at the 2.x branch (where dev.py looks completely different), once this got merged and/or I got the right to merge myself.

@icemac Maybe you can enable me to get forward here.

PR #549, opened 2021-02-20
update setuptools requirements

### LINKED ISSUE #543: setuptools 52.0.0 breaks buildout 2.13.3 installs
In https://github.com/pypa/setuptools/commit/2885ca26494e6f555fae85f8f9983c2361c93829 `setuptools.command.easy_install.main` was removed.
It seems to be needed for a successful installation using buildout 2.13.3, so CI of various packages breaks, e. g.

* https://gitlab.com/gocept/union.cms/products.orderedbtreefolder/-/jobs/998458181
* https://github.com/zopefoundation/Products.ZCatalog/runs/1801048436

jugmac00:
setuptools is now pinned to <52.

setuptools version 52 broke support for easy_install
https://github.com/pypa/setuptools/blob/main/CHANGES.rst#breaking-changes-1

info below is now invalid
```
setuptools>=42.0.1 is necessary so buildout can download e.g. latest
versions of `cryptography`, possibly related to
https://github.com/pyca/cryptography/blob/main/CHANGELOG.rst#34---2021-02-07
also see
https://github.com/pypa/setuptools/blob/main/CHANGES.rst#v4201
```

This fixes #543

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,9 @@ Change History
 2.13.4 (unreleased)
 ===================
 
-- Nothing changed yet.
+- Pin `setuptools` < 52 as setuptools removed support for `easy_install` in
+  newer versions.
+  (`#543 <https://github.com/buildout/buildout/issues/543>`_)
 
 
 2.13.3 (2020-02-11)
--- a/setup.py
+++ b/setup.py
@@ -45,9 +45,7 @@ setup(
     packages = ['zc', 'zc.buildout'],
     package_dir = {'': 'src'},
     namespace_packages = ['zc'],
-    install_requires = [
-        'setuptools>=8.0',
-    ],
+    install_requires = ['setuptools>=8.0,<52',],
     include_package_data = True,
     entry_points = entry_points,
     extras_require = dict(


jugmac00:
Hm.... 

```
ez_setup.py is deprecated and when using it setuptools will be pinned to 33.1.1 since it's the last version that supports setuptools self upgrade/installation, check https://github.com/pypa/setuptools/issues/581 for more info; use pip to install setuptools
```

This does not fit well with `setuptools is now pinned to >=42.0.1 and <52.`.

On the other hand, `ez_setup.py` is only used for the dev setup.

Anyway, if version 3 alpha is in a working state, we should push forward to use it and fix possible edge case, instead of putting too much effort into the 2.13 line.

mgedmin:
> On the other hand, ez_setup.py is only used for the dev setup.

I hope I'm not misrepresenting anything, but Jim Fulton's goal for zc.buildout was to make it usable for clean non-system Python installations that don't have any 3rd-party packages (like setuptools) installed.  That was the job of bootstrap.py and it's dev-only equivalent dev.py.  (Unfortunately I don't remember what the exact difference between these two scripts was.)

Things in the ecosystem have changed sufficiently that we dropped bootstrap.py and now expect the users to already have a working `pip` that they can use to install zc.buildout.  I think it may be time to consider retiring `dev.py` as well.

(Of course the devil is in the details.)

icemac:
I do not feel competent to review this PR.

StefanRijnhart (approved):
I can confirm setuptools needs to be pinned < 52.

idgserpro:
Is `2.x` branch planning to still support Python 2? The last setuptools release that supports Python 2 is 44.1.1 (check https://setuptools.readthedocs.io/en/latest/history.html#v45-0-0)

jugmac00:
> Is `2.x` branch planning to still support Python 2, as [set in classifiers](https://github.com/buildout/buildout/blob/6e4eb0d9e8d21f0da8efca8e30d0c7631cc23c90/setup.py#L62)? The last setuptools release that supports Python 2 is 44.1.1 (check https://setuptools.readthedocs.io/en/latest/history.html#v45-0-0)

As long as `setuptools` uses `python_requires` we should be fine I guess.

mauritsvanrees:
The pin of less than 52 seems good.

Minimum 42.0.1 seems reasonable. But is it really needed? Most packages can be installed with much older setuptools versions. If someone wants `cryptography` 3.4+ and gets an error installing it, then it is their responsibility to use a newer setuptools.
I am just saying: this minimum pin may make life easier for a few, but may make it harder for others who need or want to keep using an older setuptools for whatever reason. For starters, 42.0.0 has major breaking changes.

jugmac00:
@mauritsvanrees Thanks for your review - I updated the version pin accordingly.

I still think it is a frustrating experience to install a package which leads to a broken environment, but on the other hand, I can fix it by changing the version in my versions.cfg.

On the other hand, if we followed my proposal, and for some "reasons" somebody needs e.g. setuptools 9, there would be no way to fix it - except pinning down buildout...

So, let's get this merged and look forward to buildout 3!

mauritsvanrees (approved):
LGTM now.

PR #552, opened 2021-03-03
Try to fix tests in GHA: pin to a Python 2.7 compatible version.

icemac:


icemac:
What a pity TravisCI is broken. @gotcha do we still need the tests on TravisCI?

icemac:
It seems `['setup.py', '-q', 'develop', '-m', '-x', '-d', 'develop-eggs']` in `dev.py` installs `setuptools` version 54 on TravisCI breaking the build. See https://travis-ci.org/github/buildout/buildout/jobs/761403095

I do not know what we can do against this misbehavior as `setuptools` is already pinned to `<52` in `setup.py`.

dev.py
```diff
@@ -54,7 +54,7 @@
 setuptools_path = os.path.dirname(os.path.dirname(setuptools.__file__))
 
 ######################################################################
-# Install buildout
+print("Installing buildout ...")
```
jamadden, line 57: Is this here on purpose or leftover from debugging?

jamadden (approved):
The GHA changes seem fine to me.

dev.py
icemac: It is mostly a leftover from debugging. I thought it could be helpful in the future, too.

.github/workflows/scripts-versions-2.7.cfg
```diff
@@ -1,2 +1,5 @@
 [versions]
+configparser = 4.0.2
+cryptography = 3.3.2
+keyring = 18.0.1
 more-itertools = 5.0.0
```
mgedmin, line 5: I'm mildly wondering if it would be a better idea to use `<` pins, e.g. `keyring = <19`, in case a bugfix 18.0.2 ever comes out?

I'm probably overthinking things like I usually do.

mgedmin (approved):
LGTM.

Maybe let's `rm .travis.yml`?  Even if it works (well, fails) today, we're going to run out of free minutes and things wills top running eventually.

.github/workflows/scripts-versions-2.7.cfg
icemac: This is a good idea I was jus following the example provided for `more-itertools`.

icemac:
> Maybe let's `rm .travis.yml`? Even if it works (well, fails) today, we're going to run out of free minutes and things wills top running eventually.

We are still running on travis-ci.org where the minutes used do not seem to count against the free minutes, see https://travis-ci.com/organizations/buildout/plan/usage … but travis-ci.org is really slow and they are planning to shut it down for some months now.

.github/workflows/scripts-versions-2.7.cfg
icemac: Done in https://github.com/buildout/buildout/pull/552/commits/538e2d176fcc5f73ebc63e6272103e1b88b6ed8d

gotcha:
@icemac @jamadden @jugmac00 Is it absolutely needed to have passing CI for a new 2.13 release ? IIRC, fixing `dev.py` for master took me a lot of time last spring. A middle ground could be to cut a 2.13.4b1 beta release before a final 2.13.4. This way, people could test the new `setuptools` pinning without breaking all setups.

BTW, I have just cut a new release for 3.0: https://pypi.org/project/zc.buildout/3.0.0b1/ We could encourage people to test 3.0.0b1 at the same time. We might realize there is no need for 2.13 anymore.

icemac:
@gotcha Thank you for commenting this issue. I am going to remove the `.travis.yml` as there is no easy way to fix the tests over there. I do not expect problems with a new 2.13 release so I'd create a 2.13.4 release right away.

icemac:
Thank you all for reviewing this PR. 😃

gotcha:
@icemac Thanks for the release.

PR #553, opened 2021-03-05
Requirement parse error

### LINKED ISSUE #550: buildout 3.x: requirements with extras fail with version constraints
This cfg fails to install with zc.buildout 3.x but works with 2.x:

```
[buildout]
parts = checkversions

[checkversions]
recipe = zc.recipe.egg
eggs = z3c.checkversions [buildout]
```

I think [``_constrained_requirement``](https://github.com/buildout/buildout/blob/master/src/zc/buildout/easy_install.py#L1620) needs to remove the extra part before calling ``pkg_resources.Requirement.parse``.
Currently the ``requirement`` passed in is ``Requirement.parse('zc.buildout; extra == "buildout"')`` with no ``specs`` and then it just bolts on the constraint ``>=3.0.0a2`` which results in the string ``'zc.buildout; extra == "buildout" >=3.0.0a2'`` to be passed to ``pkg_resources.Requirement.parse``.

Looking a bit further, this might work as a fix: ``pkg_resources.Requirement.parse(requirement.name + ' ' + constraint)``, but I'm not sure if ``requirement`` is always a ``Requirement`` object.

Besides this issue I can report that a fairly complex buildout config works just fine with buildout 3.x. It has custom buildout extensions and lots of other non standard things going on.

gotcha:
Fix #550

### DIFF
--- a/dev.py
+++ b/dev.py
@@ -96,10 +96,26 @@ def check_upgrade(package):
     except subprocess.CalledProcessError:
         raise RuntimeError("Upgrade %s failed." % package)
 
+
+def show(package):
+    try:
+        sys.stdout.flush()
+        output = subprocess.check_output(
+            [sys.executable, '-m', 'pip', 'show', package],
+        )
+        for line in output.splitlines():
+            if line.startswith(b'Name') or line.startswith(b'Version'):
+                print(line.decode('utf8'))
+    except subprocess.CalledProcessError:
+        raise RuntimeError("Upgrade %s failed." % package)
+
+
 need_restart = False
 for package in ['pip', 'setuptools', 'wheel']:
     did_upgrade = check_upgrade(package)
+    show(package)
     need_restart = need_restart or did_upgrade
+
 if need_restart:
     print("Restart")
     sys.stdout.flush()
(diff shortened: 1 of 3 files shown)


dev.py
```diff
@@ -96,10 +96,26 @@ def check_upgrade(package):
...
 
+
+def show(package):
+    try:
+        sys.stdout.flush()
+        output = subprocess.check_output(
+            [sys.executable] + ['-m', 'pip', 'show', package],
+        )
+        for line in output.splitlines():
+           if line.startswith(b'Name') or line.startswith(b'Version'):
```
mgedmin, line 107: This looks like a 3-space indent to me?

dev.py
```diff
@@ -96,10 +96,26 @@ def check_upgrade(package):
     except subprocess.CalledProcessError:
         raise RuntimeError("Upgrade %s failed." % package)
 
+
+def show(package):
+    try:
+        sys.stdout.flush()
+        output = subprocess.check_output(
+            [sys.executable] + ['-m', 'pip', 'show', package],
```
mgedmin, line 104: Why not 
```suggestion
            [sys.executable, '-m', 'pip', 'show', package],
```
?

mgedmin (commented):
I'd say LGTM (other than the suspicious indentation), but TBH I don't understand the logic.

dev.py
gotcha: Thx

dev.py
gotcha: :-) copy paste

Issue #556, opened 2021-03-08
Cannot install FormEncode 1.3.1


gotcha:
`FileNotFoundError` due to `os.rename` into a missing directory.

@jensens  found it via https://github.com/plone/buildout.coredev/runs/2057051639

jugmac00:
I tried to reproduce it and it worked on my machine.

Can you tell me what I did wrong?

```
❯ bat buildout.cfg 
───────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: buildout.cfg
───────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ [buildout]
   2   │ parts = one
   3   │ 
   4   │ [one]
   5   │ recipe = zc.recipe.egg
   6   │ eggs = formencode
   7   │ 
   8   │ [versions]
   9   │ formencode = 1.2.1
───────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

/tmp/buildout-test via 🐍 v3.6.9 (venv)
❯ buildout
Installing one.
Getting distribution for 'formencode==1.2.1'.
Got FormEncode 1.2.1.
```

gotcha:
@jugmac00 `FormEncode 1.3.1` 😉

jugmac00:
Oh my 😔

Thank you!

gotcha:
Fixed in #557

PR #557, opened 2021-03-08
Problem: egg installation fails with FileNotFoundError

gotcha:
Solution: create missing dir before `os.rename`

See #556


icemac (approved):
LGTM.

PR #562, opened 2021-03-22
WIP mypy-fication

gotcha:
I would like the opinion of others before merging this.

Using `mypy` did point to potential problems.
OTOH, it is quite wordy.

What's your take on this ?

### DIFF
--- a/dev.py
+++ b/dev.py
@@ -124,6 +124,26 @@ if need_restart:
     )
     sys.exit(return_code)
 ######################################################################
+def install_pyannotate():
+    print('')
+    print('Install pyannotate')
+    print('')
+    bin_pip = os.path.join('bin', 'pip')
+    if subprocess.call(
+        [sys.executable] +
+        ['-m', 'pip', 'install', 'pyannotate'],
+        ):
+        raise RuntimeError("pyannotate install failed.")
+
+
+if "PYANNOTATE" in os.environ:
+    try:
+        import pyannotate
+    except ImportError:
+        install_pyannotate()
+
+######################################################################
+######################################################################
 print('')
 print('Install buildout')
 print('')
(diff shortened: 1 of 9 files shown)


jugmac00 (approved):
I use type annotations in all new projects, and I also added them to three or four legacy projects.

While adding type annotations is more fun when you do not have to support Python 2.7 and 3.5, it still could help a lot when migrating from Python 2 to 3, but well... buildout is already Python 3 compatible.

Anyway.. I think type annotations are the future.

I like the gradual approach of Python. So, everytime we change something, we could add type annotations. No need to type annotate the complete code base with one PR.

icemac:
Sorry, I do not have the energy to read through the complete diff. In general I like the idea to annotate the code with types but I personally haven't used it too much yet.

What about the problems DeepSource reported? (See https://deepsource.io/gh/buildout/buildout/run/a42a6964-9e6c-4e1a-8cf8-5e94c1dd1b6a/python/) On the last build of `master` DeepSource is green.

mauritsvanrees:
I am not using mypy yet, so I can't comment on details.
I am not against. It may make sense, and for me it would be an example project of how it is done.

I would need to read up on how mypy works, and who sees errors or warnings during development, or during production use.
Does anything change for recipes or extensions? Can they run into errors or warning with this?

I can imagine it is more useful for new code or completely new projects. Then again, this may help in cleaning up old code for corner cases where recipes pass in arguments of unexpected types.

gotcha:
@icemac
- I am not asking approval about every line details but rather about your position around typing and `mypy`.
- I still need to dig what is happening with DeepSource, at first look, I did not understand what happened.

@mauritsvanrees 
- `mypy` is not a runtime tool. It is used during development to supposedly help finding places in the code that do not fit with the types declaration. IOW, it would not change runtime comportment. And recipes would not be impacted.
- It might lead to rethink some method calls. But in that case, there should be a compatibility layer kept for recipes with proper deprecation warnings. I would not do this as a first step anyway.

mauritsvanrees:
I close this because there was no activity since 4 years, and the master branch has changed a lot, so this PR would likely need to be restarted from scratch.  Feel free to reopen if you want to pick this up again.

In general: yes, type annotations may help.

PR #579, opened 2021-09-29
Fix #574 wrongly matched eggs when multiple python versions

### LINKED ISSUE #574: With buildout 3.x shared eggs are wrongly matched
I haven't found a simple reproduction case yet, but I often see wrong eggs being matched from the eggs cache when using multiple Python versions. I saw it when switching the used Python version when there are eggs from the previous version left in the eggs directory and in CI where multiple Python versions run and use a shared cache. The sys.path in scripts has eggs from the wrong Python version in those cases.

gotcha:


mauritsvanrees:
I don't really know how to test buildout on my machine, but the main commit looks okay. 
I agree with your gripe about that this used to be a nice hint that an egg has C extensions. But if there is no better way, then it should be fine.

I guess an alternative for finding packages with C extensions, is to do `find ~/eggs -name '*.so'`.

icemac:
Having a hint about C extensions in the eggs path might be nice but it should not prevent this PR from being merged as there are other ways to find out about the C extensions as @mauritsvanrees pointed out.

icemac (approved):
LGTM from reading through the code.

Issue #586, opened 2021-11-24
buildout 3 fails with a traceback when `pip install` fails.


gotcha:

gotcha:
Fixed and released in https://pypi.org/project/zc.buildout/3.0.0b4/

Issue #585, opened 2021-11-23
extends-cache fails when it includes ${section:variable} 


gotcha:
Which is normal but the failure is not explicit at all.

There should be a check and an error message explaining that `${section:variable}` are not allowed in `extends-cache`.

gotcha:
Fixed and released in https://pypi.org/project/zc.buildout/3.0.0b4/

Issue #582, opened 2021-10-12
missing top_level.txt causes error in 3.0.0b3


goschtl:
when i try to install orjson in an zc.recipe.egg i get the following error:

Getting distribution for 'orjson'.
While:
  Installing app.
  Getting distribution for 'orjson'.

An internal error occurred due to a bug in either zc.buildout or in a
recipe being used:
Traceback (most recent call last):
  File "/Users/ck/work/reha/tmp/project/lib/python3.8/site-packages/zc/buildout/buildout.py", line 2222, in main
    getattr(buildout, command)(args)
  File "/Users/ck/work/reha/tmp/project/lib/python3.8/site-packages/zc/buildout/buildout.py", line 834, in install
    installed_files = self[part]._call(recipe.install)
  File "/Users/ck/work/reha/tmp/project/lib/python3.8/site-packages/zc/buildout/buildout.py", line 1630, in _call
    return f()
  File "/Users/ck/work/reha/tmp/project/p6/eggs/zc.recipe.egg-2.0.7-py3.8-macosx-11.5-x86_64.egg/zc/recipe/egg/egg.py", line 227, in install
    reqs, ws = self.working_set()
  File "/Users/ck/work/reha/tmp/project/p6/eggs/zc.recipe.egg-2.0.7-py3.8-macosx-11.5-x86_64.egg/zc/recipe/egg/egg.py", line 78, in working_set
    ws = self._working_set(
  File "/Users/ck/work/reha/tmp/project/p6/eggs/zc.recipe.egg-2.0.7-py3.8-macosx-11.5-x86_64.egg/zc/recipe/egg/egg.py", line 161, in _working_set
    ws = zc.buildout.easy_install.install(
  File "/Users/ck/work/reha/tmp/project/lib/python3.8/site-packages/zc/buildout/easy_install.py", line 966, in install
    return installer.install(specs, working_set)
  File "/Users/ck/work/reha/tmp/project/lib/python3.8/site-packages/zc/buildout/easy_install.py", line 739, in install
    for dist in self._get_dist(req, ws):
  File "/Users/ck/work/reha/tmp/project/lib/python3.8/site-packages/zc/buildout/easy_install.py", line 582, in _get_dist
    dists = [_move_to_eggs_dir_and_compile(dist, self._dest)]
  File "/Users/ck/work/reha/tmp/project/lib/python3.8/site-packages/zc/buildout/easy_install.py", line 1895, in _move_to_eggs_dir_and_compile
[...]

Issue #576, opened 2021-08-28
buildout 3.x doesn't forward verbose option to pip


fschulze:
When using ``-v`` with buildout there is no verbose output from pip to see compilation info etc.

gotcha:
Fixed and released in https://pypi.org/project/zc.buildout/3.0.0b4/

PR #587, opened 2021-11-24
Before beta4

### LINKED ISSUE #570: Support running buildout in FIPS enabled environments
Using md5 to create hashes in a FIPS enabled environment fails
```
recipe being used:                                                                                                                                                                                                          
  File "/opt/plone/buildout/lib64/python3.6/site-packages/zc/buildout/buildout.py", line 2174, in main                                                                                                                      
  File "/opt/plone/buildout/lib64/python3.6/site-packages/zc/buildout/buildout.py", line 716, in install                                                                                                                    
    self._compute_part_signatures(install_parts)                                                                                                                                                                            
  File "/opt/plone/buildout/lib64/python3.6/site-packages/zc/buildout/buildout.py", line 962, in _compute_part_signatures                                          
    sig = _dists_sig(pkg_resources.working_set.resolve([req]))                          
  File "/opt/plone/buildout/lib64/python3.6/site-packages/zc/buildout/buildout.py", line 1880, in _dists_sig
    result.append(dist.project_name + '-' + _dir_hash(location))                        
  File "/opt/plone/buildout/lib64/python3.6/site-packages/zc/buildout/buildout.py", line 1839, in _dir_hash
    hash = md5()                                                                                                                                                        
ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS
```

There is a flag that can be used with md5 `usedforsecurity=False` when you want to use md5 for non security related hashing in FIPS enabled environments. However this fails when FIPS is not enabled `TypeError: 'usedforsecurity' is an invalid keyword argument for md5()`

I will submit a PR shortly with a proposed fix

gotcha:
Closes #570 #576 #582 #585 #586


icemac:
One of the problems I spotted in the failing tests is that in Python 3.6 the behavior of `pdb` seems to have changed, see https://github.com/buildout/buildout/runs/4318169989?check_suite_focus=true#step:3:576 (The `(Pdb)` is no longer in front of the `While`.) — maybe another reason to think about dropping the burden of outdated Python versions in favor of having the chance to solve the other problems which are still hard enough.

gotcha:
My gut feeling is that the actual issue is that we are using doctests which are often far too sensitive to output changes.

gotcha:
Merged and released in https://pypi.org/project/zc.buildout/3.0.0b4/

PR #614, opened 2022-09-12
Support python310-315 in conditional section expressions. [2.x]

mauritsvanrees:
If you use `zc.buildout` 2.13.7 to install [Plone 6.0.0b2](https://community.plone.org/t/plone-6-0-0b2-released/15643/2?u=mauritsvanrees) you will get an error:

    NameError: name 'python310' is not defined

See [another example of the same problem](https://github.com/collective/collective.honeypot/pull/8#issuecomment-1244568342).

Easy buildout config to try, on any Python version:

```
[buildout]
parts = test

[test]
recipe = zc.recipe.egg
eggs = build

[test:python310]
recipe = zc.recipe.egg
eggs = requests
```

This PR should fix it by supporting five more Python versions. I guess this should be enough for the foreseen lifetime of buildout 2.x. :-)

I wonder if there are better ways to solve this, but for now this should do the trick.

### DIFF
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -9,7 +9,7 @@ exclude_patterns = ["builds/**"]
 
 [[analyzers]]
 name = "python"
-enabled = false
+enabled = true
 
   [analyzers.meta]
   runtime_version = "3.x.x"
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -1,6 +1,15 @@
-# We do not want this on branch 2.x, so we disable the analyzers..
 version = 1
+
+test_patterns = [
+    "zc.recipe.egg_/src/zc/recipe/egg/tests.py",
+    "src/zc/buildout/tests/**"
+]
+
+exclude_patterns = ["builds/**"]
 
 [[analyzers]]
 name = "python"
 enabled = false
+
+  [analyzers.meta]
+  runtime_version = "3.x.x"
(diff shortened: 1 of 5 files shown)


mauritsvanrees:
The tests pass, except that the DeepSource action fails, because there is no `.deepsource.toml` file in the 2.x branch.
Should I copy it?
Or can we ignore this error? Maybe we can disable this check for this branch.

icemac (approved):
LGTM.

Cool idea to support future Python versions in advance.

I think we can ignore DeepSource for the 2.x branch.
On a quick look I did not find a way to disable it for a specific branch.

mauritsvanrees:
I copied the `.deepsource.toml` file from master and now it works. I tried disabling it at first, but only got uninformative errors.

icemac:
> I copied the `.deepsource.toml` file from master and now it works. I tried disabling it at first, but only got uninformative errors.

I thought using deepsource on 2.x would produce a lot of errors. As it does not it seems to be the best solution.

gotcha (approved):
LGTM

Thanks for the work. @mauritsvanrees

mauritsvanrees:
@gotcha I have merged this. Could you do a release of 2.13.8, or grant me PyPI rights? User id `maurits`.

icemac:
@mauritsvanrees I just released https://pypi.org/project/zc.buildout/2.13.8/
(I was unable to grant you PyPI rights because I am just a maintainer not an owner.)

mauritsvanrees:
Thanks @icemac.

gotcha:
Thanks @icemac 

@mauritsvanrees I am also just maintainer...

PR #615, opened 2022-09-13
Add support for python311-315 in conditional section expressions. [master]

mauritsvanrees:


icemac (approved):
LGTM besides the failing tests but this seems already to be a problem on `master`.

Issue #613, opened 2022-09-09
Error message with pip 22.2+


mauritsvanrees:
I get this traceback:

```
$ bin/buildout 
Traceback (most recent call last):
  File "/Users/maurits/tmp/foo/lib/python3.8/site-packages/zc/buildout/patches.py", line 52, in patch_PackageIndex
    from pip._internal.index.collector import HTMLPage
ImportError: cannot import name 'HTMLPage' from 'pip._internal.index.collector' (/Users/maurits/tmp/foo/lib/python3.8/site-packages/pip/_internal/index/collector.py)
```

Buildout *does* continue after this, so it is not fatal.

To reproduce:

```
$ python3.8 -m venv foo  # tried with 3.10 as well, same thing
$ cd foo
$ bin/pip install --pre -U pip setuptools zc.buildout
$ cat buildout.cfg
[buildout]
parts = test

[test]
recipe = zc.recipe.egg
eggs = build
$ bin/buildout
```

Note: I also get a warning: "Requires-Python support missing", but I think that is now thrown if a package that we need to install is missing the `Requires-Python` metadata. So not something buildout can do anything about.

When I install `pip 22.1.2`, all is well again.

icemac:
I also saw this traceback but it just seems to be a warning, buildout still runs.

mauritsvanrees:
I am seeing it in tests failures targeting buildout main now too. See for example [here](https://github.com/buildout/buildout/actions/runs/3044061625/jobs/4904021394#step:5:6676), although the page is practically unreadable in my browser.
I am trying a fix in a PR.

mauritsvanrees:
> Note: I also get a warning: "Requires-Python support missing", but I think that is now thrown if a package that we need to install is missing the Requires-Python metadata. So not something buildout can do anything about.

I was wrong. This is a message logged by Buildout itself. It prints this when the patch fails.
So I realise the impact of the warning messages: if you use pip 22.2 or higher, then buildout does not have Requires-Python support. When run on Python 2 it will happily try to install a package that is marked as Python 3 only.
I am trying to fix it in PR #616.

PR #616, opened 2022-09-13
Make compatible with pip 22.2+, restoring Requires-Python patch

mauritsvanrees:
Fixes https://github.com/buildout/buildout/issues/613.

Note: we are patching `process_url` from `setuptools`. The existing comment says that this method was copied over from setuptools 46.1.3. I was wondering, so I checked: the method is still the same in latest setuptools.  And it is largely unchanged since setuptools 42.0.2. So for that part we should still be compatible with quite a long range of setuptools versions.

This PR also expands the warning that is printed when the patch fails. Until now it said "Requires-Python support missing." There was no indication that this was a message from buildout, so I assumed this was something from a newer setuptools or pip, and it warned about a package that did not have a `requires_python` in its `setup.py/cfg`. Now it should be clear that it is a warning from buildout itself.

### DIFF
--- a/news/613.bugfix
+++ b/news/613.bugfix
@@ -1,0 +1,3 @@
+Make compatible with pip 22.2+, restoring Requires-Python functionality there.
+Fixes `issue 613 <https://github.com/buildout/buildout/issues/613>`_.
+[maurits]
--- a/src/zc/buildout/patches.py
+++ b/src/zc/buildout/patches.py
@@ -40,18 +40,27 @@ patch_Distribution()
 
 
 def patch_PackageIndex():
+    """Patch the package index from setuptools.
+
+    Main goal: check the package urls on an index page to see if they are
+    compatible with the Python version.
+    """
+
     try:
         import logging
         logging.getLogger('pip._internal.index.collector').setLevel(logging.ERROR)
         from setuptools.package_index import PackageIndex
         from setuptools.package_index import URL_SCHEME
-        from setuptools.package_index import HREF
         from setuptools.package_index import distros_for_url
-        from setuptools.package_index import htmldecode
 
-        from pip._internal.index.collector import HTMLPage
+        try:
+            # pip 22.2+
+            from pip._internal.index.collector import IndexContent
+        except ImportError:
+            # pip 22.1-
+            from pip._internal.index.collector import HTMLPage as IndexContent
+
         from pip._internal.index.collector import parse_links
-        from pip._internal.index.collector import _clean_link
         from pip._internal.index.package_finder import _check_link_requires_python
         from pip._internal.models.target_python import TargetPython
         from pip._vendor import six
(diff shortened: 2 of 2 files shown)


mauritsvanrees:
About half of the jobs fail. Looks like it is a failure in only one test file, for example [here](https://github.com/buildout/buildout/actions/runs/3046036311/jobs/4908342006#step:5:281):

```
Failure in test /Users/runner/work/buildout/buildout/src/zc/buildout/tests/easy_install.txt
Failed doctest test for easy_install.txt
  File "/Users/runner/work/buildout/buildout/src/zc/buildout/tests/easy_install.txt", line 0

----------------------------------------------------------------------
File "/Users/runner/work/buildout/buildout/src/zc/buildout/tests/easy_install.txt", line 1277, in easy_install.txt
Failed example:
    zc.buildout.easy_install.build(
      'extdemo', dest,
      {'include_dirs': os.path.join(sample_buildout, 'include')},
      links=[link_server], index=link_server+'index/')
Expected:
    ['/sample-install/extdemo-1.5-pyN.N.egg']
Got:
    ['/sample-install/extdemo-1.4-pyN.N.egg']
```

It is hard to get a good overview of which jobs fail, but it looks like the 3.6 and lower jobs pass, while most 3.7+ jobs fail.

If someone can assist, I would appreciate it. After all these years I have not figured out yet how to run the tests locally. Feel free to edit the PR branch.

gotcha:
Thanks a lot for this. (I had tried to tackle the issue but did not get to satisfying code).

I will check if I can fix the remaining broken test.

I will also add some documentation how to run the tests locally.

mauritsvanrees:
I have rebased and force pushed this branch, after merging my unrelated other two much simpler PRs.

mauritsvanrees:
This contains just as many test failures as on master. So perhaps it could be merged and released. The new release would have this bug fix, and support for `[versions:python311]`, which starts getting needed.

I know, it would be better to not have test failures. But it is notoriously difficult to keep them passing on all environments that we support. On the other hand, it seems just one failure in [`easy_install.txt`](https://github.com/buildout/buildout/actions/runs/3157813641/jobs/5139127392#step:3:888) so that should be fixable.

gotcha:
I am worried that the test fail is related to changes in latest `setuptools`.
I would like to confirm this is not the case by testing with previous `setuptools` before releasing. Does that sound reasonable ?

mauritsvanrees:
@gotcha Sure, sounds reasonable.
I am doing a Plone beta 3 release later today, would be nice to have the latest fixes in, but it is no blocker. Don't rush it.

mauritsvanrees:
I think I have found the problem. My new code in this PR was missing `cache_link_parsing=False`, and this is still needed. I have pushed a fix.

mauritsvanrees:
BTW, I think this was only a problem in the tests: in real world usage the index page would be read only once and it would not change during a single buildout run.

mauritsvanrees:
@gotcha Yes, all is green!

gotcha:
Great !
Thanks @mauritsvanrees 

I have a small patch to merge and will do a release this weekend

gotcha:
Done !

https://pypi.org/project/zc.buildout/3.0.0/

Issue #617, opened 2022-09-13
gh-actions are run twice for PRs


mauritsvanrees:
When you create a PR for the master branch, over 80 gh-actions are run. This contains duplicates.
I tried fixing this along the lines of how it is done in the [`zopefoundation` repos](https://github.com/zopefoundation/meta/issues/145#issuecomment-1127295753) but I got this error:

> The workflow is not valid. `.github/workflows/debian-sys.yml` (Line: 13, Col: 9): Unexpected symbol: '"pull_request"'. Located at position 22 within expression: `github.event_name != "pull_request" || github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name`

mgedmin:
IIRC the GHA expression language is very strict and wants string literals to only use single quotes.  This is what the working expression used for zopefoundation packages looks like:
https://github.com/zopefoundation/meta/blob/9c24f7380973de989c5132ec9f28fa71f1397f20/config/c-code/tests.yml.j2#L93
```
    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
```

mauritsvanrees:
@mgedmin Thanks, it seems I copied an older version of that line.
I have created PR #619.

mauritsvanrees:
PR was merged.

PR #618, opened 2022-09-15
Prevent running gh-actions twice for push/pull.

mauritsvanrees:
Fixes issue #617.
Copied from the zopefoundation repositories.
See https://github.com/zopefoundation/meta/issues/145 and the final version with single quotes at [meta](https://github.com/zopefoundation/meta/blob/9c24f7380973de989c5132ec9f28fa71f1397f20/config/c-code/tests.yml.j2#L93).

I also added `workflow_dispatch` to allow to run these workflows manually from the Actions tab.

### DIFF
--- a/.github/workflows/centos.yml
+++ b/.github/workflows/centos.yml
@@ -1,12 +1,18 @@
 name: Centos containers
 
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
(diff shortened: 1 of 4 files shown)


mauritsvanrees:
I had an error and rebased, but let me close this PR and add a new one. Some checks were skipped, which might be because of the force push. I want to start fresh.

PR #619, opened 2022-09-15
Prevent running gh-actions twice for push/pull.

mauritsvanrees:
Fixes issue #617.
Copied from the zopefoundation repositories.
See https://github.com/zopefoundation/meta/issues/145 and the final version with single quotes at [meta](https://github.com/zopefoundation/meta/blob/9c24f7380973de989c5132ec9f28fa71f1397f20/config/c-code/tests.yml.j2#L93).

I also added `workflow_dispatch` to allow to run these workflows manually from the Actions tab.

### DIFF
--- a/.github/workflows/centos.yml
+++ b/.github/workflows/centos.yml
@@ -1,12 +1,18 @@
 name: Centos containers
 
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  # Allow to run this workflow manually from the Actions tab
+  workflow_dispatch:
 
 jobs:
   build:
 
     runs-on: ubuntu-latest
-    strategy:
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
+
+strategy:
       fail-fast: false
       matrix:
         python-version: ["2.7", "3.5", "3.6", "3.7", "3.8", "3.9", "3.10"]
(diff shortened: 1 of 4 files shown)


mauritsvanrees:
Seems to work, GitHub says: "40 queued, 3 skipped, and 1 successful checks".

Skipped is for example "Debian container with system Python / build (pull_request)". I don't know why this is skipped, instead of simply not shown. But in the queue we do have "Debian container with system Python / build (3.9) (push)", so that seems okay.

mauritsvanrees:
Python 3.6 and lower pass. 3.7 and higher fail, which is probably because they get a newer pip, which needs some fixes that I am working on in #616.
Since the current PR is only about preventing running duplicate GitHub Actions, I think this can be merged, and my other PR rebased.

.github/workflows/centos.yml
```diff
@@ -1,12 +1,18 @@
...
+  workflow_dispatch:
 
 jobs:
   build:
 
     runs-on: ubuntu-latest
-    strategy:
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
+
+strategy:
```
icemac, line 15: This indentation looks odd, at least changed from the previous version.

icemac (approved):
LGTM besides one comment.

.github/workflows/centos.yml
mauritsvanrees: Ah, an accidental change. Fixed in commit 343519a7f26f8df5c5dbdcd08f239c6bceb6f938.

Issue #621, opened 2022-11-03
Support PEP 508 markers in section expressions/conditions


mauritsvanrees:
Buildout currently supports section conditions like `[versions:python27]`.
It would be good if Buildout also supports the markers used in constraints files, like `python_version < "3.11" or platform_system == "Windows"`. See https://packaging.pypa.io/en/stable/markers.html

Main reason for me: this makes it much easier to create a script/package/extension that takes a pip constraints file and translates it to a Buildout versions file.
The basis is easy of course: `package==1.0` becomes `package = 1.0`. But how about this:

```
package==2.0
package==1.0; python_version <= "3.9"
```

In current buildout this would become the following, depending on how many Python versions you want to support:

```
[versions]
package = 2.0

[versions:python3.8]
package = 1.0

[versions:python3.9]
package = 1.0
```

I propose that we keep the current possibilities, but add support for the PEP 508 markers:

```
[versions]
package = 2.0

[versions:python_version <= "3.9"]
package = 1.0
```

I have some code and will add a PR.

PR #622, opened 2022-11-03
Add support for PEP 508 markers in section condition expressions.

mauritsvanrees:
For example: `[versions:python_version <= "3.9"]`. Fixes https://github.com/buildout/buildout/issues/621

I have not added tests yet. Difficult to run them locally: the pure Python that the Makefile tries to create is somehow lacking openssl support on my Mac, though the Makefile tries it. One of the Dockerfiles from gh-actions should help, but I run into network connection errors.

If you want to try it, here is a Buildout config with a conditional section that you can play with:

```
[buildout]
show-picked-versions = true
newest = false
eggs-directory = eggs
download-cache = downloads
parts = test

[test]
recipe = plone.recipe.command
command = echo "normal test"
update-command = ${:command}

[test:python_version < "3.11" or platform_system != "Windows"]
command = echo "difficult condition"

[versions]
plone.recipe.command = 1.1
```


mauritsvanrees:
I have rebased on master and added tests. I can run the tests locally with one of the Docker images now. Ready for review.

gotcha (approved):
Good idea and thanks for the implementation !

gotcha:
Included in https://pypi.org/project/zc.buildout/3.0.0/

mauritsvanrees:
> Included in https://pypi.org/project/zc.buildout/3.0.0/

Thanks!

Trying it out in [coredev](https://github.com/plone/buildout.coredev/pull/826) now with latest pip, setuptools, wheel. Unfortunately this [fails on Windows](https://github.com/plone/buildout.coredev/actions/runs/3409910424/jobs/5672247013):

```
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\zc\buildout\configparser.py", line 194, in parse
    marker = packaging.markers.Marker(expr)
AttributeError: module 'packaging' has no attribute 'markers'
```

I don't know why this would fail on Windows and pass on Linux and Mac.

mauritsvanrees:
I have made a fix (I hope): https://github.com/buildout/buildout/pull/624

Issue #627, opened 2023-03-07
vscode python extension passes `-I` option to the designated python interpreter which is unknown to the python interpreter wrapper installed by zc.recipe.egg


janwijbrand:
The recent (as of early 20230 Python extension for VSCode will call the designated Python interpreter for the workspace with the `-I` option. This option however is not known to the interpreter wrapper script installed using the zc.recipe.egg recipe. This will fail initializing the Python VSCode extension rendering it not usable.

The wrapper template that is used is:

https://github.com/buildout/buildout/blob/master/src/zc/buildout/easy_install.py#L1524

The `-I` option for Python is described as: "isolate Python from the user's environment".

Would it be an acceptable solution to allow the `-I` option in the wrapper template, even though it won't be used?

PR #628, opened 2023-03-07
#627 allow for the -I option in the Python interpreter wrapper

janwijbrand:
See issue #627.

### DIFF
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -1574,11 +1574,8 @@ sys.path[0:0] = [
 
 _interactive = True
 if len(sys.argv) > 1:
-    # The Python interpreter wrapper allows some of the options that a
-    # "regular" Python interpreter accepts, but it only acts on the -i, -c
-    # and -m options. The other option(s) are ignored. The specific use-case
-    # that led to this change is how the Python language extension for Visual
-    # Studio Code calls the Python interpreter when initializing the extension.
+    # The Python interpreter wrapper allows only some of the options that a
+    # "regular" Python interpreter accepts.
     _options, _args = __import__("getopt").getopt(sys.argv[1:], 'Iic:m:')
     _interactive = False
     for (_opt, _val) in _options:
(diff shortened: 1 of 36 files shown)


gotcha:
I am a bit uneasy with the fact that the option is accepted but silently swallowed. If ever anyone actually needs its normal effect, it will be hard to debug.

Would there be a way to advertise that is the flag is accepted but swallowed that would support your VSCode use case ?

janwijbrand:
I see your point. I don't have a very good answer, but: VSCode cannot really know it is not calling a "real" interpreter, so I see how it expects the `-I` to be accepted. 

`-I` indicates "I want an isolated python environment" that the wrapper actually sort of already provides - that was the line of reasoning I used for thinking this PR could be an OK solution.

Perhaps we could warn the user when they start the python wrapper in "interactive" mode and did pass in `-I` that the option is ignored? We can also add a comment in the wrapper about the `-I` being ignored, would that help?

Alternatively, we actually implement the "correct" `-I` behaviour. That seems overkill, but would provide the least surprise to the developer? 

There's a fair chance we will be playing catch up anyway: as soon as VSCode sees a good reason to pass another option native to the "normal" Python interpreter, we're back at square one.

janwijbrand:
I have no clear idea why https://github.com/buildout/buildout/actions/runs/4352805641/jobs/7670723752 is failing now. I can't really see a relation to my change. Anyone an idea of an approach to tackle this?

janwijbrand:
> Nevertheless, it could be easy to miss the silent swallow so I think that adding a comment does make sense.

I'll add a comment.

> Any chance to have VSCode provide some configuration to avoid the issue ?

I don't see how to make that into a sensible request to the extension developer/maintainer. In a way, buildout is the odd one out here I'm afraid.

src/zc/buildout/easy_install.py
```diff
@@ -1521,7 +1521,11 @@ def _pyscript(path, dest, rsetup, initialization=''):
 
 _interactive = True
 if len(sys.argv) > 1:
-    _options, _args = __import__("getopt").getopt(sys.argv[1:], 'ic:m:')
+    # The Python interpreter wrapper allows some of the options that a
+    # "regular" Python interpreter accepts, but it only acts on the -i, -c
+    # and -m options. The other option(s) are ignored. See
+    # https://github.com/buildout/buildout/issues/627 for more information.
```
gotcha, line 1527: @janwijbrand I used to refer to issues on websites, but they do eventually disappear. Would you mind explaining about VSCode need here ?

gotcha:
I see you did not add a test (which I am fine with). I guess you are also fine with the low risk that someone removes that 'useless' options. Is that right ?

janwijbrand:
I did not a new test indeed. I considered having had to update the existing test https://github.com/buildout/buildout/blob/3b14c3102e7f6ffd291987ae97bda9bd37adaa4c/src/zc/buildout/tests/easy_install.txt#L709 to be sufficient and also to express the intention. If someone were to remove the "useless" option, the test would fail. Right?

janwijbrand:
I've synced the PR with the master branch. It would be helpful to know if this PR is acceptable in the current state as I think (but I'm not 100% sure) I've addressed all concerns that have been raised. @gotcha @icemac ?

icemac (approved):

PR #633, opened 2023-10-30
remove an outdated note in custom.rst

perrinjerome:
This mentionned a zip_safe warning that is no longer emitted.


perrinjerome:
In https://github.com/buildout/buildout/commit/44046ba274e8ba8ceeea7af4a7e6a319bac9c23f the 
`"zip_safe flag not set; analyzing archive contents..."` message was removed from the output and ignored by a renormalizer, in 768502ce the rest of these messages were all removed, I believe this comment is no longer relevant.

Issue #577, opened 2021-09-03
Buildout can ignore pinned versions when processing "buildout" section


distributist:
There is a particular circumstance - when a substitution inside the buildout section comes from a section with a recipe - that versions specifications are ignored, and the latest version installed, even if allow-picked-versions is false.

What happens is that when `options = self['buildout']` is run at line 482 (all references to 2.13.4), if there are substitutions, `_dosub` automatically loads the appropriate section and creates options. But at line 1431, if the part has a recipe, `Options.initialize` is called, which calls `install_and_load` at line 1364. But although `install_and_load` is passed the buildout instance, it doesn't pass the versions, and so the newest version is picked.

The following test file reproduces the problem:
```
[buildout]
allow-picked-versions = false
show-picked-versions = true

test = ${data:test}
parts =

[versions]
mr.scripty = 1.0b3

[data]
recipe = mr.scripty
test =
    ... return 'TEST'
```

 When you run buildout with it, the output is:

```
Installing data.
Versions had to be automatically picked.
The following part definition lists the versions picked:
[versions]
mr.scripty = 1.0
```
Instead, 1.0b3 should have been installed, or an error been thrown because allow-picked-versions was false.

We can't easily avoid this in our configuration, so I've created a patch which passes the versions to the install command and enforces allow-picked-versions. I will be happy to share if interested.

icemac:
Thank you for writing this issue. Could you please create a pull request with your changes?

distributist:
> Thank you for writing this issue. Could you please create a pull request with your changes?

I won't be able to get to it for a few days, but will be happy to do so.

mauritsvanrees:
Fixed in your PR #637.

PR #637, opened 2023-12-21
Pin versions when processing buildout section

distributist:
Resolves Issue #577, where versions are not pinned when processing the buildout section, and any prerequisite sections that are processed with it.

I can also create a patch for the buildout 2.x branch, if that is actively being maintained.


mauritsvanrees (approved):
LGTM, thanks.

It seems tricky to use this corner case though.  At least it could lead to circles, where a buildout option references a section that references another buildout option, or maybe the same option.  But it is up to the user to avoid that.

PR #639, opened 2024-01-04
Fix accidental changes to PYTHONPATH in os.environ

xavth:
When generating an environment dict for subprocess calls to pip, os.environ was accidentally modified despite efforts to copy it and modify only the copy, as copy.copy(os.environ) is not enough.

Quick proof in a python shell:
```
>>> import os
>>> import copy
>>> env = copy.copy(os.environ)
>>> 'abc' in os.environ
False
>>> env['abc'] = 'xyz'
>>> 'abc' in os.environ
True
```


perrinjerome:
That's true. pylint has a linter rule for this https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/shallow-copy-environ.html 

BTW, in the discussions from https://bugs.python.org/issue15373 this nice approach is discussed:

```py
    exit_code = subprocess.call(list(args), env=dict(os.environ, PYTHONPATH=python_path))
```

mauritsvanrees (approved):
Sorry, late reaction.

The change makes sense to me.  I have merged the master branch into your branch, and the tests are now being run.  Let's see.

You could also rebase on master.  Or maybe you could incorporate the suggestion from [this comment](https://github.com/buildout/buildout/pull/639/#issuecomment-1882386818).

But the change should be fine, so I approve.

mauritsvanrees:
I have released `zc.buildout` 4.1.9 with this fix.  Thanks!

Issue #640, opened 2024-01-29
Make can fail if modules exist in ~/.local


distributist:
When make is run, if pip finds modules such as virtualenv in ~/.local, it will determine they don't need to be installed, and  later on make will fail when it tries to load them.

The solution is to define PYTHONUSERBASE to point to the Makefile's PYTHONPATH.

mauritsvanrees:
The Makefile has changed completely since this report, so I am going to assume this is no problem anymore.  You can reopen if you still run into errors.

PR #642, opened 2024-01-29
Issue 640 - Don't look for modules in ~/.local when running make

distributist:
Add PYTHONUSERBASE to makefile, so pip doesn't look for packages in ~/.local. Fixes issues #640


mauritsvanrees:
Sorry, very late review.

The change sounds fine.  But the `Makefile` on current master has changed rather drastically, and the change can't be applied.  Do you think something like this is still needed, either in `Makefile`, `prepare.sh` or `dev.py`?

mauritsvanrees:
Sorry, I did not mean to close this immediately.

mauritsvanrees:
I close this PR because it needs redoing, if it is still needed.

Issue #648, opened 2024-05-22
ImportError: cannot import name 'packaging' from 'pkg_resources' with setuptools 70


specialunderwear:
With setuptools 70 buildout will have an importerror in easy_install like this:

```
Traceback (most recent call last):
  File "/bin/buildout", line 5, in <module>
    from zc.buildout.buildout import main
  File "/lib/python3.10/site-packages/zc/buildout/buildout.py", line 18, in <module>
    import zc.buildout.easy_install
  File "/lib/python3.10/site-packages/zc/buildout/easy_install.py", line 28, in <module>
    from pkg_resources import packaging
ImportError: cannot import name 'packaging' from 'pkg_resources' (/lib/python3.10/site-packages/pkg_resources/__init__.py)
```

icemac:
@specialunderwear Thank you for this bugreport. For now please pin `setuptools` to a version smaller than `70` until this issue is fixed.

perrinjerome:
Note that the whole `pkg_resources` API (that buildout uses) is deprecated , there's a warning in 
https://github.com/pypa/setuptools/blob/52d732434377dac92df9b50a93eb78fe743b9e24/pkg_resources/__init__.py#L104-L109 referencing https://setuptools.pypa.io/en/latest/pkg_resources.html , so it's possible that more usages becomes broken in the future.

mauritsvanrees:
I have a PR to fix this: https://github.com/buildout/buildout/pull/649

But the tests fail in the same way as they were failing earlier on master.  I tried fixing it, but failed, so help would be appreciated.

npilon:
I think that properly fixing this requires making some headway on the PEP 420 implicit namespace packages issue; I've started to make some headway on that here: https://github.com/npilon/buildout/tree/setuptools-70-prep

But:

- I've got a lot of slightly different doctest output to adjust for, including a couple of warnings I'd like to get rid of.
- The tests that create a new distribution need adjustment.
- I need `zc.recipe.testrunner`, `zc.zdaemonrecipe`, `zc.recipe.deployment` built using implicit namespace packages - I worked around this for some tests by creating a `zc.recipe.testrunner_` local just like `zc.recipe.egg_` but I don't think anyone will be thrilled by that solution.

mauritsvanrees:
[...]

Issue #647, opened 2024-04-25
Cannot install recently created source distributions.


mauritsvanrees:
I have just released Plone 6.0.11.  Installing it with Buildout gives an error:

```
root: Reading https://pypi.org/simple/plone.app.locales/
Getting distribution for 'plone.app.locales==6.0.21'.
While:
  Updating instance.
  Getting distribution for 'plone.app.locales==6.0.21'.
Error: Couldn't find a distribution for 'plone.app.locales==6.0.21'.
```

When you follow the [PyPI link](https://pypi.org/simple/plone.app.locales/), the last two releases are this:

* `plone.app.locales-6.0.20.tar.gz`
* `plone_app_locales-6.0.21.tar.gz`

Version 6.0.20 from January (included in Plone 6.0.10) can be installed fine.
Version 6.0.21, released earlier this week, cannot.
The difference is underscores in the filename instead of dots.

A `pip install` worked fine.

I have fixed it by uploading a wheel, which should have been done anyway, and then it works.  The wheel filename is `plone.app.locales-6.0.21-py3-none-any.whl`, so with dots, which apparently makes it work.

`plone.app.locales` was released by someone else, but how about packages I released as Plone release manager?  Look at [`plone.base`](https://pypi.org/simple/plone-base/) and you see the same. In March I released `plone.base-1.3.0.tar.gz`, and this week `plone_base-1.4.0.tar.gz`.

Why is there a difference in the name?  Not sure.  I use `zest.releaser` to create releases, usually with a source checkout from the master branch as I am one of the developers of this tool.  During release, this basically calls `python -m build .`  This uses the [`build`](https://build.pypa.io/en/stable/changelog.html) package.  Let's try a bit in a checkout of [`plone.base`](https://github.com/plone/plone.base).  Note that this has a `pyproject.toml` which since January defines:

```
[build-system]
requires = ["setuptools>=68.2"]
```

That may influence things.  Let's try to build stuff:

```
$ python3.12 -mvenv venv
$ . venv/bin/activate
$ pip install build
$ pip freeze --all
build==1.2.1
packaging==24.0
pip==24.0
pyproject_hooks==1.0.0
$ python -m build .
...
Successfully built plone_base-1.4.1.dev0.tar.gz and plone.base-1.4.1.dev0-py3-none-any.whl
$ ls dist
plone.base-1.4.1.dev0-py3-none-any.whl plone_base-1.4.1.dev0.tar.gz
```

[...]

PR #649, opened 2024-05-24
Fix ImportError: cannot import name 'packaging' from 'pkg_resources'

mauritsvanrees:
This is with setuptools 70.
Done by adding a compatibility module that tries to import `packaging` from several places. 

Fixes https://github.com/buildout/buildout/issues/648

This should partially fix GitHub Actions, but they were already failing for a long time due to other reasons.

### DIFF
--- a/src/zc/buildout/_compat.py
+++ b/src/zc/buildout/_compat.py
@@ -1,21 +1,31 @@
 try:
+    # The happy path is if someone has the 'packaging' module installed .
+    # We may want to add this to our own dependencies at some point.
     from packaging import markers
     from packaging import specifiers
     from packaging import utils as packaging_utils
     from packaging import version
 except ImportError:
     try:
+        # It is quite likely that pip has a version of packaging vendored.
+        # But it is in a '_vendor' module, marked as private with an underscore,
+        # so we should not rely on this: it may get moved or removed.
         from pip._vendor.packaging import markers
         from pip._vendor.packaging import specifiers
         from pip._vendor.packaging import utils as packaging_utils
         from pip._vendor.packaging import version
     except ImportError:
         try:
+            # This works with the pkg_resources from setuptools 19.3
+            # until at least 70, but I wonder if that may change.
             from pkg_resources.extern.packaging import markers
             from pkg_resources.extern.packaging import specifiers
             from pkg_resources.extern.packaging import utils as packaging_utils
             from pkg_resources.extern.packaging import version
         except ImportError:
+            # Should work in most setuptools versions, but that may just be
+            # because they import it, and could change if they restructure
+            # their code.
             from pkg_resources.packaging import markers
             from pkg_resources.packaging import specifiers
             from pkg_resources.packaging import utils as packaging_utils
--- a/src/zc/buildout/_compat.py
+++ b/src/zc/buildout/_compat.py
@@ -1,0 +1,20 @@
+try:
+    from packaging import markers
+    from packaging import specifiers
+    from packaging import utils as packaging_utils
+except ImportError:
+    try:
+        from pip._vendor.packaging import markers
+        from pip._vendor.packaging import specifiers
+        from pip._vendor.packaging import utils as packaging_utils
+    except ImportError:
+        try:
+            from pkg_resources.extern.packaging import markers
+            from pkg_resources.extern.packaging import specifiers
+            from pkg_resources.extern.packaging import utils as packaging_utils
+        except ImportError:
+            from pkg_resources.packaging import markers
+            from pkg_resources.packaging import specifiers
+            from pkg_resources.packaging import utils as packaging_utils
+
+__all__ = ["markers", "specifiers"]
--- a/src/zc/buildout/_compat.py
+++ b/src/zc/buildout/_compat.py
@@ -2,19 +2,23 @@ try:
     from packaging import markers
     from packaging import specifiers
     from packaging import utils as packaging_utils
+    from packaging import version
 except ImportError:
     try:
         from pip._vendor.packaging import markers
         from pip._vendor.packaging import specifiers
         from pip._vendor.packaging import utils as packaging_utils
+        from pip._vendor.packaging import version
     except ImportError:
         try:
             from pkg_resources.extern.packaging import markers
             from pkg_resources.extern.packaging import specifiers
             from pkg_resources.extern.packaging import utils as packaging_utils
+            from pkg_resources.extern.packaging import version
         except ImportError:
             from pkg_resources.packaging import markers
             from pkg_resources.packaging import specifiers
             from pkg_resources.packaging import utils as packaging_utils
+            from pkg_resources.packaging import version
 
 __all__ = ["markers", "specifiers"]
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -25,7 +25,6 @@ import glob
 import logging
 import os
 import pkg_resources
-from pkg_resources import packaging
 import py_compile
 import re
 import setuptools.archive_util
@@ -40,6 +39,8 @@ import zc.buildout
 import zc.buildout.rmtree
 from zc.buildout import WINDOWS
 from zc.buildout import PY3
+from zc.buildout._compat import packaging_utils
+from zc.buildout._compat import specifiers
 import warnings
 import csv
 
@@ -1642,7 +1643,7 @@ def _constrained_requirement(constraint, requirement):
             msg = ("The requirement (%r) is not allowed by your [versions] "
                    "constraint (%s)" % (str(requirement), version))
             raise IncompatibleConstraintError(msg)
-        specifier = packaging.specifiers.SpecifierSet(constraint)
+        specifier = specifiers.SpecifierSet(constraint)
     else:
         specifier = requirement.specifier & constraint
     constrained = copy.deepcopy(requirement)
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -23,6 +23,7 @@ import distutils.errors
 import errno
 import glob
 import logging
+import operator
 import os
 import pkg_resources
 import py_compile
@@ -41,6 +42,7 @@ from zc.buildout import WINDOWS
 from zc.buildout import PY3
 from zc.buildout._compat import packaging_utils
 from zc.buildout._compat import specifiers
+from zc.buildout.utils import normalize_name
 import warnings
 import csv
 
@@ -123,6 +125,46 @@ class _NoWarn(object):
 
 _no_warn = _NoWarn()
 
+
+class Environment(pkg_resources.Environment):
+    """Buildout version of Environment with canonicalized names.
+
+    * pkg_resources defines the Environment class
+    * setuptools defines a PackageIndex class that inherits from Environment
+    * Buildout needs a few fixes that should be used by both.
+
+    The fixes are needed for this issue, where distributions created by
+    setuptools 69.3+ get a different name than with older versions:
+    https://github.com/buildout/buildout/issues/647
+
+    I wanted to do this with a mixin class, but somehow the changes
+    did not get picked up.  So we duplicate '__getitem__' and 'add'
+    here and in the PackageIndex class.
+    """
+
+    def __getitem__(self, project_name: str):
+        """Return a newest-to-oldest list of distributions for `project_name`
+
+        Uses case-insensitive `project_name` comparison, assuming all the
+        project's distributions use their project's name converted to all
+        lowercase as their key.
+
+        """
+        distribution_key = normalize_name(project_name)
+        return self._distmap.get(distribution_key, [])
+
+    def add(self, dist: "Distribution"):
+        """Add `dist` if we ``can_add()`` it and it has not already been added
+        """
+        if self.can_add(dist) and dist.has_version():
+            # Instead of 'dist.key' we add a normalized version.
+            distribution_key = normalize_name(dist.key)
+            dists = self._distmap.setdefault(distribution_key, [])
+            if dist not in dists:
+                dists.append(dist)
+                dists.sort(key=operator.attrgetter('hashcmp'), reverse=True)
+
+
 class AllowHostsPackageIndex(setuptools.package_index.PackageIndex):
     """Will allow urls that are local to the system.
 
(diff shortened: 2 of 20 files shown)


mauritsvanrees:
The tests fail in the same way as they were failing earlier on master:

```
    Getting distribution for 'zc.recipe.egg>=2.0.6'.
    root: Found: /tmp/tmpiotpjs5xbuildouttests/zc.recipe.egg/zc_recipe_egg-2.0.8.dev0.zip
    root: Found: /tmp/tmpiotpjs5xbuildouttests/zc.recipe.egg/zc_recipe_egg-2.0.8.dev0.zip
    root: Couldn't find index page for 'zc.recipe.egg' (maybe misspelled?)
    root: Scanning index of all packages (this may take a while)
    Couldn't load zc.buildout entry point default
    from zc.recipe.egg:
    Couldn't find a distribution for 'zc.recipe.egg>=2.0.6'..
```

I think it would help if we set `prefer-final = false` so the available development egg can be used.  But I tried this in two ways, with a command line assignment and a `HOME/.buildout/default.cfg`, but it fails.  See [command here](https://github.com/buildout/buildout/actions/runs/9223251832/job/25376078508?pr=649#step:5:292).  It is in a part of the tests where we run `buildout init` to create a buildout config on the fly, so we cannot simply add `prefer-final = false` to the buildout config.

Can anyone help?

icemac:
@gotcha Do you have any idea why the tests are failing or how this can be fixed?

mauritsvanrees:
I was just looking at this again and tried some more.  I assumed so far that it could start working again if I manage to get prefer-final=false in there, but that had no effect after I got it working (code only local).

Instead, the tests pass when I do `export SETUPTOOLS_VERSION=65.5.0`.  Well, one test still fails but that seems to only be a slightly different output, need to look at that more closely.  But at least it seems to find the dev version of zc.recipe.egg then.  So the remaining problem would be this issue: https://github.com/buildout/buildout/issues/647

mauritsvanrees:
I tried some more, see new commits, but still failed.

I suppose the part of this PR that fixes "ImportError: cannot import name 'packaging' from 'pkg_resources'" would be fine to merge and release.  But fixing the tests is currently beyond me.

mauritsvanrees:
Success!!  The main test job succeeds!  That was a beast.  Please check.

Now we need to fix the GitHub actions setup, because I get errors like "Error: Version 2.7 with arch x64 not found".  I tried one thing, but failed.  I will try it on a different branch.

amotl:
Hi. We also have been tripped by this problem, and fixed it by downgrading the CI environment to `setuptools<70`. Hope this helps for a temporary workaround.

- https://github.com/crate/crate-python/pull/617

mauritsvanrees:
I managed to fix most of the problems in CI, even on Python 2.7.
I have experimented on PR #651, partially commenting tests that were already passing during an earlier run.  The end result is:
https://github.com/buildout/buildout/actions/runs/9555363465
That shows various container tests that fail early during setup, before zc.buildout is involved.  I will leave that PR in that status for a bit, if anyone wants to see those errors to fix them.

On the current PR I took over the test yaml file, but with the temporary changes reverted.  And not 42 commits, but just one. ;-)

mauritsvanrees:
So, status of the current PR:

* It fixes issues #647.
* It fixes issues #648.
* It fixes CI.
* It drops Python 3.5 support because that is really too old to test.
* It drops testing on about half of the containers, including all of the CentOS containers, because they fail very early.  They are commented out, so someone can pick this up later.

Please review it, and try the branch out in practice in your projects!

cc @npilon @specialunderwear

npilon:
> Please review it, and try the branch out in practice in your projects!

Oh I see, the root logger ignoring is very clever. I believe the change that necessitated it is that setuptools is in the process of changing how they pass options, so the old method doesn’t work. I thought I was getting these because I needed to pass quite differently when invoking it through pip instead of its own UI, but if you’re still getting them that was clearly wrong.

I’m curious, why not just depend on packaging instead of a three-level fallback?

I also don’t think continuing to ignore setuptools depreciation warnings is wise. They’ve clearly become very serious about removing deprecated features.

icemac:
Maybe we can/should drop support for Python < 3.8 (incl. 2.7) to reduce the maintenance burden.

icemac:
@mauritsvanrees Thank you for investing time and effort here to fix this beast.

src/zc/buildout/_compat.py
icemac, line 1: Could you please add to the code paths when they are triggered?
(So we can remove them when a certain version or combination is no longer supported.)

src/zc/buildout/easy_install.py
```diff
@@ -122,6 +125,46 @@ def warn(self, *args, **kw):
...
+    * setuptools defines a PackageIndex class that inherits from Environment
+    * Buildout needs a few fixes that should be used by both.
+
+    The fixes are needed for this issue, where distributions created by
+    setuptools 69.3+ get a different name than with older versions:
+    https://github.com/buildout/buildout/issues/647
+
+    I wanted to do this with a mixin class, but somehow the changes
+    did not get picked up.  So we duplicate '__getitem__' and 'add'
+    here and in the PackageIndex class.
```
icemac, line 142: You probably had the MixIn class already the first item in the list of base classes. – This is my usual recipe to ensure the methods of the MixIn class get picked up.

src/zc/buildout/easy_install.py
```diff
@@ -560,6 +625,9 @@ def _get_dist(self, requirement, ws):
 
             # Retrieve the dist:
             if avail is None:
+                # TODO I don't think the next line is useful.  When we get
+                # here, I expect that 'obtain' was already called and has
+                # returned None, which is why we are here.
```
icemac, line 630: Is there any change in behaviour of the tests when omitting the next line? Or is this code path untested?

icemac (approved):
I admire the work you did here. I read through the code and left some comments.

mauritsvanrees:
> I’m curious, why not just depend on packaging instead of a three-level fallback?

We could do that, but I don't want to add a dependency at this time for a bug.  Could also be tricky to get a working version for each Python version.
 
> I also don’t think continuing to ignore setuptools depreciation warnings is wise. They’ve clearly become very serious about removing deprecated features.

We are only ignoring them in tests because otherwise we would have to make thousands of changes in the doctests to make the expected output match the actual output.

In real life usage, you still see all the setuptools warnings.

mauritsvanrees:
> Maybe we can/should drop support for Python < 3.8 (incl. 2.7) to reduce the maintenance burden.

For this PR it was mostly okay, except for trying to support 3.5.  Well, 2.7 was no fun due to how it needs to be handled specially on GitHub Actions, but in the end it worked.  So for this round it is still fine, and we can get a release out that works on those old releases.  Probably a 3.1.0b1 at first.

But I would definitely want to drop support for 3.7 and lower.  That would be a 4.0.0 release.

(BTW, we should test on 3.12 as well, but that can be another PR.  And 3.13 may be a nightmare due to `distutils` being gone.)

src/zc/buildout/_compat.py
mauritsvanrees: Done.

src/zc/buildout/easy_install.py
mauritsvanrees: Ah, I had expected that the method from the last class would be taken over.  I did not try it the other way around.  But the first one wins.  I have added the mixin class back.

src/zc/buildout/easy_install.py
mauritsvanrees: I tried, and the extra `_obtain` call is safe to remove.  I have removed it now, and added some explanatory comments in this method.

pradyunsg:
Hi folks! Is there a particular reason you aren't adding a dependency on packaging explicitly?

mauritsvanrees:
> Hi folks! Is there a particular reason you aren't adding a dependency on packaging explicitly?

We could do that, but I don't want to add a dependency at this time for a bug.  Could also be tricky to get a working version for each Python version.  We could do that when we drop support for older Pythons.  Note that we still support Python 2.7...  Well, 20.9 seems the last version that works there.  So it can be done, but so far I am okay with a small compatibility module.

gotcha:
@mauritsvanrees I do have time in the coming days. Will come back asap.

mauritsvanrees:
I am mostly unavailable the next two weeks.  So if some extra changes are needed, contributors can push changes to this PR branch as well.

Note that I have commented out the CentOS jobs because they failed very early.  But I recently heard that development on CentOS has mostly stopped, so that may be fine.

icemac:
@gotcha Did you find the time to work on this pull request? (Currently the tests of the recipes fail and I did not find a way to successfully pin `setuptools` there.)

tseaver:
+1 from me on dropping support for *all* Pythons < 3.8:  I was astonished that `buildout` still carries the burden of `2.7` compatibility at this late date.

icemac:
All running tests are green now. Do we need to re-enable the tests for centos and debian before merging?

mauritsvanrees:
> All running tests are green now. Do we need to re-enable the tests for centos and debian before merging?

For Debian they are running.  I see them in the checks here.
You may be confused (or I myself certainly am) because because there are different runs:

* Some tests are run because this is a pull request: https://github.com/buildout/buildout/actions/runs/10508191100
* Other tests are run because I push a commit to a branch: https://github.com/buildout/buildout/actions/runs/10508190342

So there may be something wrong with the gitHub Actions conditions.  But at least on that last one the Debian containers run.

What is lost in this PR, is:

* I disabled the Debian tests on Python 3.5 and 3.6.
* I disabled the Debian container with system Python.
* I disabled the CentOS tests.

I had earlier tried fixing some of this in https://github.com/buildout/buildout/pull/651, and the needed parts of that are included in the current PR.
Latest run from that branch is here: https://github.com/buildout/buildout/actions/runs/9555363465

* Debian Python 3.5 seems to [fail](https://github.com/buildout/buildout/actions/runs/9555363465/job/26338465968#step:3:655) because `/home/buildout/sandbox/src` does not exist, so something may be wrong with the source checkout, but I have no clue why that would be different from the more recent Pythons.
* Debian Python3.6 [fails](https://github.com/buildout/buildout/actions/runs/9555363465/job/26338466622#step:3:709) fails with a `SIGSEGV` when calling pip.
* System Python [fails](https://github.com/buildout/buildout/actions/runs/9555363465/job/26338463691#step:3:795) because the environment is externally managed.
* CentOS [fails](https://github.com/buildout/buildout/actions/runs/9555363465/job/26338470509) installing packages with yum: "Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist"

CentOS as a whole is basically deprecated/abondoned as far as I know, so for me it is fine if they stay disabled.  If there are CentOS fans here they can try to resurrect this later.

Meanwhile I saw yesterday dat [Circle CI fails](https://app.circleci.com/pipelines/github/buildout/buildout/329/workflows/8e18b717-6ab0-4097-a42b-6439c4362015) for this branch.  I have made separate PR #655 to try and fix that, which partially already works.  So I could already cherry-pick some commits from there.  Okay, I did so just now.

Latest status on CircleCI is that the [Ubuntu tests pass](https://app.circleci.com/pipelines/github/buildout/buildout/335/workflows/24778ad4-ce1d-4f22-8e42-0501ec617888), but that triggers test runs for Debian and CentOS.

* For CentOS it fails in the same way as on GitHub Actions: `yum` does not work.
* Debian with Python 3.5 and 3.6 it fails in the same way as above.
* Extra on CircleCI are tests on MacOS, but these fail with "Job was rejected because resource class macos.m1.medium.gen1, image xcode:12.5.1 is not a valid resource class"
* The Debian system container tests are skipped, because those other tests fail.

I wonder how much we should care about the skipped stuff.  And maybe we should also dump CircleCI: I currently don't see the added value of CircleCI next to GitHub Actions.
I appreciate the Herculean effort @gotcha put into setting up this whole testing infrastructure.
But I think an inordinate amount of work is going into fixing those test machine definitions to reach a situation where `zc.buildout` can actually be installed so we can finally start the real tests.  I think the only way of installing Buildout that we can properly support and test, is to create a virtualenv, and let pip install setuptools, zc.buildout and wheel.  We should focus on that.

And instead of testing with such a gigantic matrix of Python and OS versions, we should test with a few different versions of pip and setuptools.  I think we now always get the latest version that can be installed on the Python version.  There could be some breakage with pip 22 and setuptools 61 that we never catch.  And setuptools seems to keep breaking stuff that we rely on with each new major (or sometimes minor) release.

Anyway, I am very much in favour of merging this now (once the existing tests are green again) and releasing this as version 3.1.0a1.

mauritsvanrees:
I could not resist a few more tries for CircleCI.  Now the MacOS test at least passes.  And I have disabled the Debian system Python test.
Latest [Circle CI result](https://app.circleci.com/pipelines/github/buildout/buildout/340) is now fully green.

Here in the PR the reported checks are still a bit weird, maybe GitHub gets confused because there are so many.  You need to look at two workflow runs:

* [Here](https://github.com/buildout/buildout/actions?query=event%3Apush+branch%3Amaurits-setuptools-70) is the list of runs for _pushing_ to this branch.  Those are the main interesting ones that run the 600 or so tests.
* [Here](https://github.com/buildout/buildout/actions/workflows/run-tests.yml?query=event%3Apull_request+branch%3Amaurits-setuptools-70) is the list of runs for the _pull request_ event of this branch.

In both cases the last run was successful.

This is ready for merge and release.

mauritsvanrees:
I fixed the double GitHub actions confusion.  Two of the jobs did not have the proper condition that the others had.  That was my mistake.
Now in the checks I see 6 skipped checks (these are the ones acting on a pull_request event, but the extra condition is not true) and 29 successful checks (these are the ones acting on a push event).

I also did some fixes for the long broken weekly "Script generation with setuptools head" tests, but those can only be tested once this is merged.  The last success was in [June last year](https://github.com/buildout/buildout/actions/workflows/scripts-setuptools-head.yml?query=is%3Asuccess), after which it failed because setuptools head is only for Python 3.8+ and we also test it with Python 3.7.

icemac (approved):
LGTM from scanning through the code.

@mauritsvanrees Do you think it is ready to be merged?

dataflake:
I'm eager to see this merged, too.

mauritsvanrees:
Yes, this is ready to be merged.  I don't have any more changes coming.

PR #651, opened 2024-06-17
WIP: try to fix CI

mauritsvanrees:
This builds on my PR #649.  CI is still broken there, so I try to fix it.
Do not merge, I will temporarily rip out lines, and try stuff that does not work.


icemac:
What is the difference to #649 besides that the tests there are all green?

mauritsvanrees:
The difference is that over here most of the passing jobs have been commented out, so the focus can be on the remaining failing jobs.

Also the dependencies ('needs' keyword) on earlier jobs have been removed, so you don't need to wait on those, but immediately the remaining jobs are started.

On the other PR, the failing jobs (like all CentOS jobs) have been commented out, as I don't want those to hold up a release.

I will close this PR.  The reports from the last run will be available for a while if someone wants to jump in to fix them.

Issue #656, opened 2024-08-22
Multiple +=/-= in one file loses assignment in a previous file


distributist:
If you define an option in one file, then add or subtract from it twice in a subsequent file, the original value is lost, not extended. This happens when you extend the option in the second file, then need to use a conditional section in the same file to extend it further, depending on some condition.

**buildout.cfg**
```
[buildout]
extends =
    b1.cfg
    b2.cfg

parts = check

[check]
recipe = plone.recipe.command
command =
    echo "${test:foo}"
update-command = ${:command}
```
**b1.cfg**
```
[test]
foo =
    abc
    def
```
**b2.cfg**
```
[test]
foo +=
    ghi
    jkl

[test:False]
foo +=
    xzy

[test:True]
foo +=
    mno
```

**Expected:**
```
abc
def
ghi
jkl
mno
```
**Actual:**
```
ghi
jkl
mno
```

I have a patch and will be submitting a pull request shortly

mauritsvanrees:
Your fix is in release [3.1.1](https://pypi.org/project/zc.buildout/3.1.1/).

Issue #641, opened 2024-01-29
Variable defined with initial += is undefined and .installed.cfg corrupted


distributist:
A variable which is initially defined with += will end up with the variable undefined, or missing the initial definition. This is most likely to happen in complex buildouts with multiple overrides.

Such a case will corrupt .installed.cfg by adding an unparsable line "var + = base0" to the end of the offending part:
```
[base1]
__buildout_installed__ = 
__buildout_signature__ = wheel-339de3cbbb8d2a18c6d8133fb43aa777 zc.buildout-1a1dfc3477a86238088ba01adcf15436 pip-339de3cbbb8d2a18c6d8133fb43aa777 setuptools-339de3cbbb8d2a18c6d8133fb43aa777
recipe = zc.buildout:debug
var + = base0
```

mauritsvanrees:
The fix is in release [3.1.1](https://pypi.org/project/zc.buildout/3.1.1/).

PR #659, opened 2024-09-05
Updated Patches for Issues 641 and 656

distributist:
Issue #656 : My earlier patch did not take into account all the various combinations of operators that could occur in multiple sections. This one does, and handles all of them properly. It also calls _update_section to ensure that any +=/-= operators are applied properly. 

The problem in Issue #641 needs to be dealt with for conditional sections, and occurs in the same place as the _update_section patch for 656, so it is included here.

I've also applied these changes to the latest buildout master branch. Tests have been updated, and are all passing under both Python 2.7 and 3.9 on my system.

### DIFF
--- a/src/zc/buildout/buildout.py
+++ b/src/zc/buildout/buildout.py
@@ -1996,8 +1996,32 @@ def _update(in1, d2):
     for section in d2:
         if section in d1:
             d1[section] = _update_section(d1[section], d2[section])
+        elif '<' not in d2[section].keys():
+            # Skip sections that extend in other sections (macros), as we don't
+            # have all the data (these will be processed when the section is
+            # extended)
+            temp = copy.deepcopy(d2[section])
+            # 641 - Process base definitions done with += and -=
+            for k, v in sorted(temp.items(), key=lambda item: item[0]):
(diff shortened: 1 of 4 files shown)


mauritsvanrees (approved):
Excellent. Thanks for adding tests, that gives confidence in the updated code.  I tried it in one project that has a few `+=` assignments, though not in such a difficult way as in your examples, and that still worked.

distributist:
That's wonderful. Thank you very much.

mauritsvanrees:
I am inclined to let someone else merge this, as I have no release rights on PyPI.

icemac:
@mauritsvanrees I would have liked to give you the PyPI rights but I only have the maintainer role, not the owner one, so I am not able to do so.

With a change log entry for this PR I could cut a release.

mauritsvanrees:
I have added changelog entries.

mauritsvanrees:
@icemac Can you see on PyPI who has the owner role?  As normal user I only see a list of maintainers.

icemac:
@mauritsvanrees Sorry, even as a maintainer I only can see the list of persons but not the roles they have.

PR #665, opened 2024-09-23
Add config option: optional-extends.

mauritsvanrees:
This is the same as the `extends` option, but then for optional files. The names must be file paths, not URLs.  If the path does not exist,  it is silently ignored.

This is useful for optionally loading a `local.cfg` or `custom.cfg` with options specific for the developer or the server.

I have seen various ways of having local options without checking this into version control:

* Let git ignore `buildout.cfg` and have a sample `buildout.cfg.in` that you copy to `buildout.cfg` and adapt for the machine.
* Have a `local.cfg` that extends `buildout.cfg` or an alternative like `devel.cfg` or `production.cfg`.
* Let `buildout.cfg` extend `custom.cfg` which is not in version control, so you get an error when you don't yet have this file.

When this PR is merged, I intend to use this in the Plone core development buildout.


icemac (approved):
The idea seems to be interesting. I am not sure whether I'll be going to use it (went without it for > 10 years).

Mabye a message would be handy if the optional file does not exist. Could be rendered only in vebose mode (?).

mauritsvanrees:
> Mabye a message would be handy if the optional file does not exist. Could be rendered only in vebose mode (?).

Good point.  I now print a message when not found, regardless of verbosity:

    optional-extends file not found: somefile.cfg

It would usually be the first line that is printed.  This seems a useful hint for anyone using a Buildout with optional-extends.

gotcha:
Good idea !

Thanks for working on this !

mauritsvanrees:
Thanks for the reviews.  I have released 3.2.0 with this.

distributist:
I just saw this in the pypi release notes. I've been using a workaround for this for a few years now; optional-extends is a great idea. Thank you!

PR #666, opened 2024-09-27
Nuke trailing white space.

mauritsvanrees:
My editor is configured to remove trailing whitespace on save. Several files that I touched lately would in this way get irrelevant changes, that I then mostly undid (using `git add --patch`) to avoid distracting reviewers. But now I removes all trailing whitespace.

That *did* lead to one test failure, which I fixed by normalizing whitespace in one doctest. Failure was (:

```
File "/Users/maurits/community/buildout/venvs/3.8/src/zc/buildout/tests/test_increment.py", line 375, in zc.buildout.tests.test_increment.no_default_with_extends_increment_in_base2_and_base3
Failed example:
    print_(system(buildout+' annotate buildout'), end='')
    # doctest: +ELLIPSIS
Expected:
    <BLANKLINE>
    Annotated sections
    ==================
    <BLANKLINE>
    [buildout]
    ...
    extensions=
    demo2
    demo3
        IMPLICIT_VALUE
    +=  base2.cfg
    +=  base3.cfg
    ...
    versions= versions
        DEFAULT_VALUE
Got:
```

And then under `GOT:` I got seemingly the same output. The difference was that we now expect `extensions=` and the output has `extensions= ` with a space at the end. I looked briefly at fixing that in the `annotate` command, but did not see how. So I added `+ELLIPSIS +NORMALIZE_WHITESPACE` to this doctest.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -169,7 +169,7 @@ Bug fixes:
 
 
 - Fix support of ``pip>=21.1`` (#567)
-- Fix confusion when using multiple Python versions and 
+- Fix confusion when using multiple Python versions and
   installing packages with C extensions
   without proper binary wheel available. (#574)
 
--- a/doc/reference.rst
+++ b/doc/reference.rst
@@ -187,9 +187,9 @@ When you query the ``buildout`` section, you can pass the key only. For instance
 .. code-block:: console
 
    buildout query parts
-   
+
 is equivalent to the command above.
-   
+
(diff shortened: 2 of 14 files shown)


mauritsvanrees:
Hint: you can review this without showing whitespace changes by adding [`?w=1`](https://github.com/buildout/buildout/pull/666/files?diff=unified&w=1) to the files url.

PR #667, opened 2024-10-04
Drop Python 2, 3.7, add Python 3.12, 3.13, simplify testing

mauritsvanrees:
These are a lot of commits that together hopefully make it easier to maintain and develop Buildout, especially with newer pip and setuptools versions coming in that regularly mean some breakage.  This should end up in a new major release, so before merge we should make a 3.x branch for possible maintenance.
Let me highlight the main points of this PR.

## Python versions

* This drops support for Python 2.
* This drops support for Python 3.7 and lower.
* This adds support for Python 3.12 and 3.13.  I think this only needed a few test fixes.
* I manually cleaned up the code, getting rid of conditions and other compatibility constructs that are no longer needed.
* I did **not** run large scale automatic code modernisers like isort, black, pyupgrade.  I did not want to bloat the current PR with it.

## CI

* This drops CircleCI.  I don't see any remaining benefit in keeping using that.
* This drops testing on bare machines, testing on CentOS and Debian, downloading Python tarballs, testing without any `setuptools` previously installed.  That may have made sense when we still had `bootstrap.py` as officially supported installation mechanism.  Now it is the responsibility of the user to create a virtualenv and use `pip install zc.buildout`.
* This restores testing on Windows.  This got lost during my big fixes for 3.1.0.  I don't remember if I could not get it to work then with reasonable effort, or if I forgot to re-enable it after temporarily switching it off.

## GitHub Actions testing matrix

* We first test on a range of setuptools versions, from 49 to 75.  All on ubuntu with Python 3.10.
* If those pass, we run on a range of Python versions, from 3.8 to 3.13.0-rc.3.  All on Ubuntu with setuptools 75.  We skip Python 3.10 because this combination is already in the first batch.
* If those pass, we run on a range of pip versions, from 20.3.4 to 24.2.  We combine this with setuptools 65 and 75.  All on Ubuntu with Python 3.10.
* If those pass, then we run the last three batches:
  * macos-latest, Python 3.10, setuptools 75
  * windows-latest, Python 3.10, setuptools 75
  * generate-scripts: use Buildout to install `zest.releaser` and `pyspf` on Python 3.8-3.13, Ubuntu.

## Local development

* This simplifies local development, because we no longer install bare Pythons.
* We have a short `Makefile` (call `make help`), which calls a new `prepare.sh` that creates a venv and then calls a shortened `dev.py` that creates a `bin/buildout` script and then calls it.  The end result is basically the same as before: you get a `bin/test` script that you can call.
* When you run `make test`, we run `PYTHONWARNINGS=ignore bin/test`.  This environment variable is used in as much places as possible.  This avoids an uphill battle against setuptools deprecation warnings that lead to doctest failures and avoids having 1000 lines of output that you have to wade through to get to the important log lines.

## Setuptools

I made an effort to discover the oldest setuptools version that we are compatible with.  In a few commits I updated the minimum version in `setup.py` and in the GitHub Actions matrix.

* We want at least 38.2.3, so we are sure to have support for wheels.
* We need at least 42.0.2, as earlier versions give ImportErrors in some of the Python versions we support.
* I settled on 49.0.0 for a minimum because this is the first version with proper support for PEP 496 environment markers, for example `demo==0.1; python_version < '3.9'`.  This setuptools version updates its vendored `packaging` version to one that correctly handles two-digit versions in `python_version`, so 3.10 and higher.  This fixed a last failing test.
* Incidentally, I saw that the test were 2-3 times slower with setuptools 60, so I changed the matrix to test with version 59 which fares a lot better.

## Packaging

* This adds a dependency on `packaging`, instead of the recently introduced `_compat.py` which tried to import this from a wide range of modules.

## Future PRs

Some ideas for near future improvements, in follow-up PRs:

* Run large scale automatic code modernisers: isort, black, pyupgrade.  I did not want to bloat the current PR with it.
* Get rid of `bootstrap.py` completely.  Some tests are still mentioning it, being executed only when this script is there.
* Run the `zc.recipe.egg` tests.  Either combined, or in separate jobs.  We do currently have the `bin/oltest` script (meant to run only tests that can be run when offline) and this does run the `zc.recipe.egg` tests.
* Get rid of the `bin/buildout setup` command and its ancient original name `runsetup`.  I have never used this, did not know it existed, and can see no use for it in these days.
* Maybe get rid of the `bin/buildout bootstrap` and `init` commands.  I don't think I ever used them.  I *did* see someone mentioning it this week, so maybe I am missing out on something. :-)  It does not seem to be much code, though I don't know how much test infrastructure is used by this.  Maybe fine to keep.
* Add `build` as dependency, and no longer call the `setup.py` script when doing editable installs (`[buildout] develop =`).  This is needed for editable installs of packages that only have a `pyproject.toml`.  I already install it when you run `make`, as we need it for generating the egg-info directory before creating the `bin/buildout` script in the repo.
* See if we can get rid of some of the patches.  Especially the setuptools `PackageIndex` patch is an ugly one that breaks every now and then. 

Please review the code, and also try it out locally.

@distributist Can you have a look as well, since you recently made fixes?


.github/workflows/scripts-setuptools-head.yml
```diff
@@ -11,9 +11,8 @@ jobs:
 
     runs-on: ubuntu-latest
     strategy:
-      max-parallel: 4
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13.0-rc.3']
```
dataflake, line 15: You don't need to specify the RC level. Just specifying "3.13" will get any latest available version, be it alpha, beta, RC or final.

dataflake:
That looks like a lot of work, thanks for that. I was convinced for a long time that zc.buildout is a dead end due to lack of development, this helps dispel that fear.

.github/workflows/scripts-setuptools-head.yml
mauritsvanrees: I did it like this because with 3.13 it did not work for me.

dataflake: We made the change in https://github.com/zopefoundation/meta/pull/239 but I forgot one detail, you also have to add `allow-prereleases: true` as an argument o the `setup-python` GHA action, which is safe to do because it only picks up prereleases if there is no final release.

mauritsvanrees: Thanks.  I have updated it to use `allow-prereleases: true`.

distributist:
@mauritsvanrees - I've gone through all the changes; I don't see anything that should have an impact on what I added. Thank you for slogging through all of this; it's definitely one of the least fun development tasks, and the time you've spent on it is greatly appreciated!

icemac:
Thank you for your time investment and the many simplifications. I just scrolled through the diff, so I did not add an approval.
I'd say: go for it as every combination is green now.

.github/workflows/scripts-setuptools-head.yml
icemac: 3.13.0 final is released now at least in a couple of days it should be usable in GHA.

mauritsvanrees:
I have created a branch 3.x from current master.

And I have merged master into this PR.   But I get [errors on Windows](https://github.com/buildout/buildout/actions/runs/11437173298/job/31818886720#step:5:4650) now, like this:

```
File "d:\a\buildout\buildout\src\zc\buildout\tests\test_all.py", line 1741, in zc.buildout.tests.test_all.abnormal_exit
Failed example:
    print_(system(buildout), end='')
Expected:
    Develop: '/sample-buildout/recipes'
    Updating p1.
    Updating p2.
    Installing p3.
Got:
    Develop: 'C:/Program /sample-buildout/recipes'
    Updating p1.
    Updating p2.
    Installing p3.
```

That may be fixable with some test output normalisation, although it should already be in place.  But it points to a real problem down the road:

```
File "d:\a\buildout\buildout\src\zc\buildout\tests\test_all.py", line 1747, in zc.buildout.tests.test_all.abnormal_exit
Failed example:
    print_(system(buildout+' buildout:parts='), end='')
Expected:
    Develop: '/sample-buildout/recipes'
    Uninstalling p2.
    Uninstalling p1.
Got:
    'C:/Program' is not recognized as an internal or external command,
    operable program or batch file.
```

It could be something in the merged PR #628, but those changes look unrelated at first glance.

gotcha:
@mauritsvanrees This is a lot of arduous work that will help keeping buildout useful. Thanks a lot.

mauritsvanrees:
All is green again, so I merge.  Thanks for the reviews and supportive comments.

mauritsvanrees:
There is lots more that could be done, see the "Future PRs" section in my original comment above, but for now this is ready.
I have released version 4.0.0a1 with this.

icemac:
@mauritsvanrees Thank you for your invested time. I just started to try it out in our company.

PR #668, opened 2024-10-21
WIP: fix Windows test failure

mauritsvanrees:
See https://github.com/buildout/buildout/pull/667#issuecomment-2426501060


gotcha:
Did you check if it could be due to a change in setup-python Github Action ? AFAICS specifying v5 means pointing at a moving target.

IOW, I would try with `setup-python@v5.1.1`

gotcha:
Trying to think further, the windows tests run on "windows-latest". That might also be a moving target.

Actually, I guess from https://github.com/actions/runner-images?tab=readme-ov-file#available-images that a new windows-latest image was deployed on October 15.

If this is actually the issue, it might be worth keeping some jobs on circle ci or elsewhere so that the chances that they break at the same time would be reduced and thus hint at a breaking change in the CI infrastructure rather than a new bug.

mauritsvanrees:
The temporary directory used to be in `C:\Users\runneradmin\AppData\Local\Temp` and now it is in `C:\Program Files\Git\AppData\Local\Temp` even though we try to set it in GitHub Actions.  I am trying to work around that.

Also, in some tests we call `system(buildout+ ' annotate')` which used to work but now it fails and needs to be `system([buildout, 'annotate'])`, although then we need to pass `shell=False`.

mauritsvanrees:
We can actually set the TMPDIR to C:\tmp and avoid most problems...
I am making a new PR.

mauritsvanrees:
For completeness:

> IOW, I would try with `setup-python@v5.1.1`

I tried this, but it did not help.

> Trying to think further, the windows tests run on "windows-latest". That might also be a moving target.
> 
> Actually, I guess from https://github.com/actions/runner-images?tab=readme-ov-file#available-images that a new windows-latest image was deployed on October 15.

True, but there was no major bump: least windows-latest was and is the same as windows-2022.  I tried windows-2019, but that failed in the same way.

> If this is actually the issue, it might be worth keeping some jobs on circle ci or elsewhere so that the chances that they break at the same time would be reduced and thus hint at a breaking change in the CI infrastructure rather than a new bug.

Problem with this would be that we would need to keep two separate CI systems running.  The benefit could indeed be that if the tests still pass on one system, you would know that the test failures on the other system are likely not a problem due to changes in the Buildout code itself.  But you still need to figure out what the problem actually is.
And in this case: we were not running any Windows tests on Circle CI (this was already the case before my major changes earlier this year).

gotcha:
Any idea why it started to fail ?

Thanks for fixing the bugs. And for your detailed answer.

mauritsvanrees:
> Any idea why it started to fail ?

This is a line from the last [successful run](https://github.com/buildout/buildout/actions/runs/11212561835/job/31165087698#step:5:235) of two weeks ago:

    Processing c:\users\runneradmin\appdata\local\temp\tmpdm9uimmcget_dist\demoneeded-1.1.zip

And this is a similar line from the first [unsuccessful run](https://github.com/buildout/buildout/actions/runs/11437173298/job/31818886720#step:5:3007) yesterday:

    Processing c:\program files\git\appdata\local\temp\tmpsvaywm48get_dist\demoneeded-1.1.zip

It seems the `HOME` environment variable got changed on the Windows machine, resulting in the temporary directories getting generated in a path with a space, which tripped us up.

icemac:
After the PR is closed now, do we still need the branch?

mauritsvanrees:
> After the PR is closed now, do we still need the branch?

Theoretically yes, if someone wants to follow along my train of thought in how I tried fixing it in each new commit.  But there are perhaps more interesting ways of spending your time. ;-)

I have deleted the branch.

PR #669, opened 2024-10-22
WIP Fix Windows test failures, take two

mauritsvanrees:
Due to some change in the Windows environment on GitHub, the temporary directory ends up somewhere in `C:\Program Files`, which messes things up due to the space.  Setting the TMPDIR to `C:\tmp` fixes this.  We also need to turn the command that we pass to our `system` function into a list when it is not simply `buildout`.  Actually, maybe that is no longer needed after getting rid of that space.  Trying that out here.
As far as I see, this is only a problem due to doctest output normalisation: the buildout commands on Windows work just fine when they are in a directory with a space.


mauritsvanrees:
This has worked, and I have cherry-picked the needed commits into PR #667.

PR #671, opened 2024-12-06
Test with latest setuptools 75.6.0 and pip 24.3.1. Drop Python 3.8.

mauritsvanrees:


mauritsvanrees:
I had to drop Python 3.8 support due to a failure to install the latest setuptools.  See [this test run](https://github.com/buildout/buildout/actions/runs/12206355882/job/34056007822?pr=671#step:5:35).

dataflake:
I find it problematic to integrate the drop of Python 3.8 when the entire range of Zope Foundation projects still supports it.

mauritsvanrees:
> I find it problematic to integrate the drop of Python 3.8 when the entire range of Zope Foundation projects still supports it.

Fair point.  Note that the same is already true for setuptools, which is why I did it here: 75.3.0 is the last that runs on Python 3.8.

I have updated the GHA yaml files and am testing on 3.8 again.  The set of jobs where we test all Python versions now uses setuptools 75.3.0.

icemac:
Sorry for being late to the party: Maybe it is time to start dropping Python 3.8 support as it was sunsetted in October this year. Some dependencies already started to drop 3.8 support.

I'll create an issue in the Zope repository to discuss it there.
Let's keep 3.8 support here for now.

.github/workflows/scripts-setuptools-head.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
```
icemac, line 15: Here 3.8 did not get restored.

icemac (commented):
Might we need a change log entry to be able to have a new release, so we can change our `setuptools  < 74` requirement set via `meta/confg` to a newer `setuptools` version?

icemac (approved):

.github/workflows/scripts-setuptools-head.yml
mauritsvanrees: Correct: this uses setuptools HEAD, so it will fail with 3.8.

mauritsvanrees:
> Might we need a change log entry to be able to have a new release, so we can change our `setuptools < 74` requirement set via `meta/confg` to a newer `setuptools` version?

I did not make any changes to the code.  I just updated the last tested version of setuptools from 75.2 to 75.6.  So no release needed.

PR #672, opened 2024-12-18
Add change log entry for #671

icemac:
Reading https://github.com/zopefoundation/zope.interface/issues/333 I saw the missing change log entry for #671.

Askinfg for review here as I am not sure about the naming of the file.


mauritsvanrees:
Looking at the currently available [towncrier suffixes](https://github.com/buildout/buildout/blob/master/pyproject.toml#L23) I would say: rename it to `671.develop`.

icemac:
Thank you!

PR #674, opened 2024-12-18
Drop Python 3.8 support.  Require 3.9 as minimum.

mauritsvanrees:


dataflake (approved):

news/38.breaking
icemac, line 1: The part before the dot should be the ticket number, shouldn't it?

zc.recipe.egg_/CHANGES.rst
```diff
@@ -4,7 +4,7 @@ Change History
 3.0.0 (unreleased)
 ==================
 
-- Drop support for Python 2.  Require Python 3.8 as minimum.
+- Drop support for Python 2.  Require Python 3.9 as minimum.
   Require ``zc.buildout`` 3.1.0 as minimum.
```
icemac, line 8: Should we also have a release of this subpackage?

icemac (approved):

zc.recipe.egg_/CHANGES.rst
mauritsvanrees: It could be good, though it seems not strictly necessary.  But we are currently not actually running the tests of `zc.recipe.egg`.  And `bin/oltest` which *does* run these tests (plus the buildout ones) fails locally.  So there is some work to do still before we can really release that one.

mauritsvanrees:
I have started a PR for improving the `zc.recipe.egg` testing, but it is just a start and I won't continue on that currently:
https://github.com/buildout/buildout/pull/675

PR #675, opened 2024-12-19
Test zc.recipe.egg

mauritsvanrees:
We have not run the `zc.recipe.egg` tests in a long time, maybe even years.  And they fail locally.
This is work in progress, help needed, and I probably won't continue on it this year.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,13 @@ bin/test: bin/buildout buildout.cfg
 test: bin/test
 	PYTHONWARNINGS=ignore bin/test -pvc
 
+test-recipe: bin/test
+	PYTHONWARNINGS=ignore bin/test-recipe
+
+# oltest = offline test.  But this needs test fixes.
+oltest: bin/test
+	PYTHONWARNINGS=ignore bin/oltest
+
 help:
 	./prepare.sh --help
 
--- a/Makefile
+++ b/Makefile
@@ -13,10 +13,6 @@ test: bin/test
 test-recipe: bin/test
 	PYTHONWARNINGS=ignore bin/test-recipe
 
-# oltest = offline test.  But this needs test fixes.
-oltest: bin/test
-	PYTHONWARNINGS=ignore bin/oltest
-
 help:
 	./prepare.sh --help
 
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -1,6 +1,6 @@
 [buildout]
 develop = zc.recipe.egg_ .
-parts = test oltest py
+parts = test oltest test-recipe py
 versions = versions
 
 [versions]
@@ -41,6 +41,10 @@ initialization =
   path_to_coveragerc = os.path.join('${buildout:directory}', '.coveragerc')
   setup_coverage(path_to_coveragerc)
 environment = test-environment
+
+[test-recipe]
+<= test
+defaults = ['-s', 'zc.recipe.egg']
 
 # Tests that can be run wo a network
 [oltest]
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -1,6 +1,6 @@
 [buildout]
 develop = zc.recipe.egg_ .
-parts = test oltest test-buildout test-recipe py
+parts = test test-buildout test-recipe py
 versions = versions
 
 [versions]
@@ -49,15 +49,3 @@ defaults = ['-s', 'zc.buildout']
 [test-recipe]
 <= test
 defaults = ['-s', 'zc.recipe.egg']
-
-# Tests that can be run wo a
-# TODO This currently fails.
-[oltest]
-<= test
-recipe = zc.recipe.testrunner
-eggs = ${test:eggs}
-defaults =
-  [
-  '-t',
-  '!(bootstrap|selectingpython|selecting-python)',
-  ]
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -1,6 +1,6 @@
 [buildout]
 develop = zc.recipe.egg_ .
-parts = test oltest test-recipe py
+parts = test oltest test-buildout test-recipe py
 versions = versions
 
 [versions]
@@ -34,7 +34,7 @@ eggs =
   zc.buildout[test]
   zc.recipe.egg
   coverage
-defaults = ['-s', 'zc.buildout']
+defaults = ['-s', 'zc.buildout', '-s', 'zc.recipe.egg']
 initialization =
   import os.path
   from zc.buildout.testing import setup_coverage
@@ -42,11 +42,16 @@ initialization =
   setup_coverage(path_to_coveragerc)
 environment = test-environment
 
+[test-buildout]
+<= test
+defaults = ['-s', 'zc.buildout']
+
 [test-recipe]
 <= test
 defaults = ['-s', 'zc.recipe.egg']
 
-# Tests that can be run wo a network
+# Tests that can be run wo a
+# TODO This currently fails.
 [oltest]
 <= test
 recipe = zc.recipe.testrunner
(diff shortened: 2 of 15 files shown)


icemac:
I was always wondering why `zc.recipe.egg` lives inside `zc.buildout`.
Do we have a chance to move `zc.recipe.egg` into its own repository or is it so deeply coupled with the repository of `zc.buildout` so that is not possible?

mauritsvanrees:
I suppose this was done to make it easier to check that a change in `zc.buildout` works together with the latest `zc.recipe.egg`.  And if this shows that changes are needed in `zc.recipe.egg`, then it is easier to have it in the same repo, instead of having to clone another repo, and probably a specific branch.  So for now I would keep it that way.

dataflake:
Reality check: There has not been any release of `zc.recipe.egg` since July 2018. So it seems there's no good reason to keep them together. I'm not saying "this needs to be done now", just that there's no good reason to keep it the way it is.

mauritsvanrees:
I do wonder if changes will be needed there when/if we start messing with low level stuff like supporting projects that have no `setup.py` but only a `pyproject.toml`.

dataflake:
Would it still not be like any other dependency relationship that can be handled by pinning a minimum version? Just wondering.

mauritsvanrees:
I am actually booking progress here.  I have updated `bin/test` to run both the `zc.buildout` and `zc.recipe.egg` tests, so you can see what works and what does not work.

I have now reached a point where the tests with the oldest supported `setuptools` (49.0.0) [pass](https://github.com/buildout/buildout/actions/runs/14107093863/job/39516043914?pr=675)!  But with all other setuptools versions, they [fail](https://github.com/buildout/buildout/actions/runs/14107093863/job/39516043947?pr=675).

Sample error in `README.rst`:

```
File "/home/runner/work/buildout/buildout/zc.recipe.egg_/src/zc/recipe/egg/README.rst", line 262, in README.rst
Failed example:
    print_(system(buildout+' -o'), end='')
Expected:
    Updating demo.
Got:
    Updating demo.
    <BLANKLINE>
    <BLANKLINE>
    While:
      Updating demo.
      Getting distribution for 'demoneeded'.
    Error: We don't have a distribution for demoneeded
    and can't install one in offline (no-install) mode.
    <BLANKLINE>
```


Sample error in `working_set_caching.rst `:

```
File "/home/runner/work/buildout/buildout/zc.recipe.egg_/src/zc/recipe/egg/working_set_caching.rst", line 59, in working_set_caching.rst
Failed example:
    recipe._working_set(**ws_args_1)
Exception raised:
      File "/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/doctest.py", line 1350, in __run
        exec(compile(example.source, filename, "single",
      File "<doctest working_set_caching.rst[18]>", line 1, in <module>
        recipe._working_set(**ws_args_1)
      File "/home/runner/work/buildout/buildout/zc.recipe.egg_/src/zc/recipe/egg/egg.py", line 129, in _working_set
        ws = zc.buildout.easy_install.working_set(
      File "/home/runner/work/buildout/buildout/src/zc/buildout/easy_install.py", line 1180, in working_set
        return install(specs, None, path=path)
      File "/home/runner/work/buildout/buildout/src/zc/buildout/easy_install.py", line 1011, in install
      File "<doctest working_set_caching.rst[13]>", line 3, in new_install
        return old_install(*args, **kwargs)
      File "/home/runner/work/buildout/buildout/src/zc/buildout/easy_install.py", line 784, in install
      File "/home/runner/work/buildout/buildout/src/zc/buildout/easy_install.py", line 593, in _get_dist
        raise zc.buildout.UserError(
    zc.buildout.UserError: We don't have a distribution for demoneeded
    and can't install one in offline (no-install) mode.
```

These errors all have to do with offline mode.  The [reference documentation](https://github.com/buildout/buildout/blob/4.1.4/doc/reference.rst?plain=1#L442-L462) says about offline mode:

> Warning: Offline mode is deprecated.
> 
> Its purpose has evolved over time and the end result doesn't make much sense, but it is retained (indefinitely) for backward compatibility.
> 
> If you think you want an offline mode, you probably want either the non-newest mode or the install-from-cache mode instead.

With that in mind, I think we should remove the tests for offline mode, at least the ones that currently fail.  I have no idea why they would pass in setuptools 49 and fail in later versions.

There are a few errors in `custom.rst` as well, but they seem to have a different cause.

mauritsvanrees:
This is ready for review now.
I have rebased on master and force pushed.  Let me copy the changelog snippets as summary.

For zc.buildout:

- While creating sample packages for testing, mostly create wheels instead of eggs.
  For the sample source distributions, create ``tar.gz`` instead of ``zip`` files.
 Then our package index for testing is more like the actual PyPI.

For zc.recipe.egg:


- Drop support for Python 2.  Require Python 3.9 as minimum.  [This was already done earlier on the master branch.]
  Require ``zc.buildout`` 4.0.0 as minimum.

- Removed tests for the 'offline' mode of buildout, which were broken since setuptools 59.
  This option was deprecated long time ago, and its current working is not defined.
  See the [reference documentation](https://www.buildout.org/en/latest/reference.html), which adds:
  "If you think you want an offline mode, you probably want either the non-newest mode or the install-from-cache mode instead."

- Removed tests for demoing usage of a custom egg, which were broken since setuptools 49.6.0.
  It is not clear what this means for how well custom egg creation currently works, but the remaining base tests still pass.
  This is when you are using ``recipe = zc.recipe.egg:custom`` or ``recipe = zc.recipe.egg:develop``, which should be rare.

When merged, I can make a `zc.recipe.egg` 3.0.0 release.

mauritsvanrees:
All green!

dataflake (approved):
Looking good as far as I can tell

mauritsvanrees:
I have released `zc.buildout` 4.1.6 with the testing change.

But I cannot make a PyPI release of `zc.recipe.egg`.
I have run `fullrelease` to create the [`zc.recipe.egg-3.0.0` tag](https://github.com/buildout/buildout/tree/zc.recipe.egg-3.0.0) and push the changes to GitHub, so that part is handled.  But PyPI says no:

> Response status code: 403
> Reason: The user 'maurits' isn't allowed to upload to project 'zc.recipe.egg'. See https://pypi.org/help/#project-name for more information.
> There were errors or warnings.
> ERROR Package upload has failed.

Some data:

* My PyPI user name is `maurits`
* I am in the PyPI maintainers list of [`zc.buildout`](https://pypi.org/project/zc.buildout/), but not of [`zc.recipe.egg`](https://pypi.org/project/zc.recipe.egg/).
* zc.recipe.egg is owned by the PyPI `zope` organisation.
* I have the Manager role in that organisation, but that is apparently not enough.
* I am also the single member of the Release Managers team there, but that team only has extra rights on 9 repositories, and zc.recipe.egg is not one of them.

@icemac @dataflake Can one of you grant me or the Release Managers team more rights on zc.recipe.egg?  I suppose it would work to add the zc.recipe.egg project to the Release Managers team.  Maybe zc.buildout too.  I don't know how that works, the other two PyPI organisations that I am a member of, have no teams.

dataflake:
I have added the Release Managers team to zc.recipe.egg as Maintainer. Can you try again?

dataflake:
(I have no rights on the zc.buildout project and it is not part of the zope organization account, only @icemac could add it)

mauritsvanrees:
Yes, that worked:

https://pypi.org/project/zc.recipe.egg/3.0.0/


Thanks.

icemac:
@dataflake I added `zc.buildout` to `zope` organization on PyPI.

Issue #679, opened 2025-03-04
Version `4.1` needs current `setuptools`


petschki:
I've updated my `zc.buildout==4.1` with `setuptools==75.6.0` and get the following traceback:

```
Traceback (most recent call last):
  File "/Users/peterm/.pyenv/versions/3.13.2/envs/py-3.13-project/lib/python3.13/site-packages/zc/buildout/buildout.py", line 2281, in main
    buildout = Buildout(config_file, options,
                        use_user_defaults, command, args)
  File "/Users/peterm/.pyenv/versions/3.13.2/envs/py-3.13-project/lib/python3.13/site-packages/zc/buildout/buildout.py", line 413, in __init__
    ).version),
      ^^^^^^^
AttributeError: 'NoneType' object has no attribute 'version'
```

with current `setuptools==75.8.2` it works.

MacOS: 15.3.1
Python: 3.13.2

mauritsvanrees:
Should be fixed with PR #680.

But we may want to add a note to the changelog, advising people to use `setuptools` 75.8.2.  Its [`pkg_resources.working_set.find` method](https://github.com/pypa/setuptools/blob/8e39bc907b93bd5f89c0361c66f60a9c9f983c37/pkg_resources/__init__.py#L701) uses multiple spellings of the requirement name to search for an installed distribution:

```
        candidates = (
            req.key,
            self.normalized_to_canonical_keys.get(req.key),
            safe_name(req.key).replace(".", "-"),
        )
```

PR #680, opened 2025-03-04
Fix error adding minimum ``zc.buildout`` version as requirement.

mauritsvanrees:
Fixes https://github.com/buildout/buildout/issues/679

Also, previously the code was trying to get a distribution for `zc.buildout` even when it was not needed, that is when `zc.buildout` already had a version pin.

### DIFF
--- a/news/679.bugfix
+++ b/news/679.bugfix
@@ -1,0 +1,2 @@
+Fix error adding minimum ``zc.buildout`` version as requirement.
+[maurits]
--- a/src/zc/buildout/buildout.py
+++ b/src/zc/buildout/buildout.py
@@ -403,19 +403,45 @@ class Buildout(DictMixin):
             versions = data[versions_section_name]
         else:
             versions = {}
-        versions.update(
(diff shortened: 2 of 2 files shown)


dataflake:
This setuptools version issue is hitting us in production. If you can get this merged and released quickly please do. I guess that means we will be forced to update every single zopefoundation package because we do not pin zc.buildout, but setuptools.

mauritsvanrees:
@dataflake If you pin `setuptools==75.8.2` in your environments, it should work as well.

@icemac or anyone else: can I get an approving review?

BTW, I see a few more lines in the code where we may need to do something similar, for example when getting a distribution for a recipe.  But I did not see reports about it going wrong there yet.  Also there, latest setuptools 75.8.2 could help.  When similar reports come in, we should ask the reporter to install latest setuptools and try again.

dataflake (approved):

mauritsvanrees:
Technically a review by someone with write access is required.  I chose to circumvent this.

I have released `zc.buildout` 4.1.1 with this fix.

dataflake:
Thank you!

Issue #681, opened 2025-03-05
AttributeError: 'NoneType' object has no attribute 'location'


mauritsvanrees:
As reported by @datafake in a [GenericSetup PR](https://github.com/zopefoundation/Products.GenericSetup/pull/135#issuecomment-2700074336)

Using latest `zc.buildout` 4.1.1 and slightly older `setuptools` 75.6.0:

```
Traceback (most recent call last):
  File "/home/runner/work/Products.GenericSetup/Products.GenericSetup/.tox/py39/lib/python3.9/site-packages/zc/buildout/buildout.py", line 2309, in main
    getattr(buildout, command)(args)
  File "/home/runner/work/Products.GenericSetup/Products.GenericSetup/.tox/py39/lib/python3.9/site-packages/zc/buildout/buildout.py", line 717, in install
    self._maybe_upgrade()
  File "/home/runner/work/Products.GenericSetup/Products.GenericSetup/.tox/py39/lib/python3.9/site-packages/zc/buildout/buildout.py", line 1165, in _maybe_upgrade
    if not inspect.getfile(sys.modules[project]).startswith(dist.location):
AttributeError: 'NoneType' object has no attribute 'location'
```

PR #682, opened 2025-03-05
Fix error finding zc.buildout dist in upgrade check.

mauritsvanrees:
This depends on your `setuptools` version.
Additionally, check if we need to upgrade due to the new `packaging` dependency.

Fixes https://github.com/buildout/buildout/issues/681


icemac (approved):
LGTM – I've probably already seen a place where this change is needed.

Issue #683, opened 2025-03-05
Problems loading entry points


mauritsvanrees:
Buildout uses entry points to handle recipes and extensions.  This fails, depending on your setuptools version.  
See this [GenericSetup PR](https://github.com/zopefoundation/Products.GenericSetup/pull/135#issuecomment-2701175834), using latest `zc.buildout` 4.1.1 and slightly older `setuptools` 75.6.0:

```
While:
  Installing.
  Getting section test.
  Initializing section test.
  Loading zc.buildout recipe entry zc.recipe.testrunner:default.

An internal error occurred due to a bug in either zc.buildout or in a
recipe being used:
Traceback (most recent call last):
  File "/Users/maurits/community/plone-coredev/6.2/src/Products.GenericSetup/.tox/py313/lib/python3.13/site-packages/zc/buildout/buildout.py", line 2325, in main
    getattr(buildout, command)(args)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/Users/maurits/community/plone-coredev/6.2/src/Products.GenericSetup/.tox/py313/lib/python3.13/site-packages/zc/buildout/buildout.py", line 753, in install
    [self[part]['recipe'] for part in install_parts]
     ~~~~^^^^^^
  File "/Users/maurits/community/plone-coredev/6.2/src/Products.GenericSetup/.tox/py313/lib/python3.13/site-packages/zc/buildout/buildout.py", line 1401, in __getitem__
    options._initialize()
    ~~~~~~~~~~~~~~~~~~~^^
  File "/Users/maurits/community/plone-coredev/6.2/src/Products.GenericSetup/.tox/py313/lib/python3.13/site-packages/zc/buildout/buildout.py", line 1513, in _initialize
    self.initialize()
    ~~~~~~~~~~~~~~~^^
  File "/Users/maurits/community/plone-coredev/6.2/src/Products.GenericSetup/.tox/py313/lib/python3.13/site-packages/zc/buildout/buildout.py", line 1519, in initialize
    recipe_class = _install_and_load(reqs, 'zc.buildout', entry, buildout)
  File "/Users/maurits/community/plone-coredev/6.2/src/Products.GenericSetup/.tox/py313/lib/python3.13/site-packages/zc/buildout/buildout.py", line 1473, in _install_and_load
    return pkg_resources.load_entry_point(
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        req.project_name, group, entry)
[...]

PR #684, opened 2025-03-05
Patch the find method from pkg_resources.WorkingSet.

mauritsvanrees:
Let this use the code from `setuptools` 75.8.2, if the currently used version is older. This is better at finding installed distributions.

Fixes https://github.com/buildout/buildout/issues/683

### DIFF
--- a/news/682.bugfix
+++ b/news/682.bugfix
@@ -1,4 +1,5 @@
 Patch the ``find`` method from ``pkg_resources.WorkingSet``.
 Let this use the code from ``setuptools`` 75.8.2, if the currently used version is older.
 This is better at finding installed distributions.
+But don't patch ``setuptools`` versions older than 61: the new version of the method would give an error there.
 [maurits]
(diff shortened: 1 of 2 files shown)


icemac:
Hm, 49.0.0 seems to be too old for this approach – time to drop support for it?

mauritsvanrees:
> Hm, 49.0.0 and 59.8.0 seem to be too old for this approach – time to drop support for them?

I have pushed a commit that fixes the AttributeError.  When setuptools is older than 61, do not apply the patch.  Of course then on older versions you may run into problems exactly because the patch is not there.

For me it would be fine to drop support for `setuptools` versions older than 61.  But if my last commit works, it may be okay to keep them.  I have updated the changelog so people may see this.

mauritsvanrees:
The tests failed on Python 3.9 due to a problem with type hints.  So I have removed the type hints.  No sense in spending time on this.

I have rebased on master.

mauritsvanrees:
It's green!  Please review.

dataflake (approved):
I don't have much knowledge about setuptools internals, but if this works and the tests are green...

mauritsvanrees:
I have released this in 4.1.3.

Issue #686, opened 2025-03-05
Scripts with recent binary wheels may have unfindable distributions


mauritsvanrees:
TLDR: for now don't use setuptools 75.8.1/2 to build binary wheels for namespace packages.

Pfff, where to start?  I get this error:

```
$ bin/fullrelease
Traceback (most recent call last):
  File "/Users/maurits/.local/share/uv/python/cpython-3.13.2-macos-x86_64-none/lib/python3.13/importlib/metadata/__init__.py", line 407, in from_name
    return next(iter(cls.discover(name=name)))
StopIteration

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/maurits/tmp/test/bin/fullrelease", line 36, in <module>
    import zest.releaser.fullrelease
  File "/Users/maurits/shared-eggs/cp313/zest_releaser-9.4.0-py3.13.egg/zest/releaser/__init__.py", line 11, in <module>
    __version__ = version("zest.releaser")
  File "/Users/maurits/.local/share/uv/python/cpython-3.13.2-macos-x86_64-none/lib/python3.13/importlib/metadata/__init__.py", line 987, in version
    return distribution(distribution_name).version
           ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/Users/maurits/.local/share/uv/python/cpython-3.13.2-macos-x86_64-none/lib/python3.13/importlib/metadata/__init__.py", line 960, in distribution
    return Distribution.from_name(distribution_name)
           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/Users/maurits/.local/share/uv/python/cpython-3.13.2-macos-x86_64-none/lib/python3.13/importlib/metadata/__init__.py", line 409, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: No package metadata was found for zest.releaser
```

This is with latest zc.buildout, setuptools, zest.releaser, Python 3.13.

Trying it differently in a Python interpreter generated by zc.recipe.egg:

```
>>> from importlib.metadata import version
>>> version("zest.releaser")
Traceback (most recent call last):
  File "/Users/maurits/.local/share/uv/python/cpython-3.13.2-macos-x86_64-none/lib/python3.13/importlib/metadata/__init__.py", line 407, in from_name
    return next(iter(cls.discover(name=name)))
StopIteration

During handling of the above exception, another exception occurred:

[...]

PR #687, opened 2025-03-06
If needed, copy and rename wheels before making an egg out of them.

mauritsvanrees:
This helps for wheels of namespace packages created with `setuptools` 75.8.1 or higher. For namespace package we need a dot instead of an underscore in the resulting egg name.

Fixes https://github.com/buildout/buildout/issues/686


mauritsvanrees:
I tested this in Plone's `buildout.coredev` by repeatedly removing the generated egg of `zest.releaser` 9.4.0, and of `check-manifest` 0.50.  Both have an underscore in the wheel name.  This PR only makes a difference for `zest.releaser`, because for check-manifest there is no problem to begin with.

dataflake (approved):
Looks like it worked during your tests, so by all means it should be good to go.

PR #690, opened 2025-03-20
Implements PEP 503

### LINKED ISSUE #634: PEP 503 has not been implemented
I found out that PEP 503 ( https://peps.python.org/pep-0503/ ) was not implemented, In particular the this section:

> Repositories **MAY** redirect unnormalized URLs to the canonical normalized URL (e.g. /Foobar/ may redirect to /foobar/), however clients MUST NOT rely on this redirection and **MUST request the normalized URL**.

In my scenario I use zc.buildout with a gitlab pypi package repository as index.
In the documentation they claim to accept not-normalized forms of package names (as pypi does) but it in reality they do not: both from browser and from buildout, namspace-package works, namepsace.package dont.

I thought I could open a ticket to GITLAB for the pypi repo, but they MAY redirect unnormalized url, setuptools on the other hand, clients MUST request the normalized url.

It has already been asked various times both on setuptools and pkg_resources but with no solution in sight.

- https://github.com/pypa/setuptools/issues/991
- https://github.com/pypa/setuptools/issues/3797
- https://github.com/pypa/pkg_resources/issues/13

I did a minimal POC to implement the PEP but I really do not know if it could have other impacts.

My workaround would be to patch pkg_resources.safe_name removing the "."  from the regex that defines allowed package name chars in order to get also the dots replaced with dashes. 
Here it is the original method

```python
def safe_name(name):
    """Convert an arbitrary string to a standard distribution name

    Any runs of non-alphanumeric/. characters are replaced with a single '-'.
    """
    return re.sub('[^A-Za-z0-9.]+', '-', name)
```

Last line would be (dot removed from regexp after '9')
```python
    return re.sub('[^A-Za-z0-9]+', '-', name)
```

Given the fact that looks like the aforementioned issues will not be solved any time soon, maybe the method could be monkey-patched in zc.buildout as we do for process_url. If it's not the case I'm open to new ideas.

andreclimaco:
fixes #634 

@mauritsvanrees

### DIFF
--- a/src/zc/buildout/patches.py
+++ b/src/zc/buildout/patches.py
@@ -11,7 +11,6 @@
 # FOR A PARTICULAR PURPOSE.
 #
 ##############################################################################
-import re
 
 def patch_Distribution():
     try:
@@ -193,27 +192,7 @@ def patch_PackageIndex():
         if url.startswith(self.index_url) and getattr(f, 'code', None) != 404:
             page = self.process_index(url, page)
 
-    # method copied over from setuptools 46.1.3
-    # Unchanged in setuptools 77.0.1.
-    # Implements PEP 503
-    def find_packages(self, requirement):
-        url_name = re.sub(r"[-_.]+", "-", requirement.unsafe_name).lower()
-        self.scan_url(self.index_url + url_name + '/')
-
-        if not self.package_pages.get(requirement.key):
-            # Fall back to safe version of the name
-            self.scan_url(self.index_url + requirement.project_name + '/')
-
-        if not self.package_pages.get(requirement.key):
-            # We couldn't find the target package, so search the index page too
-            self.not_found_in_index(requirement)
-
-        for url in list(self.package_pages.get(requirement.key, ())):
-            # scan each page that might be related to the desired package
-            self.scan_url(url)
-
     setattr(PackageIndex, 'process_url', process_url)
-    setattr(PackageIndex, 'find_packages', find_packages)
 
 
 patch_PackageIndex()
@@ -325,7 +304,7 @@ patch_pkg_resources_requirement_contains()
 
 
 def patch_pkg_resources_working_set_find():
-    """Patch pkg_resources.WorkingSet find method.
+    """Patch pkg_resources.patch_find_packages find method.
 
     setuptools 75.8.1 fixed wheel file naming to follow the binary distribution
     specification.  This broke a lot for us, especially when using editable
@@ -409,3 +388,38 @@ def patch_pkg_resources_working_set_find():
 
 
 patch_pkg_resources_working_set_find()
+
+
+def patch_find_packages():
+    """
+    Patch setuptools.package_index.PackageIndex find_packages method.
+    Implements PEP 503
+    """
+    try:
+        from setuptools.package_index import PackageIndex
+        import re
+    except ImportError:
+        return
+
+    # method copied over from setuptools 46.1.3
+    # Unchanged in setuptools 77.0.1.    
+    def find_packages(self, requirement):
+        url_name = re.sub(r"[-_.]+", "-", requirement.unsafe_name).lower()
+        self.scan_url(self.index_url + url_name + '/')
+
+        if not self.package_pages.get(requirement.key):
+            # Fall back to safe version of the name
+            self.scan_url(self.index_url + requirement.project_name + '/')
+
+        if not self.package_pages.get(requirement.key):
+            # We couldn't find the target package, so search the index page too
+            self.not_found_in_index(requirement)
+
+        for url in list(self.package_pages.get(requirement.key, ())):
+            # scan each page that might be related to the desired package
+            self.scan_url(url)
+    
+    setattr(PackageIndex, 'find_packages', find_packages)
+
+
+patch_find_packages()
--- a/src/zc/buildout/patches.py
+++ b/src/zc/buildout/patches.py
@@ -11,6 +11,7 @@
 # FOR A PARTICULAR PURPOSE.
 #
 ##############################################################################
+import re
 
 def patch_Distribution():
     try:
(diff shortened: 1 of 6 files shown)


mauritsvanrees (commented):
I did not try it yet, but this looks safe.  The original method in setuptools has indeed not changed in years.

Can you do this patch in a new function though, instead of adding it to `patch_PackageIndex`?  Maybe `patch_find_packages`.  Otherwise if any of the about 10 imports at the top of `patch_PackageIndex` fails, the patch won't apply.

mauritsvanrees:
Also, this needs a small fix in the [tests](https://github.com/buildout/buildout/actions/runs/13960365674/job/39144134678?pr=690#step:5:320), where there now is one extra line of output, which makes sense here.

andreclimaco:
@mauritsvanrees I created a new function and updated the tests.

src/zc/buildout/patches.py
```diff
@@ -304,7 +304,7 @@ def __contains__(self, item):
 
 
 def patch_pkg_resources_working_set_find():
-    """Patch pkg_resources.WorkingSet find method.
+    """Patch pkg_resources.patch_find_packages find method.
```
mauritsvanrees, line 307:
```suggestion
    """Patch pkg_resources.WorkingSet find method.
```

src/zc/buildout/tests/buildout.txt
```diff
@@ -3374,7 +3374,7 @@ or paths to use::
 
     >>> cd(sample_bootstrapped)
     >>> remove('setup.cfg')
-    >>> print_(system([buildout, '-csetup.cfg', 'init', 'demo', 'other', './src']), end='')
+    >>> print_(system([buildout, '-csetup.cfg', 'init', 'demo', 'other', './src']), end='')  # doctest: +ELLIPSIS -DONT_ACCEPT_BLANKLINE
```
mauritsvanrees, line 3377: Can you remove all additions of `# doctest: +ELLIPSIS -DONT_ACCEPT_BLANKLINE`?
When you look at `test_all.py`, the `ELLIPSIS` optionflag is already passed to all test cases, so there is no need to add it.  Yes, I see we already do that a lot, but let's not needlessly add it.

And `-DONT_ACCEPT_BLANKLINE` would set `DONT_ACCEPT_BLANKLINE` to False, but this is already the default in Python, so I don't see why this would be needed.  If I understand the [docs](https://docs.python.org/3/library/doctest.html#doctest.DONT_ACCEPT_BLANKLINE) correctly, this option only needs to be set if you need to check if the output *literally* contains the string `<BLANKLINE>`.

src/zc/buildout/tests/buildout.txt
```diff
@@ -3388,7 +3388,7 @@ or paths to use::
     Got demoneeded 1.1.
     Generated script '/sample-bootstrapped/bin/demo'.
     Generated script '/sample-bootstrapped/bin/distutilsscript'.
-    Generated interpreter '/sample-bootstrapped/bin/py'.
+    Generated interpreter '/sample-bootstrapped/bin/py'...
```
mauritsvanrees, line 3391: I don't like adding `...` at the end of a doctest example.  This would mask any errors  that happen at the end.  Can you remove those, also from the end of the other examples?

src/zc/buildout/tests/update.txt
```diff
@@ -275,11 +273,11 @@ gives us a sys.exit:
...
-    >>> print_(system(buildout, with_exit_code=True), end='')
+    >>> print_(system(buildout, with_exit_code=True), end='') # doctest: +ELLIPSIS -DONT_ACCEPT_BLANKLINE
     Upgraded:
-      zc.buildout version 1.4.4,
+      zc.buildout V.V
     Restarting.
     Generated script '/sample-buildout/bin/buildout'.
     Develop: '/sample-buildout/failrecipe'
-    recipe sys-exits
+    ...ecipe sys-exits
```
mauritsvanrees, line 282: Why did you change this?  Did you get `Recipe` with a capital R locally?  That would be weird, because the recipe in line 239 just does `sys.exit('recipe sys-exits')`.

I don't really mind the change too much, because I know these doctests can be tricky to get green.  Just wondering.

mauritsvanrees (changes requested):
The change seems good, thanks.  But I don't like some of the testing changes.  They should not be needed, and may hide future errors, especially the `...` at the end of some tests.  Could you get rid of those?

I was looking at it more in details because `update.txt` fails for me locally.  But I see it has started failing for me on the `master` branch as well, so it has nothing to do with this PR.
Error on master is:

```
File "/Users/maurits/community/buildout/src/zc/buildout/tests/update.txt", line 114, in update.txt
Failed example:
    print_(system(buildout), end='')
Expected:
    Upgraded:
      zc.buildout V.V
    Restarting.
    Generated script '/sample-buildout/bin/buildout'.
    Develop: '/sample-buildout/showversions'
    Updating show-versions.
    zc.buildout V.V
Got:
    <BLANKLINE>
    <BLANKLINE>
    Upgraded:
      zc.buildout V.V
      wheel version 0.45.1;
    Restarting.
    Generated script '/sample-buildout/bin/buildout'.
    Develop: '/sample-buildout/showversions'
    Updating show-versions.
    zc.buildout V.V
```

This has something to do with Buildout/pip finding both the proper `wheel` package, and the vendorized `wheel` within `setuptools`... :-/  Apparently not a problem on GitHub Actions, so that may be something for me to look at locally.

src/zc/buildout/tests/buildout.txt
andreclimaco: Tests when run with **setuptools < 60.0.0** have a different output (**Not found: /new_releases/zc-buildout/**).

```
File "/home/runner/work/buildout/buildout/src/zc/buildout/tests/update.txt", line 113, in update.txt
Failed example:
    print_(system(buildout), end='')
Expected:
    Upgraded:
      zc.buildout V.V
    Restarting.
    Generated script '/sample-buildout/bin/buildout'.
    Develop: '/sample-buildout/showversions'
    Updating show-versions.
    zc.buildout V.V
Got:
    Upgraded:
      zc.buildout V.V
    Restarting.
    Generated script '/sample-buildout/bin/buildout'.
    Develop: '/sample-buildout/showversions'
    Updating show-versions.
    zc.buildout V.V
    Not found: /new_releases/zc-buildout/
```

In doctest, I don't know how to validate this different output. As a workaround, I tried adding ELLIPSIS.

How should I handle this situation?

mauritsvanrees: I see.  You need to edit one line in [`testing.py`](https://github.com/buildout/buildout/blob/4.1.4/src/zc/buildout/testing.py#L605).  Change it into:

```
not_found = (re.compile(r'Not found: [^\n]+/(\w|\.|-)+/\r?\n'), '')
```

This definition is used as one of the many normalisers that we use to normalise the output.  I tried it locally and this helps for me.

andreclimaco: Now all tests have been run successfully. Thank you!

mauritsvanrees: Thanks, looks good now.  I have started the tests again (apparently I need to approve running these each time).  We can merge this when green.

mauritsvanrees (approved):

Issue #691, opened 2025-03-26
Buildout may needlessly restart for upgrade of wheel package


mauritsvanrees:
See [this comment](https://github.com/buildout/buildout/pull/690#pullrequestreview-2710188030) where I first noticed this in tests of a PR, but it already happened on master.

Locally, the `update.txt` tests have three failures like this:

```
File "/Users/maurits/community/buildout/src/zc/buildout/tests/update.txt", line 113, in update.txt
Failed example:
    print_(system(buildout), end='')
Expected:
    Upgraded:
      zc.buildout V.V
    Restarting.
    Generated script '/sample-buildout/bin/buildout'.
    Develop: '/sample-buildout/showversions'
    Updating show-versions.
    zc.buildout V.V
Got:
    <BLANKLINE>
    <BLANKLINE>
    Upgraded:
      zc.buildout V.V
      wheel version 0.45.1;
    Restarting.
    Generated script '/sample-buildout/bin/buildout'.
    Develop: '/sample-buildout/showversions'
    Updating show-versions.
    zc.buildout V.V
```

The problem here is that the doctest expects an upgrade and restart because `zc.buildout` has a new version, but unexpectedly `wheel` has a new version as well.  It goes wrong in the [check on this line](https://github.com/buildout/buildout/blob/4.1.4/src/zc/buildout/buildout.py#L1181).  For debugging, I added two print statements:

```
                print(inspect.getfile(sys.modules[project]))
                print(dist.location)
```

This prints:

```
    /Users/maurits/community/buildout/venvs/python3/lib/python3.10/site-packages/wheel/__init__.py
    /Users/maurits/community/buildout/venvs/python3/lib/python3.10/site-packages/setuptools/_vendor
```

So there is confusion over where the `wheel` package comes from, and this leads to a restart.

Strangely, this goes fine on GitHub Actions, but it fails for me locally on Mac.  It may depend on the Python and setuptools version.

I have a fix, so I will make a PR.  I just wanted to write an issue first.

PR #692, opened 2025-03-26
Install wheel before setuptools when maybe upgrading.

mauritsvanrees:
This avoids confusion between the true `wheel` package and the one vendorized by `setuptools`. Fixes https://github.com/buildout/buildout/issues/691

### DIFF
--- a/src/zc/buildout/buildout.py
+++ b/src/zc/buildout/buildout.py
@@ -1138,7 +1138,7 @@ class Buildout(DictMixin):
         self._log_level = level
 
     def _maybe_upgrade(self):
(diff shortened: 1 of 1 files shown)


news/691.bugfix
icemac, line 1: Is this file intentionally empty?

icemac (approved):

news/691.bugfix
mauritsvanrees: Oops, I forgot to fill it in.  Done now.  Thanks.

dataflake (approved):
Thank you for the ongoing work you are putting into this.

Issue #695, opened 2025-03-28
Installing a dependency that uses hatchling messes up dependency packages


mauritsvanrees:
I am using version pins from fresh Plone 6.1.1, with zc.buildout 4.1.4 and setuptools 75.8.0 and Python 3.13.

And I have as extra egg `pas.plugins.authomatic` version 2.0.0rc1, which uses [`hatchling` as build system](https://github.com/collective/pas.plugins.authomatic/blob/2.0.0rc1/pyproject.toml#L66-L68).  It has as [dependencies](https://github.com/collective/pas.plugins.authomatic/blob/2.0.0rc1/pyproject.toml#L37-L41): `Products.CMFPlone`, `authomatic`, and `plone.restapi`.

Buildout config:

```
[buildout]
extends = https://dist.plone.org/release/6.1.1/versions.cfg
parts = instance
show-picked-versions = true

[instance]
recipe = plone.recipe.zope2instance
eggs =
    Products.CMFPlone
# plone.api is a missing dependency of pas.plugins.authomatic, I made a PR.
    plone.api
    pas.plugins.authomatic
user = admin:admin

[versions]
pas.plugins.authomatic = 2.0.0rc1
```

Setup a virtual env and run buildout:

```
python3.13 -mvenv .
bin/pip install -r https://dist.plone.org/release/6.1.1/requirements.txt
bin/buildout
```

Result:

```
$ bin/buildout 
Uninstalling instance.
Installing instance.
root: Reading https://pypi.org/simple/Products.CMFPlone/
Getting distribution for 'Products.CMFPlone==6.1.1'.
Got Products.CMFPlone 6.1.1.
root: Reading https://pypi.org/simple/plone-restapi/
root: Reading https://pypi.org/simple/authomatic/
Generated script '/Users/maurits/tmp/ale/bin/instance'.
Generated interpreter '/Users/maurits/tmp/ale/parts/instance/bin/interpreter'.
Versions had to be automatically picked.
The following part definition lists the versions picked:
[versions]
authomatic = 1.3.0
plone.restapi = 9.13.3
```

First problem: this reports that `plone.restapi` is not pinned.  This is wrong: it is pinned in the `versions.cfg` that we are extending.  This may be a variant of https://github.com/buildout/buildout/issues/689

Grep in the generated script, and we see that the `plone.restapi` egg has an underscore:

```
$ grep -i restapi bin/instance 
[...]

Issue #689, opened 2025-03-17
allow-picked-versions and normalised package names


mauritsvanrees:
I use `zc.buildout` 4.1.4 and `setuptools` 75.8.2 on Python 3.12.

With `allow-picked-versions = false`, a buildout gives this error:

```
While:
  Installing.
  Getting section crontab-backup.
  Initializing section crontab-backup.
  Installing recipe z3c.recipe.usercrontab.
  Getting distribution for 'z3c.recipe.usercrontab'.
Error: Picked: z3c-recipe-usercrontab = 1.5

The `z3c-recipe-usercrontab` egg does not have a version pin and `allow-picked-versions = false`.

To resolve this, add

    z3c-recipe-usercrontab = 1.5

to the [versions] section,

OR set `allow-picked-versions = true`.
```

So I add `z3c-recipe-usercrontab = 1.5` to the versions section and run buildout again.  Same error.

So the recommended line that is printed here, should have dots instead.  Or the code that checks if a version is pinned, should recognise that this is the same version pin.  This part can be seen in the hidden traceback:

```
Traceback (most recent call last):
  File "/home/plone/website_irceline/lib/python3.12/site-packages/zc/buildout/buildout.py", line 2325, in main
    getattr(buildout, command)(args)
  File "/home/plone/website_irceline/lib/python3.12/site-packages/zc/buildout/buildout.py", line 753, in install
    [self[part]['recipe'] for part in install_parts]
     ~~~~^^^^^^
  File "/home/plone/website_irceline/lib/python3.12/site-packages/zc/buildout/buildout.py", line 1401, in __getitem__
    options._initialize()
  File "/home/plone/website_irceline/lib/python3.12/site-packages/zc/buildout/buildout.py", line 1513, in _initialize
    self.initialize()
  File "/home/plone/website_irceline/lib/python3.12/site-packages/zc/buildout/buildout.py", line 1519, in initialize
    recipe_class = _install_and_load(reqs, 'zc.buildout', entry, buildout)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/plone/website_irceline/lib/python3.12/site-packages/zc/buildout/buildout.py", line 1461, in _install_and_load
    zc.buildout.easy_install.install(
  File "/home/plone/website_irceline/lib/python3.12/site-packages/zc/buildout/easy_install.py", line 1011, in install
    return installer.install(specs, working_set)
[...]

PR #696, opened 2025-04-01
Use version pins to determine which name to give to an egg.

mauritsvanrees:
This fixes problems for dependencies of packages that were built by `hatchling`. Fixes issue #695.
Probably fixes issue #689 as well.

This builds on PR #675 because I want to have its test fixes.  But I first want that PR merged and released, so I will make this one a draft PR.  But locally it works as I want to, and the tests pass, so you can try it out.

### DIFF
--- a/news/695.feature
+++ b/news/695.feature
@@ -1,0 +1,3 @@
+Use version pins to determine which name to give to an egg.
+This fixes problems for dependencies of packages that were built by ``hatchling``.
+[maurits]
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -1792,7 +1792,7 @@ class IncompatibleConstraintError(zc.buildout.UserError):
 IncompatibleVersionError = IncompatibleConstraintError # Backward compatibility
 
 
-def call_pip_install(spec, dest):
+def call_pip_install(spec, dest, project_name=""):
     """
     Call `pip install` from a subprocess to install a
     distribution specified by `spec` into `dest`.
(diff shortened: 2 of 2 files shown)


davisagli:
@mauritsvanrees It makes sense to normalize the dist name when comparing it to the version pins. That's consistent with https://packaging.python.org/en/latest/specifications/core-metadata/#name which says "For comparison purposes, the names should be normalized before comparing."

I'm not sure the version pins are the best place to get the original name. I have a project here where I installed Products.CMFPlone 6.1.1 using pip. In site-packages, I can see products_cmfplone-6.1.1.dist-info/METADATA which conatins `Name: Products.CMFPlone`. So it seems like the package metadata (dist-info/METADATA for a wheel or PKG-INFO for an sdist) might be a better place to get the original name.

I'm not sure the best way to get that -- this is what importlib.metadata is for, but only for distributions that are already in sys.path.

Anyway, I see your experiment in #697 which might make this unnecessary.

mauritsvanrees:
> I'm not sure the version pins are the best place to get the original name. I have a project here where I installed Products.CMFPlone 6.1.1 using pip. In site-packages, I can see products_cmfplone-6.1.1.dist-info/METADATA which contains `Name: Products.CMFPlone`. So it seems like the package metadata (dist-info/METADATA for a wheel or PKG-INFO for an sdist) might be a better place to get the original name.

Yes, I see we can parse the project name from the dist-info directory in a wheel, and from the dist-info directory that pip creates while installing a source dist.
So I replaced this PR with a new one, PR #698.

I kept the changes for comparing version constraints.

PR #698, opened 2025-04-03
Get actual project name from dist.

mauritsvanrees:
Use this for naming the egg that gets created after installing a wheel or after doing a pip install of a source dist. This fixes issue #695.

This replaces PR #696.

Also: use the canonical name of a package when checking for a version constraint. This fixes issue #689.

Combined, this fixes problems for dependencies of packages that were built by `hatchling`.

<s>Internal detail: changed `_maybe_copy_and_rename_wheel` to do only that: maybe copy and rename a wheel, and return that new wheel. Previously, `_move_to_eggs_dir_and_compile` would call this function, and if a rename was indeed needed, it would call `_move_to_eggs_dir_and_compile` with the new wheel and then return the egg. Now this function simply replaces one dist with another, and then `_move_to_eggs_dir_and_compile` continues where it left off.</s> [Update: that broke things, so I undid the change.]

### DIFF
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -20,12 +20,14 @@ installed.
 
 import copy
 import distutils.errors
+import email
 import errno
 import glob
 import logging
 import operator
 import os
 import pkg_resources
+import posixpath
 import py_compile
 import re
 import setuptools.archive_util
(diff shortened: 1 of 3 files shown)


mauritsvanrees:
There was some temporary problem in the tests, but I reran the failed ones, and all is green now.

dataflake (approved):

src/zc/buildout/easy_install.py
```diff
@@ -361,7 +365,7 @@ def _version_conflict_information(self, name):
         """
         output = [
             "Version and requirements information containing %s:" % name]
-        version_constraint = self._versions.get(name)
+        version_constraint = self._versions.get(canonicalize_name(name))
```
icemac, line 368: Does this mean the `versions` section has not to contain canonicalized names?

icemac (approved):

mauritsvanrees:
I have released `zc.buildout` 4.1.8 with this fix.

PR #704, opened 2025-04-08
Log all http errors when processing package url.

mauritsvanrees:
PyPI is having issues right now.  Sometimes I get a timeout, which is reported correctly. Other times I get an `HTTP Error 503: Service Unavailable`.  This is not reported, so you have no idea of the cause when later on this leads to a DistributionNotFound.

See https://github.com/plone/buildout.coredev/pull/1013#issuecomment-2785929350

### DIFF
--- a/news/1013.bugfix
+++ b/news/1013.bugfix
@@ -1,0 +1,2 @@
+Log all http errors when processing package url.
+[maurits]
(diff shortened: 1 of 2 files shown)


mauritsvanrees:
See also [pypi/warehouse issue](https://github.com/pypi/warehouse/issues/17919).

dataflake (approved):

icemac (approved):

mauritsvanrees:
I have released `zc.buildout` 4.1.8 with this fix.

PR #708, opened 2025-04-29
Test with setuptools 79 and 80.

mauritsvanrees:
Locally I already saw problems due to recent changes. https://setuptools.pypa.io/en/latest/history.html

### DIFF
--- a/news/708.bugfix
+++ b/news/708.bugfix
@@ -1,0 +1,7 @@
+Fix development installs to still work when using setuptools 80.0.0.
+From then on, setuptools internally calls ``pip install --editable``.
+It works until and including setuptools 80.2.0.
+Later versions have other problems.
+Note that "distutils scripts" can no longer be detected with setuptools 80.
+This seems an ancient technology, and probably hardly used.
+[maurits]


mauritsvanrees:
Indeed it [fails on setuptools 80](https://github.com/buildout/buildout/actions/runs/14741842694/job/41381433352?pr=708#step:5:245) as we are [calling `setup.py develop`](https://github.com/buildout/buildout/blob/1972dc2f879e67c93a348c47a3187c31e32828ff/src/zc/buildout/easy_install.py#L1147).  This has been deprecated for long, and now might still work, but differently.  See https://github.com/pypa/setuptools/pull/4955 for the main change.

We could try replacing that with `pip install -e`.

I am unlikely to work on this the next few days.

mauritsvanrees:
Ah, wait, it fails because we call `develop` with `-mN`.  From `python setup.py devel --help`, on setuptools 79:

```
Options for 'develop' command:
...
  --multi-version (-m)       make apps have to require() a version
...
  --no-deps (-N)             don't install dependencies
...
[skipped 20 or so other options]
```

And calling it still works, just tried on a random project outside of buildout:

```
% bin/python setup.py develop -mN
...
Creating /Users/maurits/community/plone-coredev/6.0/src/Plone/lib/python3.13/site-packages/Plone.egg-link (link to .)

Installed /Users/maurits/community/plone-coredev/6.0/src/Plone
Because this distribution was installed --multi-version, before you can
import modules from this package in an application, you will need to
'import pkg_resources' and then use a 'require()' call similar to one of
these examples, in order to select the desired version:

    pkg_resources.require("Plone")  # latest installed version
    pkg_resources.require("Plone==6.0.16.dev0")  # this exact version
    pkg_resources.require("Plone>=6.0.16.dev0")  # this version or higher
```

With setuptools 80, the help is much smaller:

```
Options for 'develop' command:
  --install-dir (-d)  install package to DIR
  --no-deps (-N)      don't install dependencies
  --user              install in user site-package
                      '/Users/maurits/.local/lib/python3.13/site-packages'
  --prefix            installation prefix
  --index-url (-i)    base URL of Python Package Index
```

I removed the `-m` option from our call, but then it just fails a few lines further, because we get a different kind of result from the installation:

```
  File "/Users/maurits/community/buildout/src/zc/buildout/buildout.py", line 2329, in main
  File "/Users/maurits/community/buildout/src/zc/buildout/buildout.py", line 731, in install
    installed_develop_eggs = self._develop()
  File "/Users/maurits/community/buildout/src/zc/buildout/buildout.py", line 974, in _develop
    zc.buildout.easy_install.develop(setup, dest)
  File "/Users/maurits/community/buildout/src/zc/buildout/easy_install.py", line 1166, in develop
    return _copyeggs(tmp3, dest, '.egg-link', undo)
  File "/Users/maurits/community/buildout/src/zc/buildout/easy_install.py", line 1055, in _copyeggs
    assert len(result) == 1, str(result)
AssertionError: []
```

So `_copyeggs` may need changes as well then.  But probably only on setuptools 80+.

mauritsvanrees:
Locally, with `export SETUPTOOLS_VERSION=79.0.1; make`, it seems to work when I remove the `-m` option, but I do get lots of extra output:

```
Develop: '/Users/maurits/community/buildout/zc.recipe.egg_'
...
TEST FAILED: /Users/maurits/community/buildout/develop-eggs/tmpxj6fmpd6build does NOT support .pth files
bad install directory or PYTHONPATH

You are attempting to install a package to a directory that is not
on PYTHONPATH and which Python does not read ".pth" files from.  The
installation directory you specified (via --install-dir, --prefix, or
the distutils default setting) was:

    /Users/maurits/community/buildout/develop-eggs/tmpxj6fmpd6build

and your PYTHONPATH environment variable currently contains:

    ''

Here are some of your options for correcting the problem:

* You can choose a different installation directory, i.e., one that is
  on PYTHONPATH or supports .pth files

* You can add the installation directory to the PYTHONPATH environment
  variable.  (It must then also be on PYTHONPATH whenever you run
  Python and want to use the package(s) you are installing.)

* You can set up the installation directory to support ".pth" files by
  using one of the approaches described here:

  https://setuptools.pypa.io/en/latest/deprecated/easy_install.html#custom-installation-locations


Please make the appropriate changes for your system and try again.
```

This comes from setuptools [here](https://github.com/pypa/setuptools/blob/6c748caaaf128b7a64a5e4629787f06780b6d68d/setuptools/command/easy_install.py#L597-L598) and [here](https://github.com/pypa/setuptools/blob/main/setuptools/command/easy_install.py#L472-L475).

Anyway, we are getting into territory of `easy_install.py` (and `package_index.py`) that will at some point [be removed](https://github.com/pypa/setuptools/pull/4963) from setuptools, which would need even more changes in Buildout.

Okay, enough for now. :-)

mauritsvanrees:
It works!  The build is green (and hopefully stays green now that I did a final rebase, edited some comments, and moved to testing setuptools 80.2.0 instead of 80.0.0).  And it works locally when I try it on the Plone coredev Buildout with setuptools 80.0.0. 

This PR fixes editable installs on setuptools 80.0.0 until and including 80.2.0.  Later versions have [different problems](#710), to be solved separately.

The main fix is that with setuptools 80 we now need to "manually" create an `.egg-link` file.  That new code is in `_create_egg_link`.  I can imagine that there are corner cases where our new code does not work, but there is only so much we can do; see comments in that function.

On setuptools 79 and lower, Buildout basically does the same thing as it did before.

"distutils scripts" are not detected anymore with setuptools 80, but I don't even know what these are, so this is probably not a problem.  See comments in the code.

mauritsvanrees:
@icemac @dataflake @gotcha This is ready.
Please test this PR on setuptools 80.2.0 or lower.

Or try my follow-up PR #712 which builds on this one plus it fixes the problems caused by the removal of `setuptools/package_index.py`.  That one works with the latest setuptools version as well (80.9.0).

mauritsvanrees:
I have released `zc.buildout` 4.1.11 with this.

PR #711, opened 2025-05-15
Override pkg_resources.Environment.can_add to have better results on Mac

mauritsvanrees:
The override only changes the result on Mac: other platforms seem not affected.
This replaces my PRs #707 and #709.

Two minor other changes in here:

* Added some `logger.debug` lines in `_move_to_eggs_dir_and_compile`.
* Added option to use uv for creating the test environment by setting environment variable USE_UV a non-empty value.  We don't use this in our current test setup on GitHub Actions though.

See original report by Jens Vagelpohl a few years ago for setuptools: https://github.com/pypa/setuptools/issues/3687

And his comment in zopefoundation/meta explains well what happens from the standpoint of Buildout:
https://github.com/zopefoundation/meta/issues/181#issuecomment-1757558915

Nowadays, the same can happen with a Python installed by uv, as I saw today:

```
% bin/zopepy
>>> import pkg_resources
>>> pkg_resources.get_platform()
'macosx-11.0-arm64'
>>> pkg_resources.get_supported_platform()
'macosx-15.4-arm64'
```

Here `macosx-11.0` is the platform on which the Python was built/compiled. And `macosx-15.4` is the current platform (my laptop).

This gives problems when we get a Mac-specific wheel.  We turn it into an egg that has the result of `pkg_resources.get_supported_platform()` in its name. Then our code in `easy_install._get_matching_dist_in_location` creates a `pkg_resources.Environment` with the egg location.  Under the hood, `pkg_resources.compatible_platforms` is called, and this does not find any matching dists because it compares the platform in the egg name with that of the system, which is `pkg_resources.get_platform()`.

As far as I see, only Mac has this problem, depending on your Python and on the packages that you install:

- uv installs a pre-built Python, so the platforms likely differ. But this PR should fix that.
- When using pyenv, you compile a Python locally, so both platforms are exactly the same, so there is no problem.
- When you upgrade to a new Mac OSX version, but keep the previously compiled pyenv, the platforms will differ. You should reinstall all pyenv Pythons.  But this PR should fix that.
- Wheels with only Python code (no C, Rust, etc), will have a name ending in py3-none-any.whl, and this does not suffer from this problem.

### DIFF
--- a/.github/workflows/Makefile-scripts
+++ b/.github/workflows/Makefile-scripts
@@ -1,7 +1,7 @@
 sandbox/bin/python:
 	pip install virtualenv
 	virtualenv sandbox
-	sandbox/bin/python -mpip install setuptools
+	sandbox/bin/python -mpip install 'setuptools<80'
 
 sandbox/bin/buildout: sandbox/bin/python
 	sandbox/bin/python setup.py develop
--- a/news/609.bugfix
+++ b/news/609.bugfix
@@ -1,0 +1,4 @@
+Override `pkg_resources.Environment.can_add` to have better results on Mac.
+Without this, a freshly created Mac-specific egg may not be considered compatible.
+This can happen when the Python you use was built on a different Mac OSX version.
+[maurits]
--- a/prepare.sh
+++ b/prepare.sh
@@ -9,6 +9,12 @@ PYTHON_VERSION="${PYTHON_VERSION:-3}"
 PIP_VERSION="${PIP_VERSION}"
 SETUPTOOLS_VERSION="${SETUPTOOLS_VERSION}"
 PIP_ARGS="${PIP_ARGS:--U}"
+USE_UV="${USE_UV}"
+if test "$USE_UV"; then
+    UV_LINE="YES (override by unsetting USE_UV environment variable or making it empty)"
+else
+    UV_LINE="NO (override by giving USE_UV environment variable a non-empty value)"
+fi
 cat << MARKER
 Prepare a virtual environment for testing zc.buildout.
 
@@ -16,6 +22,7 @@ Using:
 * Python: $PYTHON_VERSION (override with PYTHON_VERSION environment variable)
 * pip: $PIP_VERSION (override with PIP_VERSION environment variable)
 * setuptools: $SETUPTOOLS_VERSION (override with SETUPTOOLS_VERSION environment variable)
+* use uv: $UV_LINE
 
 An empty version means: use whatever is already available, or install latest.
 Extra arguments for pip install: $PIP_ARGS (override with PIP_ARGS environment variable)
@@ -54,7 +61,12 @@ $PYTHON --version
 echo
 echo "Creating virtual environment in $VENV"
 mkdir -p "$VENVS"
-$PYTHON -m venv "$VENV"
+if test "$USE_UV"; then
+  echo "using uv"
+  uv venv -p $PYTHON_VERSION --seed "$VENV"
+else
+  $PYTHON -m venv "$VENV"
+fi
 
 PIP_ARGS="$PIP_ARGS pip"
 if test $PIP_VERSION; then
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -40,6 +40,7 @@ import tempfile
 import zc.buildout
 import zc.buildout.rmtree
 import zipfile
+from functools import cached_property
 from packaging import specifiers
 from packaging.utils import canonicalize_name
 from packaging.utils import is_normalized_name
@@ -67,6 +68,7 @@ default_index_url = os.environ.get(
     )
 
 logger = logging.getLogger('zc.buildout.easy_install')
+macosVersionString = re.compile(r"macosx-(\d+)\.(\d+)-(.*)")
 
 url_match = re.compile('[a-z0-9+.-]+://').match
 is_source_encoding_line = re.compile(r'coding[:=]\s*([-\w.]+)').search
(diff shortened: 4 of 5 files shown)


mauritsvanrees:
This is ready.

@dataflake @icemac Could you test if this works for you on Mac?

The way to test it, would be to use a Python where the following code gives two different results:

```
>>> import pkg_resources
>>> pkg_resources.get_platform()
'macosx-11.0-arm64'
>>> pkg_resources.get_supported_platform()
'macosx-15.4-arm64'
```

And it is probably best to not use shared eggs or a download cache.  And you need to pull in some egg with C code (or I guess Rust).

I have a small doubt whether the change in here should be done always, like now, or with an opt-in or opt-out.  But I think this PR just fixes wrong behaviour in setuptools, so it should be fine.

dataflake:
I have not run into the issue in a long time so I am not entirely sure I am testing this the right way, this is what I did:

I did a simple check with Python 3.14 (a7), because there are no binary eggs for our packages yet, which means everything must be compiled during the buildout run. The pkg_resources calls are different, but just slightly:

```python
>>> pkg_resources.get_platform()
'macosx-15.4-arm64'
>>> pkg_resources.get_supported_platform()
'macosx-15.5-arm64'
```

I used the simple buildout configuration below to build/install AccessControl, which depends on a bunch of C packages. It worked without any problem and it was clear during the buildout run that they were all compiled.

```ini
[buildout]
extensions = mr.developer
extends =
    https://raw.githubusercontent.com/zopefoundation/Zope/refs/heads/master/sources.cfg
eggs-directory = eggs
auto-checkout =
    AccessControl
parts =
    packages

[versions]
AccessControl =

[packages]
recipe = zc.recipe.egg
eggs =
    AccessControl
```

I also tried a slightly more involved setup with an old Python 3.10 built with universal2 support:


```python
>>> pkg_resources.get_platform()
'macosx-10.15-universal2'
>>> pkg_resources.get_supported_platform()
'macosx-15.5-universal2'
```

This buildout explicitly pulls some dependencies from GitHub so they must be compiled locally, it worked without a problem:

```ini
[buildout]
extensions = mr.developer
extends =
    https://raw.githubusercontent.com/zopefoundation/Zope/refs/heads/master/sources.cfg
eggs-directory = eggs
auto-checkout =
    AccessControl
    Persistence
    persistent
    Acquisition
    ExtensionClass
parts =
    packages

[sources]
persistent = git ${remotes:github}/persistent pushurl=${remotes:github_push}/persistent

[versions]
AccessControl =
Persistence =
persistent =
Acquisition =
ExtensionClass =

[packages]
recipe = zc.recipe.egg
eggs =
    AccessControl
```

dataflake:
> I have not run into the issue in a long time so I am not entirely sure I am testing this the right way, this is what I did:

I forgot to test these two scenarios against the latest released zc.buildout version (4.1.9). While the simple scenario using Python 3.14 works, the scenario using the old Python 3.10 built with universal2 support does fail, which means the change seems to fix at least some problems:

```python
...
root: Reading https://pypi.org/simple/zope-security/
Getting distribution for 'zope.security'.
root: Downloading https://files.pythonhosted.org/packages/ff/f8/1b6795d0326970b913b5f505bb1d90138a3431624bf4b57bf3a2bc98f112/zope.security-7.3-cp310-cp310-macosx_11_0_arm64.whl#sha256=48597c276cb683faa73bfee49e85c91afcd9c37d2a6608e64ea0cf677bb86230
While:
  Installing packages.
  Getting distribution for 'zope.security'.

recipe being used:
  File "/Users/jens/Desktop/testing/lib/python3.10/site-packages/zc/buildout/buildout.py", line 2329, in main
  File "/Users/jens/Desktop/testing/lib/python3.10/site-packages/zc/buildout/buildout.py", line 869, in install
  File "/Users/jens/Desktop/testing/lib/python3.10/site-packages/zc/buildout/buildout.py", line 1688, in _call
    return f()
  File "/Users/jens/Desktop/testing/eggs/zc.recipe.egg-3.0.0-py3.10.egg/zc/recipe/egg/egg.py", line 202, in install
    reqs, ws = self.working_set()
  File "/Users/jens/Desktop/testing/eggs/zc.recipe.egg-3.0.0-py3.10.egg/zc/recipe/egg/egg.py", line 76, in working_set
    ws = self._working_set(
  File "/Users/jens/Desktop/testing/eggs/zc.recipe.egg-3.0.0-py3.10.egg/zc/recipe/egg/egg.py", line 134, in _working_set
    ws = zc.buildout.easy_install.install(
  File "/Users/jens/Desktop/testing/lib/python3.10/site-packages/zc/buildout/easy_install.py", line 1009, in install
  File "/Users/jens/Desktop/testing/lib/python3.10/site-packages/zc/buildout/easy_install.py", line 782, in install
  File "/Users/jens/Desktop/testing/lib/python3.10/site-packages/zc/buildout/easy_install.py", line 619, in _get_dist
    dists = [_move_to_eggs_dir_and_compile(dist, self._dest)]
  File "/Users/jens/Desktop/testing/lib/python3.10/site-packages/zc/buildout/easy_install.py", line 2049, in _move_to_eggs_dir_and_compile
    os.rename(tmp_loc, newloc)
OSError: [Errno 66] Directory not empty: '/Users/jens/Desktop/testing/eggs/tmpok4th8wj/zope.security-7.3-py3.10-macosx-10.15-universal2.egg' -> '/Users/jens/Desktop/testing/eggs/zope.security-7.3-py3.10-macosx-10.15-universal2.egg'
```

mauritsvanrees:
Thanks for testing!
I have released 4.1.10 with this.

Issue #710, opened 2025-05-07
setuptools has removed package_index


gotcha:
`package_index` was removed from setuptools 80.3.0 https://github.com/pypa/setuptools/issues/917

I believe we should chime in https://github.com/pypa/setuptools/issues/4983 to ask for restoration and slow Deprecation.

What do you think ?

mauritsvanrees:
Chiming in there is good.

An option could also be to copy the module in our source code. Probably adapt and simplify for our use case.

mauritsvanrees:
BTW, and the tests of buildout already [fail with setuptools 80.0.0](https://github.com/buildout/buildout/pull/708).

mauritsvanrees:
Wrong tab, this was meant for PR #710, I will copy there.

PR #712, opened 2025-06-03
Use a copy of package_index.py from setuptools 80.2.0.

mauritsvanrees:
This fixes compatibility with `setuptools` 80.3.0 where this module was removed.  Merged some of our patches into this copy.
This fixes issue #710.

This PR builds on my PR #708 which fixes unrelated errors in earlier setuptools 80 versions.

Some code of our copy of this module can probably be removed, as we don't use it.  I may have a look at that still, simply removing definitions and seeing if it still works.  The main thing we need, is the `PackageIndex` class.

This PR updates GitHub Actions to add `setuptools` 80.3.0 and latest 80.9.0 to the test matrix.


dataflake:
Just like #708 I have tested this PR's branch under Python 3.13 and Python 3.10 universal2 with the same buildout configurations from https://github.com/buildout/buildout/pull/711 and it worked.

mauritsvanrees:
I have released `zc.buildout` 4.1.11 with this.

Issue #713, opened 2025-06-11
Windows on setuptools 80.3+: get_win_launcher not found


mauritsvanrees:
I tried latest `zc.buildout` 4.1.11 in the Plone core development buildout with setuptools 80.9.0.  See [this PR](https://github.com/plone/buildout.coredev/pull/1025).  This seems fine, except on [Windows](https://github.com/plone/buildout.coredev/actions/runs/15581970718/job/43879196825?pr=1025):

```
An internal error occurred due to a bug in either zc.buildout or in a
recipe being used:
Traceback (most recent call last):
  File "C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\site-packages\zc\buildout\buildout.py", line 2330, in main
    getattr(buildout, command)(args)
  File "C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\site-packages\zc\buildout\buildout.py", line 869, in install
    installed_files = self[part]._call(recipe.install)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\site-packages\zc\buildout\buildout.py", line 1689, in _call
    return f()
           ^^^
  File "c:\a\buildout.coredev\buildout.coredev\eggs\zc.recipe.egg-3.0.0-py3.12.egg\zc\recipe\egg\egg.py", line 231, in install
    return zc.buildout.easy_install.scripts(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\site-packages\zc\buildout\easy_install.py", line 1510, in scripts
    _script(module_name, attrs, spath, sname, arguments,
  File "C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\site-packages\zc\buildout\easy_install.py", line 1629, in _script
    return _create_script(contents, dest)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\site-packages\zc\buildout\easy_install.py", line 1689, in _create_script
    new_data = setuptools.command.easy_install.get_win_launcher('cli')
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #714, opened 2025-06-11
Fix error get_win_launcher not found on Windows on setuptools 80.3+.

mauritsvanrees:
Fixes https://github.com/buildout/buildout/issues/713


mauritsvanrees:
This works both here (where we test with setuptools 75) and in the Plone coredev PR (where we test with setuptools 80.9).  I merge.

mauritsvanrees:
Released in 4.1.12.

PR #716, opened 2025-07-15
Split tests, require setuptools 61+

mauritsvanrees:
For details, see the individual commit messages.

Summary:

* Split the ``buildout.txt`` test file into multiple files.  A doctest file with over 3000 lines is not very handy.
* Removed inactive tests for no longer existing ``bootstrap.py``.
* Require at least ``setuptools`` version 61.0.0. This is needed due to the changes in the test setup.

According to semantic versioning, updating a dependency is not a breaking change, and does not require a major version bump.  It does seem significant here.  As a compromise, I mark this as a breaking change, and do a minor version bump.


mauritsvanrees:
The time it takes for a single CI job to succeed, differs between setuptools versions.  See screen shot:

Screenshot 2025-07-16 at 12 16 13

In that screen shot you can't see which version is used; I will see if I can make the names shorter.  Most are between 7 and 12 minutes, but one takes [32 minutes](https://github.com/buildout/buildout/actions/runs/16302445166/job/46040196214).  That is the one with setuptools 61.  Last year I already noticed that setuptools 60 was very slow, so I switched to 59 in commit 36f141084f5a50c73829eefdd1fd5e8929037088.  Now 61 is slow, and yes, this is consistent over a few runs.  Locally I see the same slowness on Mac, and also with 62.  Version 63 is fine again.

So I would almost be inclined to use version 63 as minimum.
Actually, we could let setuptools 61 run only a small amount of tests, so we can see that it works, and run the full tests on 63.  I will see about that.

Also, the jobs with setuptools 80 are slower: there are around 12 minutes.  I will stop testing with 80.3.0.

Also, Windows takes long (27 minutes), but I don't want to investigate that.

Issue #718, opened 2025-07-16
zc.buildout 3.3 fails at installing some distributions


mauritsvanrees:
zc.buildout 3.3 has problems installing some distributions, since the many setuptools changes from this year, even when you don't think you use a newer setuptools version.  Everything seems fine in zc.buildout 4, but that requires Python 3.9 or higher.  Currently I am moving some Plone 5.2 Python 3.8 sites to a new machine, and there I run into errors.

Spoiler alert before I continue: I have an idea for a small fix, and I think we should do that in the 3.x branch.

One problem: I manually triggered a test run on the current 3.x branch and [this fails](https://github.com/buildout/buildout/actions/runs/16320247319/job/46095881014#step:5:300).  The error is `pkg_resources.DistributionNotFound: The 'zc.recipe.testrunner' distribution was not found and is required by the application`.  There is a chance that the fix I want to make actually solves this failure.  But more likely it will still fail somewhere, and we may need to accept that.

## Prerequisites

Prerequisites for a package with problems that we can fix:

* It must be a namespace package.
* It must have a `pyproject.toml`, that does **not** specify a `build-system`.
* It must have a source distribution on PyPI with a dot in the name, **not** an underscore.  Practically speaking: the source distribution must have been made with `setuptools<69.0.3`, because newer versions create source distributions with underscores.
* It must **not** have a binary wheel on PyPI.

Other packages either do not cause the problem I see, or we can't fix them in this way.

## Test setup

I have a test namespace package on GitHub and PyPI, created last year specifically for testing stuff like this:

* https://github.com/mauritsvanrees/mauritstest.namespacepackage
* https://pypi.org/project/mauritstest.namespacepackage/

We will try to install it with Buildout.

Create a virtualenv with Python 3.8.  Add a `requirements.txt` and install it:

```
setuptools==75.3.0
zc.buildout==3.3
pip==25.0.1
wheel==0.45.1
```

Add a `buildout.cfg`:

```
[buildout]
parts = test
newest = false
eggs-directory = eggs
abi-tag-eggs = false

[versions]
# Let requirements.txt be leading for the package related pins.
pip =
setuptools =
wheel =
zc.buildout =
zc.recipe.egg = 2.0.7

[test]
recipe = zc.recipe.egg
interpreter = py
eggs = mauritstest.namespacepackage==1.0.0
```

Run buildout and call the generated interpreter:

```
$ bin/buildout 
Installing test.
[...]

PR #719, opened 2025-07-16
Add 'no build isolation' option [3.x]

mauritsvanrees:
This fixes issue #718 for branch 3.x.  See there for an explanation and alternative solutions.

When OS environment variable ``BUILDOUT_PIP_NO_BUILD_ISOLATION`` is set, call ``pip install`` with the ``no-build-isolation`` option.
This may be needed when installing packages that cannot be installed with the latest setuptools version.

Note that on Python 3.8 you may want to use at most ``setuptools`` 75.3.0, whether you use the new environment variable or not.  75.3.1 and 75.3.2 have related fixes that should help, but in practice they cause problems for buildout 3.3.  I tried fixing one problem (`zc.recipe.egg` was not found after installation, both with and without the new environment variable) but that just led to the same error further on in the code.  Those kinds of problems are already fixed in zc.buildout 4, and we are not going to backport them.

I added two tiny other fixes:

* Fix typo in import that made patch_Distribution do nothing (commit d18dbadd4f58b9d925c40b351a5887b13fee86a0).  [Update: I reverted that due to problems, see commit 03e2ce8c22207979cbd9c0eedce0c546f0cef599.]
* Check if the dist exists before trying to install its console_scripts (commit 874e80ea1ef7711db06af2625285310c1224a12a).  Then you get a useful traceback, instead of an unhelpful one that does not tell you which package has the problem that we are trying to solve here.

GitHub Actions is expected to fail, and that probably can't be helped.  See my remarks in the issue.

@icemac @dataflake Please try this.

@erral Can you try as well?  I know you have some Plone 5.2 sites running on Python 3.9 to avoid this problem.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -37,6 +37,7 @@ jobs:
         TMP: ${{ env.HOME }}/AppData/Local/Temp
         TEMP: ${{ env.HOME }}/AppData/Local/Temp
         BUILD_TYPE: bare_machines
+        SETUPTOOLS_VERSION: 75.3.0
         BUILDOUT_PIP_NO_BUILD_ISOLATION: yes
         PYTHONWARNINGS: ignore
       run: |
@@ -81,6 +82,7 @@ jobs:
         package: [zest.releaser, pyspf]
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
         os: ["ubuntu-22.04"]
+        setuptools-version: ["68.0.0"]
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v4
@@ -90,6 +92,7 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Setup buildout virtualenv
       env:
+        SETUPTOOLS_VERSION: ${{matrix.setuptools-version}}
         PYTHONWARNINGS: ignore
       run: |
         make -f .github/workflows/Makefile-scripts sandbox/bin/buildout
@@ -97,6 +100,7 @@ jobs:
       env:
         PACKAGE: ${{matrix.package}}
         PYTHON_VERSION: ${{matrix.python-version}}
+        SETUPTOOLS_VERSION: ${{matrix.setuptools-version}}
         BUILDOUT_PIP_NO_BUILD_ISOLATION: yes
         PYTHONWARNINGS: ignore
       run: |


mauritsvanrees:
I had a go at fixing GitHub Actions by setting a `SETUPTOOLS_VERSION` environment variable, but this had no effect.  So I would say we ignore the tests here.

mauritsvanrees:
Okay, I have managed to fix the tests.  Except that the 'scripts' tests on Python 2.7 and 3.6 say they have started or are in queue, but I don't see anything.  Maybe for those Python versions you can wait indefinitely.  Their dependent jobs then also don't run of course.

But this is already much better than I had hoped.

mauritsvanrees:
@dataflake You get `mr.developer` 2.0.3, which has underscores in both its [distributions on PyPI](https://pypi.org/project/mr.developer/2.0.3/#files), and we can't solve that without major back porting of code.

Pin to version 2.0.2 and it should work.  Or try again without version pin: I have just released version 2.0.4 with dots in both distribution filenames.

mauritsvanrees:
All is green on [GHA](https://github.com/buildout/buildout/actions/runs/16347618725) and on [CircleCI](https://app.circleci.com/pipelines/github/buildout/buildout/590/workflows/d33094a5-b48a-475b-8272-4455fce08050).

mauritsvanrees:
I have released this as version 3.4.

As I was about to upload this to PyPI, I noticed that the distribution file names had underscores.  So I [updated `pyproject.toml`](https://github.com/buildout/buildout/commit/2011b4dac8183afc62d3a2da95104b70e1082a5b) to use `setuptools<69`.  Now they have nice dots.

https://pypi.org/project/zc.buildout/3.4/#files

mauritsvanrees:
I see where the difference is.  

* In the project where a dot or underscore in the egg name did not matter, the `bda.cache` package was pulled in as dependency of another package.
* In the project where it only worked when the egg name had a dot, `bda.cache` was explicitly specified in the `eggs`.

Having it in the eggs parameter of a recipe, triggered buildout to run extra code, for checking if scripts needed to be created, and that resulted in a traceback similar to this one:

```
  File "/Users/maurits/community/buildout/src/zc/buildout/buildout.py", line 2300, in main
  File "/Users/maurits/community/buildout/src/zc/buildout/buildout.py", line 856, in install
  File "/Users/maurits/community/buildout/src/zc/buildout/buildout.py", line 1659, in _call
    return f()
  File "/Users/maurits/tmp/testbuildout/eggs/zc.recipe.egg-2.0.7-py3.8-macosx-15.5-arm64.egg/zc/recipe/egg/egg.py", line 256, in install
    return zc.buildout.easy_install.scripts(
  File "/Users/maurits/community/buildout/src/zc/buildout/easy_install.py", line 1238, in scripts
    assert dist is not None, 'distribution not found for %s' % req.name
AssertionError: distribution not found for mauritstest.namespacepackage
```

Anyway, on zc.buildout 3.4 with the new env variable this problem is gone.

PR #724, opened 2025-09-18
Fix logic in detecting namespace init files for deletion.

mauritsvanrees:
See [this comment](https://github.com/buildout/buildout/pull/720#issuecomment-3308264168). Add tests that catch the bug.


mauritsvanrees:
I have released this in 5.0.0a3.

PR #741, opened 2026-02-26
fix: replace 4 bare except clauses with except Exception

haosenwang1018:
Replace bare except clauses with except Exception.


mauritsvanrees (approved):
That makes sense, thanks.

If the tests stay green, this can be merged.  (Ping me if I forget.)

mauritsvanrees:
I have released this in `zc.buildout` 5.1.3.  Thanks again!

PR #742, opened 2026-03-06
Insert `import pkg_resources` in scripts if namespace packages are used.

mauritsvanrees:
You can switch this off with the new buildout setting `import-pkg-resources = false`. For reasoning, see [Zope issue 1288](https://github.com/zopefoundation/Zope/issues/1288).

This setting is only useful for `zc.buildout` 4.x.  It will not be added to version 5.

I need to fix a few tests, but otherwise you can try this out.


dataflake (approved):

mauritsvanrees:
Released in 4.2.0.

PR #743, opened 2026-04-28
Test with pip 26.1.

mauritsvanrees:


mauritsvanrees:
The Windows test fails, but that is already the case when I [trigger the tests on master](https://github.com/buildout/buildout/actions/runs/25055436309/job/73397630371).  I seem to have a fix, but I will do that in a separate PR.

PR #745, opened 2026-04-28
Restrict setuptools to less than 82 if it is not pinned.

### LINKED ISSUE #744: Buildout can break itself by upgrading to setuptools 82
Current `zc.buildout` is not compatible with `setuptools` 82, because it removes the `pkg_resources` module which we heavily use.

When buildout tries to upgrade itself and a few other packages, it may install `setuptools` 82, thus leading to a broken situation.  In theory that would work like this:

* Create a virtualenv with current zc.buildout and an older setuptools, say 81.
* Create a `buildout.cfg`, but do **not** add a version pin for `setuptools`.
* Call `bin/buildout -n` to force getting newest releases.
* During the upgrade check, the installed older setuptools gets replaced by latest setuptools 82, and things are broken.

That seems to be what is happening in the Windows tests now.  Strangely though, I don't see this go wrong on the Ubuntu and Mac tests.  Also, this was working fine earlier.

* A Windows [job on March 5](https://github.com/buildout/buildout/actions/runs/22691922862/job/65791442722) passed.
* The [job I triggered today](https://github.com/buildout/buildout/actions/runs/25055436309/job/73397630371) failed.

The failed job says that pip has installed setuptools 75.8.2 as requested, but at some point this is upgraded to setuptools 82.

My idea now is to automatically restrict setuptools to less than 82.  Then the [Windows job passes](https://github.com/buildout/buildout/actions/runs/25056128706/job/73396833849).  I will put that in a PR.
When someone explicitly pins setuptools to 82, it would still be broken, but that is to be expected.

mauritsvanrees:
When ``zc.buildout`` checks if it should upgrade itself or ``setuptools``, under some circumstances this could lead to a too new ``setuptools`` version getting installed.

Somehow in the tests only Windows needs this.  But seems useful in any case.

Fixes https://github.com/buildout/buildout/issues/744

### DIFF
--- a/news/+744.feature.rst
+++ b/news/+744.feature.rst
@@ -1,0 +1,3 @@
+Restrict ``setuptools`` to less than 82 if it is not pinned.
+When ``zc.buildout`` checks if it should upgrade itself or ``setuptools``, under some circumstances this could lead to a too new ``setuptools`` version getting installed.
+[maurits]
--- a/src/zc/buildout/easy_install.py
+++ b/src/zc/buildout/easy_install.py
@@ -352,6 +352,34 @@ def dist_needs_pkg_resources(dist):
     )
 
 
+class IncompatibleConstraintError(zc.buildout.UserError):
+    """A specified version is incompatible with a given requirement.
+    """
+
+IncompatibleVersionError = IncompatibleConstraintError # Backward compatibility
+
+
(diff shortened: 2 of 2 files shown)


dataflake (approved):

mauritsvanrees:
I have released this in 5.2.0.

PR #747, opened 2026-06-02
Fix script tests by not testing on Python 3.9.

mauritsvanrees:
They [failed since 13 April](https://github.com/buildout/buildout/actions/runs/24347417557/job/71092042746) because setuptools HEAD requires Python 3.10+.

They still [passed before that](https://github.com/buildout/buildout/actions/runs/24324527963/job/71017090977#step:4:122). But as you can see in that link:

```
Successfully installed packaging-26.0 setuptools-81.0.0 wheel-0.46.3 zc.buildout-5.1.4.dev0
```

So initially the `setuptools` main branch is used for a development installation, but then the `zc.buildout` requirement `setuptools<82` kicks in. So these tests don't actually do what they say: they don't test setuptools HEAD.

But that is a different issue.

### DIFF
--- a/.github/workflows/scripts-setuptools-head.yml
+++ b/.github/workflows/scripts-setuptools-head.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13', '3.14']
+        python-version: ['3.10', '3.11', '3.12', '3.13', '3.14']
         package: [zest.releaser, pyspf]
 
     steps:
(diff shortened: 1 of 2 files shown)


dataflake (approved):

mauritsvanrees:
I triggered the script jobs workflow and [they passed](https://github.com/buildout/buildout/actions/runs/26818775054).

In the normal tests we have a [failure on Windows](https://github.com/buildout/buildout/actions/runs/26815485583/job/79058877417?pr=747#step:4:30):

```
D:\a\buildout\buildout\prepare.sh: line 59: python3.10: command not found
```

I think windows-latest has recently changed, so maybe that is it.

This PR does not change anything for that job, so it is unrelated.  I will merge.

PR #748, opened 2026-06-02
Try to fix Windows tests

mauritsvanrees:
Work in progress.


mauritsvanrees:
Ah, we expected the `OSTYPE` variable to be set to `msys`, but that has been changed to `cygwin`.

I'll see if the Windows tests pass now, and then do some rebasing and force pushing to make this a proper PR.

Issue #685, opened 2025-03-05
Should Buildout get rid of pkg_resources completely?


mauritsvanrees:
In lots of packages we are getting rid of `pkg_resources`, see for example this [big list in Plone](https://github.com/plone/Products.CMFPlone/issues/4126).

So now I wonder if technically it would be feasible to get rid of `pkg_resources` completely in Buildout.  Some of its usage can probably relatively easily be replaced.  And at some point we could remove the pkg_resources-style namespaces from our own `zc` namespace.

But how about the central parts of `zc.buildout` and `zc.recipe.egg`?  I fear that getting rid of everything will cause breakage in recipes and extensions.  Let's look at `zc.recipe.testrunner`.  Its install method starts like this:

```
    def install(self):
        options = self.options
        dest = []
        eggs, ws = self.egg.working_set(('zope.testrunner', ))

        test_paths = [ws.find(pkg_resources.Requirement.parse(spec)).location
                      for spec in eggs]
```

This contains the only line in the recipe that directly uses `pkg_resources`.  But what would you replace it with?  [`pkg_resources.Requirement.parse`](https://github.com/pypa/setuptools/blob/v75.8.2/pkg_resources/__init__.py#L2818) returns an instance of `pkg_resources.Requirement`.  This gets passed to [`ws.find`](https://github.com/pypa/setuptools/blob/v75.8.2/pkg_resources/__init__.py#L701) which only accepts a `Requirement`, nothing else.  So if we would replace `Requirement.parse` with something else, then we would need to replace `self.egg.working_set` with something else as well.

When you look at how the [working set is created in `zc.recipe.egg`](https://github.com/buildout/buildout/blob/4.1.2/zc.recipe.egg_/src/zc/recipe/egg/egg.py#L134), this uses [`zc.buildout.easy_install.install`](https://github.com/buildout/buildout/blob/4.1.2/src/zc/buildout/easy_install.py#L984), which [creates a `pkg_resources.WorkingSet`](https://github.com/buildout/buildout/blob/4.1.2/src/zc/buildout/easy_install.py#L725).

Instead of `pkg_resources.WorkingSet` we could use `importlib.metadata.distributions`.  See for example this [commit in `plone.autoinclude`](https://github.com/plone/plone.autoinclude/pull/31/commits/cd2ddf530d898376852b25cd9730826cf17cb517).

[...]

PR #751, opened 2026-08-05
Vendor pkg_resources from setuptools 81.0.0

### LINKED ISSUE #750: Buildout dependent on vulnerable version of setuptools (<83)
OSV is reporting that versions of setuptools prior to version 83 have a published vulnerability: https://osv.dev/vulnerability/PYSEC-2026-3447. zc.buildout currently requires setuptools < 83 and prevents us from updating. Buildout needs to be updated to be compatible with setuptools version 83.

icemac:
setuptools 82.0.0 removed the `pkg_resources` module, which buildout, recipes and extensions use as their data model for distributions, requirements and working sets. As discussed in #685, replacing it with `importlib.metadata`/`packaging` would be a major breaking change, so this vendors it instead — the next step suggested there.

- `src/zc/buildout/_vendor/pkg_resources/__init__.py` is a verbatim copy from the setuptools v81.0.0 tag (the last release containing it), with three documented deviations, each marked with a `LOCAL CHANGE` comment (same style as the existing `_package_index.py` vendoring). See `src/zc/buildout/_vendor/README.rst` for provenance, the list of deviations, and the re-sync procedure.
- `zc/buildout/__init__.py` installs the copy as `sys.modules['pkg_resources']` on import, so buildout's own code and any recipe/extension running inside the buildout process keeps working unchanged, with any setuptools version. If other code imported a real `pkg_resources` first (only possible with setuptools < 82), that copy wins, preserving module identity.
- New dependency: `platformdirs` (pure Python, no dependencies of its own). The three helpers `pkg_resources` needs from `jaraco.text` are vendored as `_vendor/jaraco_text.py` to avoid pulling in jaraco.text's dependency tree.
- `patches.py` skips the `WorkingSet.find` patch for the vendored copy (the 75.8.2 fix is built in); `dev.py` no longer imports `pkg_resources` before `zc.buildout` is importable; `prepare.sh` seeds `platformdirs`; coverage excludes `_vendor`; the `working_set_caching.rst` reprs are loosened (the class `__module__` changes, as with pip's `pip._vendor.pkg_resources`); new unit tests cover the alias, the `cli.exe` lookup used for Windows script generation, and the jaraco helpers.

This PR deliberately keeps the `setuptools<82` pins so the existing CI matrix proves the vendored copy is a no-op alongside setuptools 61–81. The pins are lifted in the follow-up PR (`lift-setuptools-pin`), which fixes #750.

Refs #685

**Note for reviewers:** the commit *"Upgrade packaging in the script-generation sandboxes"* fixes a pre-existing CI breakage that is unrelated to this branch: twine 7.0.0 (released 2026-07-27) requires `packaging>=26.1`, but the `generate-scripts` sandbox is stuck on the `packaging 24.2` bundled by virtualenv, which buildout cannot upgrade once it is active. Master last ran CI on 2026-06-03, before the twine release, so its `generate-scripts` job would fail today as well. Feel free to cherry-pick that commit to master independently of this PR.

— _PR created by Claude_


mauritsvanrees (approved):
This seems fine to me.  Thanks for working on it!  Hard to test in practice without your follow-up PR #752.  But let's get this one merged first.

I have created a maintenance branch 5.x, which won't be affected by this.  I made a PR for 5.x to cherry-pick your test fix from the last commit: https://github.com/buildout/buildout/pull/754

icemac:
@mauritsvanrees I only created stacked PRs to ease reviewing them. It was not my intension to merge them individually. (I started liking the new stack feature of GitHub to make PRs smaller and easier to review by splitting them up into multiple PRs which get merged together in the end.)

mauritsvanrees:
Ah, I see!  That makes sense.  I am still getting used to these new stacked PRs.  Can be useful.

PR #752, opened 2026-08-05
Allow setuptools 82 and later

### LINKED ISSUE #750: Buildout dependent on vulnerable version of setuptools (<83)

icemac:
Builds on #751 (vendoring `pkg_resources`). With buildout carrying its own copy, setuptools no longer needs to provide it:

- Drop the `<82` upper bound from the `setuptools` requirement in `setup.py`.
- Drop the automatic `<82` constraint applied to an unpinned `setuptools` during installs (introduced for #744 — its motivation was protecting buildout's own `pkg_resources` import during self-upgrades, now covered by the vendored copy).
- **Exception:** distributions using pkg_resources-style namespace packages still get `setuptools<82` added to their working set (unless setuptools is pinned), because their generated scripts run `import pkg_resources` at script run time, outside the buildout process where the vendored copy is not importable.
- CI: test setuptools 82.0.1 and 83.0.0 in the main matrix, add Python 3.9 + 82.0.1 (the newest setuptools supporting 3.9) and 3.14 + 83.0.0 rows, and run the Windows and Mac jobs with 83.0.0 so script generation (`cli.exe`) is exercised with a setuptools that no longer ships `pkg_resources`.

setuptools 83.0.0 fixes PYSEC-2026-3447 (GHSA-h35f-9h28-mq5c); environments on Python ≥ 3.10 can now install it alongside buildout. On Python 3.9 the ceiling remains setuptools 82.0.1, per setuptools' own support policy.

Fixes #750

— _PR created by Claude_

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -15,7 +15,9 @@ jobs:
       matrix:
         os: [ubuntu-24.04]
         python-version: ["3.10"]
-        setuptools-version: ["63.0.0", "65.7.0", "69.5.1", "74.1.3", "75.9.1", "79.0.1", "80.2.0", "80.10.2", "81.0.0"]
+        # 82.0.1 and later no longer contain pkg_resources;
+        # buildout uses its own vendored copy on all versions.
+        setuptools-version: ["63.0.0", "65.7.0", "69.5.1", "74.1.3", "75.9.1", "79.0.1", "80.2.0", "80.10.2", "81.0.0", "82.0.1", "83.0.0"]
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v6.0.2
@@ -75,6 +77,15 @@ jobs:
         # Use the last setuptools version that is still supported by all Python versions.
         python-version: ["3.9", "3.11", "3.12", "3.13", "3.14"]
         setuptools-version: ["75.6.0"]
+        include:
+          # setuptools 82.0.1 is the newest version supporting Python 3.9.
+          - os: ubuntu-24.04
+            python-version: "3.9"
+            setuptools-version: "82.0.1"
+          # Newest Python with newest setuptools (requires Python >= 3.10).
+          - os: ubuntu-24.04
+            python-version: "3.14"
+            setuptools-version: "83.0.0"
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v6.0.2
@@ -128,7 +139,7 @@ jobs:
       matrix:
         os: [macos-latest]
         python-version: ["3.10"]
-        setuptools-version: ["75.8.2"]
+        setuptools-version: ["83.0.0"]
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v6.0.2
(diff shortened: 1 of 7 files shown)


mauritsvanrees (approved):
Thanks a lot!  This is mostly fine.
I tried it on the Plone 6.2 coredev buildout, and on a client project.  There were some problems, but I think I have fixes for that.  Let me do that in a different PR.  But first merge this one.

BTW, I see this in the checks:

Image

DeepSource likely fails because if the `pkg_resources` code that has problems.  I did not know we had this check.  It has some good suggestions that we can look at.  But not needed now.

The other two jobs are probably set as required, but you have changed the matrix to not run them.  I will have a look later.

My notes:

On Plone coredev 6.2, adapted to use this branch and setuptools 83:

```
$ bin/buildout 
<string>:1: UserWarning: 🐎 This Python (/Users/maurits/community/plone-coredev/6.2/bin/python3.14) uses horse-with-no-namespace to make the following pkg_resources namespace packages compatible with PEP 420 namespace packages:
  
  File "/Users/maurits/community/plone-coredev/6.2/bin/buildout", line 54, in <module>
    import zc.buildout.buildout
  File "/Users/maurits/community/buildout/src/zc/buildout/__init__.py", line 48, in <module>
    from pkg_resources import PkgResourcesDeprecationWarning
ImportError: cannot import name 'PkgResourcesDeprecationWarning' from 'horse_with_no_namespace.pkg_resources' (/Users/maurits/community/plone-coredev/6.2/lib/python3.14/site-packages/horse_with_no_namespace/pkg_resources.py)
$ bin/pip uninstall horse-with-no-namespace
...
$ bin/buildout 
/Users/maurits/community/buildout/src/zc/buildout/__init__.py:44: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  from zc.buildout._vendor import pkg_resources as _vendored_pkg_resources
```

So: if you have `horse-with-no-namespace` installed, this buildout version fails to start.  I have a few lines of code to catch this problem.

The coredev buildout then works.  It has only native namespaces.

In a client project with 6.2, I do have a few `pkg_resources` style namespace packages, currently in development because I want to fix that.  I have "the horse", but I uninstall it to fix the problem from above.  Buildout then finishes with a warning:

```
** WARNING **
Some development packages are using old style namespace packages.
You should switch to native namespaces (PEP 420).
...
* collective.faq: collective
* collective.tiles.carousel: collective, collective.tiles
* cs.adminlanguage: cs
$ bin/instance fg
...
  File "/Users/maurits/shared-eggs/v5/cp310/plone.autoinclude-3.0.0-py3.10-macosx-26.4-arm64.egg/plone/autoinclude/loader.py", line 178, in load_own_packages
    module = importlib.import_module(module_name)
...
  File "/Users/maurits/zest/sfia/sfia.site/src/collective.faq/src/collective/__init__.py", line 1, in <module>
    __import__("pkg_resources").declare_namespace(__name__)
ModuleNotFoundError: No module named 'pkg_resources'
```

I did not pin `setuptools` in this buildout, so theoretically buildout should automatically install `setuptools<82`.  This does not happen because our check in `dist_needs_pkg_resources` sees that the package *does* already have setuptools as dependency, so it concludes that we don't need to add it.  We can comment out a line here.

But let me merge this one, and then we can see.

PR #754, opened 2026-08-07
Upgrade packaging in the script-generation sandboxes.

mauritsvanrees:
I have created a maintenance branch 5.x, so on master we can make some backwards incompatible changes when needed.

On the new 5.x we then need one commit from PR #751, which I cherry-pick here.  That commit fixes a test failure on current master, so also on the new 5.x.  Yesterday I triggered a test run on master and there is indeed a [failure in the tests for generating scripts](https://github.com/buildout/buildout/actions/runs/31111504765/job/92664781891).  This PR fixes it.

Here is the commit message from @icemac 

twine 7.0.0 (released 2026-07-27) requires packaging>=26.1, but the sandbox virtualenv comes with the packaging version bundled by virtualenv (24.2) and buildout cannot upgrade a distribution that is already active in the sandbox.  This makes the generate-scripts job fail for the zest.releaser package on master as well (master last ran CI on 2026-06-03, before the twine release).

### DIFF
--- a/.github/workflows/Makefile-scripts
+++ b/.github/workflows/Makefile-scripts
@@ -1,7 +1,11 @@
+# Upgrade packaging: the version bundled by virtualenv can be too old
+# for the requirements of the packages installed in the test buildout
+# (e.g. twine 7 needs packaging>=26.1), and buildout cannot upgrade a
+# distribution that is already active in the sandbox.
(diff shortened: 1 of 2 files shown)


mauritsvanrees:
Oops, I targeted this at master instead of 5.x first... Fixed now.

PR #756, opened 2026-08-07
Newer setuptools improvements

mauritsvanrees:
See the individual commits.  And see my comment at the end of https://github.com/buildout/buildout/pull/752.
(I can't point to the comment directly because it is a review comment, and there it apparently does not work...)


icemac (commented):
LGTM.

icemac (approved):

Issue #755, opened 2026-08-07
Planning for zc.buildout 6


mauritsvanrees:
Today I have created a 5.x maintenance branch.  Some bigger changes are coming for master:

- [x] PR #746 from @erral has been waiting for review (sorry!) and fixing of tests.
- [x] I have just merged PR #751 from @icemac.
- [x] Follow-up PR #752 from @icemac. 

Additionally I think this would be a good time to increase the minimum Python, setuptools, and pip versions that we support.  That makes our test matrix less big, and allows us to get rid of a bit of compatibility code.

We can discuss here the versions that we want to keep supporting and testing.

* Python:
  * We currently require 3.9 minimum, but this is out of security support
  * I think we should require 3.10 minimum.
  * We run most tests with 3.10, we could update that to 3.13 or so.
* `setuptools`:
  * We currently require `setuptools>=61.0.0,<82`.
  * We test with `["63.0.0", "65.7.0", "69.5.1", "74.1.3", "75.9.1", "79.0.1", "80.2.0", "80.10.2", "81.0.0"]` and a separate job for 61.0.0 because that version is much slower.
  * Here is the [setuptools changelog](https://setuptools.pypa.io/en/latest/history.html).
  * In our [`patches.py`](https://github.com/buildout/buildout/blob/5.2.0/src/zc/buildout/patches.py) we have:
    * `patch_interpret_distro_name` which only needed for `setuptools<70`.
    * `patch_pkg_resources_working_set_find` which is only needed for `setuptools<75.8.2`
  * Version 70 is from 21 May 2024, so already not *that* old.
  * Version 75.8.2 is from 26 February 2025.
  * As a data point: Ubuntu 24.04.4 ships with setuptools 68.1.
  * I think it is okay to require 75.8.2 as minimum.
* `pip`:
  * We are currently testing with `["21.3.1", "22.3.1", "23.3.2", "24.3.1", "25.3", "26.1.2"]`.
  * Here is the [pip changelog](https://pip.pypa.io/en/stable/news/).
  * At first I thought of requiring `>=24.0`.  But since the minimum `setuptools` version I would go for is from 2025, we could say `pip>=25.0`.
  * I would keep our current strategy: only test on the latest version for each year.  Then we avoid hunting for bugs that were fixed in pip later in the year.

So my proposal:

* Python 3.10+
* setuptools 75.8.2+
* pip 25.0+

@icemac @dataflake @erral @davisagli What do you think?

erral:
[...]

PR #758, opened 2026-08-10
Prepare for zc.buildout 6.0.0 and zc.recipe.egg 5.0.0

icemac:
Base of the branch stack that implements the version-floor proposal in #755.

The next releases from master raise the minimum supported Python, setuptools
and pip versions, so both packages get a major version bump:

* `zc.buildout` 5.2.1.dev0 → 6.0.0.dev0
* `zc.recipe.egg` 4.0.1.dev0 → 5.0.0.dev0 (it also loses Python 3.9, and its
  4.0.0 entry is the one that introduced the 3.9 floor)

## The rest of the stack

Each PR branches off the previous one:

1. **this PR** — version bump
2. #759 — require setuptools 75.8.2
3. #760 — require pip 25.0
4. #761 — require Python 3.10

setuptools comes before Python on purpose: moving the primary CI Python to
3.13 while the matrix still contains setuptools 63/65/69 would break CI,
because those setuptools releases predate Python 3.13.

— _Comment created by Claude_

### DIFF
--- a/zc.recipe.egg_/CHANGES.rst
+++ b/zc.recipe.egg_/CHANGES.rst
@@ -1,7 +1,7 @@
 Change History
 **************
 
-4.0.1 (unreleased)
+5.0.0 (unreleased)
 ==================
 
 - Nothing changed yet.
--- a/zc.recipe.egg_/setup.py
+++ b/zc.recipe.egg_/setup.py
@@ -14,7 +14,7 @@
 """Setup for zc.recipe.egg package
 """
 
-version = '4.0.1.dev0'
+version = '5.0.0.dev0'
 
 import os
 from setuptools import setup


dataflake:
I appreciate the work, but: What is the actual value of splitting up a manageable PR into four PRs, each of them accompanied by an AI-generated large "wall of text" for what are small and understandable changes, and each one then running all these GHA tests? And then you even have to explain why you stack these individual PRs in a certain way so they work at all?

I understand stacking PRs when you want to make a large PR easier to deal with. But this seems like taking it entirely too far. To me it's counterproductive.

icemac:
@dataflake The idea behind is to make the diffs easier to review. Yes, the description is a wall of autogenerated text and it is a pity that I usually ignore reading it and directly jump to the diff. But AI is chatty and it is still hard to tame it.

I was not able to adjust the size of the diffs for each step upfront and I am still playing with this new stack feature. So please bare with me as long as I try to find the sweetspot between multiple PRs in a stack and a single one PR with multiple commits plus the hint to review them individually to ease the review.

setup.py
```diff
@@ -12,7 +12,7 @@
 #
 ##############################################################################
 name = "zc.buildout"
-version = "5.2.1.dev0"
+version = "6.0.0.dev0"
```
mauritsvanrees: In my small merged PR I have already set this to `6.0.0a1.dev0`, because I want to make an alpha release first.
Please do the same in `zc.recipe.egg`.

If you are using `bumpversion` from `zest.releaser`, you can do `bump-version --breaking --alpha`.

mauritsvanrees: @icemac Could you solve this small `setup.py` conflict?  I'm not sure how that will work in a PR stack.
If it helps, we could change the version on master back, so there would hopefully be no conflict.  Than we can change the version to alpha after merge.

mauritsvanrees: Okay, I have followed some hints in the Github UI and used `gh stack checkout 762` and `gh stack rebase` to resolve the conflict.
I did not touch the `zc.recipe.egg` version as I did not want to introduce more problems than I solve.

mauritsvanrees (approved):

mauritsvanrees:
I think the whole stack is ready to merge now, if everything is green.
Then I will have another look at the version numbers before I make alpha releases.

icemac:
Thank you for reviewing these PRs and improving them. 😃

PR #761, opened 2026-08-10
Require Python 3.10 as a minimum

icemac:
Part 4 (last) of the branch stack for #755. Branches off #760.

Python 3.9 is out of security support.

* `python_requires = '>=3.10'` and the 3.9 classifier dropped, in both
  `zc.buildout` and `zc.recipe.egg`.
* `zc.recipe.egg` also gains the Python 3.14 classifier it was missing —
  `zc.buildout` already had it.

## CI

The jobs that vary something other than the Python version (`setuptools`,
`pip`, `mac`, `windows`) move from 3.10 to **3.13**, so our default test Python
is a recent one, as suggested in the issue. The per-Python job then covers
3.10, 3.11, 3.12 and 3.14.

`scripts-3.8.cfg` and `scripts-3.9.cfg` are deleted. No matrix referenced 3.8
any more, and 3.9 goes with this change. (All `scripts-3.*.cfg` files are
byte-identical to `scripts.cfg`.)

## Dead code a 3.10 floor makes unreachable

* The `u` ABI flag in `pep425tags.get_abi_tag()` — it was gated on
  `sys.version_info < (3, 3)`.
* The `except AttributeError: # Python 2` charset fallback in the patched
  `process_url`. On Python 3 this branch could never have worked anyway:
  `email.message.Message` has no `getparam`, so the fallback would raise the
  same `AttributeError` it was catching.
* The test normalizer rewriting `NameError: name` back to the pre-3.4
  `NameError: global name` wording. That one was actually live — it was
  translating modern output into the old expected text — so the `internal_errors`
  doctest now expects the modern wording instead.

I deliberately left the `f.getheader` fallback in `process_url` alone: unlike
the charset one, that is still reachable on Python 3, since `addinfourl` has
no `getheader`.

`buildout.py`'s `minor_python_versions` is also unchanged — it drives
`[section:pythonNNN]` config conditions, and dropping entries would break user
configs.

## Docs

`CONTRIBUTING.rst` was quite stale: it described master as "the upcoming
version 4" with a `3.x` maintenance branch and promised Python 3.8-3.13. It now
says master = 6, maintenance = 5.x, Python 3.10-3.14. `doc/running-tests.rst`
gets the same treatment, and both now mention the setuptools and pip floors.

## Testing

Full suite green (648 tests, 0 failures) on Python 3.10 with setuptools 75.8.2
and pip 25.3 — that is, against all three new minimums at once.

— _Comment created by Claude_


mauritsvanrees (commented):
LGTM, but the Windows job [fails quickly](https://github.com/buildout/buildout/actions/runs/31366258563/job/93390890444?pr=761):

```
D:\a\buildout\buildout\prepare.sh: line 96:
/d/a/buildout/buildout/venvs/python/Scripts/python3.exe: No such file or directory
```

I don't know why that is.  Maybe just try an older Python for Windows.

mauritsvanrees:
> LGTM, but the Windows job [fails quickly](https://github.com/buildout/buildout/actions/runs/31366258563/job/93390890444?pr=761):
> 
> ```
> D:\a\buildout\buildout\prepare.sh: line 96:
> /d/a/buildout/buildout/venvs/python/Scripts/python3.exe: No such file or directory
> ```
> 
> I don't know why that is. Maybe just try an older Python for Windows.

I investigated on a branch.  It seems a change in Python 3.13 `venv` which results in only having a `python.exe`, not a `python3.exe`.  This may change.  See https://github.com/python/cpython/issues/87915#issuecomment-4325103611

If you cherry-pick commit 263a2f0f7256667c4c3522befcb9371f02152e0f that should fix it.  See [this Windows-only test run](https://github.com/buildout/buildout/actions/runs/31538474541/job/93935171001), which is still running, but is at least successfully past the setup stage.

mauritsvanrees (approved):

PR #765, opened 2026-08-14
Test on Python 3.15 and setuptools 84.0.0.

mauritsvanrees:
And fix using the `USE_UV` environment variable.
I did not have Python 3.15 installed yet, so wanted to use `uv`, but this failed. Now this works:

```
export USE_UV=1
export PYTHON_VERSION=3.15
make test
```


icemac (approved):
I like it.

PR #766, opened 2026-08-19
Add simple but real test projects.

mauritsvanrees:
Run `make test-projects` to use the generated `bin/buildout` to run various small buildouts with test packages.  The source code for the test packages, and the buildout configs, are in the new `projects` directory`.

I have updated the main GHA file to first run `make test-projects`, and then run the other tests.  This will make it fail much sooner in case of problems.  And they will be *real* problems instead of just some unimportant difference in doctest output.

I actually started this to have a small, reproducible buildout that would show a problem in an environment with some non-working mix of setuptools/hatchling/namespaces packages.  But everything works!

Let me close with a paste of the new `projects/README-projects.md` file, which explains everything, and should be nicely readable here:

# Test projects

## Introduction

Within this `projects` directory are three directories: `packages`, `buildouts`, and `requirements`.
I want to use this to really test some common scenarios in a debuggable way.
Meaning: no doctests, simply run `bin/buildout` with a config, so you can add a breakpoint.

## Usage:

Within the main directory, so one level up, you can run:

```
make test-projects
```

This runs `make` within the `projects` directory, so you can also go to the `projects` directory yourself and run make:

```
make
```

Or run a single one:

```
make simple
```

In most of the `Makefile` targets we go to one of the buildout directories.
There we run `bin/buildout` from the top-level directory.

## Buildouts

Each sub directory has a `buildout.cfg` that does the following:

* `develop` zero, one, or more of our packages.
* Create a `bin/testscript` that has all those packages.
  With this you can check if you can import them.
* Automatically create the console scripts that each package defines.
  With this you can check that entry points are found correctly and the script works.

### Shared config

The buildouts extend `projects/buildout/shared.cfg` which has the common config.
Among others, it defines a download cache in `projects/downloads/dist`, and an eggs cache in `projects/eggs`.

### Downloads

The `make downloads` target creates final wheels of all our packages, and puts them in `projects/downloads/dist`, so buildouts can find them.

### Current buildouts

* `simple`: `hatchdemo` and `setuptoolsdemo` packages
* `namespaces`: `ns.ancient`, `ns.hat` and `ns.tools` packages
* `final`: all packages in final form, installed from a wheel in the downloads cache
* `mixed`: all namespace packages, some in development, some final

The buildout that I most expect to fail, is the mixed one.
But so far it works.

## Packages

Each package registers a console script so you can check that the package can find itself.
I want to focus on modern packaging methods:

* No `setup.py`, just `pyproject.toml`.
* Everything in a `src` directory.
* Not just `setuptools`, also `hatchling` or other projects.
* Only native namespaces, when used.

### Current packages:

* `hatchdemo`: basic package using `hatchling`
* `setuptoolsdemo`: basic package using `setuptools`
* `ns.hat`: namespace package using `hatchling`
* `ns.tools`: namespace package using `setuptools`
* `ns.ancient`: namespace package using `setuptools`, and still a `setup.py`

### New package

I created the basis of the `hatchdemo` package with this command:

```
uvx hatch new hatchdemo
```

So if you want to create a new package, you can use the same command.
But copying and editing is likely easier.
That is what I did for the other projects.
I removed lots of metadata though, which we did not need.

If you create your own package by hand, make sure to register a console script in `pyroject.toml`:

```
[project.scripts]
<script name> = "<package name>:main"
```

And add a function in the `__init__.py` of the package, printing the name of the package:

```
def main():
    print("<package name> main script")
```

Or you copy an existing demo package and go from there.

## Compare with pip/uv

Each buildout config should have a mirror in a requirements file in the `requirements` directory.
If any scripts don't work there, there is not much point in trying to get this to work in Buildout.

```
make pips
```

Note: when you run `make` or `make all`, the `pips` target is not run, as we only want to know if our buildouts work.

### DIFF
--- a/news/+1159ab01.tests.rst
+++ b/news/+1159ab01.tests.rst
@@ -1,0 +1,1 @@
+Add simple but real test projects.  [maurits]
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -36,6 +36,7 @@ jobs:
 
   setuptools:
     name: setuptools ${{ matrix.setuptools-version}}
+    needs: projects
     if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
     strategy:
       fail-fast: false
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -7,6 +7,33 @@ on:
   workflow_dispatch:
 
 jobs:
+  projects:
+    name: projects
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-24.04]
+        python-version: ["3.14"]
+        setuptools-version: ["84.0.0"]
+    runs-on: ${{ matrix.os }}
+    steps:
+    - uses: actions/checkout@v6.0.2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v6.2.0
+      with:
+        python-version: ${{ matrix.python-version }}
+        cache: 'pip' # caching pip dependencies
+    - name: Run project buildouts
+      env:
+        # When you install packages, you see far too many setuptools warnings.
+        # They drown out useful output, so ignore all warnings.
+        PYTHONWARNINGS: ignore
+        PYTHON_VERSION: ${{matrix.python-version}}
+        # PIP_VERSION: ${{matrix.pip-version}}
+        SETUPTOOLS_VERSION: ${{matrix.setuptools-version}}
+      run: make test-projects
+
   setuptools:
     name: setuptools ${{ matrix.setuptools-version}}
     if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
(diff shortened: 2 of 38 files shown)


news/+1159ab01.tests.rst
icemac, line 1: Interesting filename, how will it be rendered in the change log?

icemac (approved):
Cool idea, I really like it!

news/+1159ab01.tests.rst
mauritsvanrees: I created this with `towncrier create`, which gave it this filename with a random number so it is (most likely) unique.  Any news snippet starting with a plus sign will be rendered without showing that random number:

```
$ towncrier build --draft --version X
Loading template...
Finding news fragments...
Rendering news fragments...
Draft only -- nothing has been written.
What is seen below is what would be written.

X (2026-08-20)
--------------

New features:


- Test on Python 3.15 and setuptools 84.0.0.  [maurits]


Tests:


- Add simple but real test projects.  [maurits]
```

Compare with two other items in the current `CHANGES.rst` for which the news snippet name *was* added:

```
- Require setuptools 75.8.2 (...). (#755)
- Fix support for PEP 660 (...) files. (pep660)
```

mauritsvanrees:
FWIW, this served me as a good example of how you should configure a namespace package when using `hatchling`.  This allowed me to fix a problem I had in a branch on the Plone coredev buildout where I tried switching `plone.batching` to use `hatchling`.  Result was that `plone.batching` could not be found.  I thought this was a problem in `zc.buildout`.  But the same problem turned up when trying it with `pip`.  So the problem was in my `plone.batching` configuration.

Comparing with the examples from this PR, I had to make the following change in `plone.batching/pyproject.toml`:

```
 [tool.hatch.build.targets.wheel]
-packages = ["src/"]
+packages = ["src/plone"]
```

